PR #69, opened 2024-03-06
SDK - APIs code generation based on OpenApi definition

### LINKED ISSUE #65: SDK - Integrate OpenApi generator
Integrate model class generation through OpenApi specifications files. This will help us for future implementations, for example: https://github.com/Automattic/Gravatar-SDK-Android/pull/59#issuecomment-1966919460

As a first iteration, we can work with the `fetchProfile` specification that we probably need to create manually.

hamorillo:
Closes #65 

# Description

We want to avoid manually creating the APIs and model classes. The best way to do this is to define the public API in a `.yaml` file following the OpenApi format.

At the moment, Gravatar doesn't have those definitions, but we want to be ready, so we are adding the ProfileApi definition in [OpenApi file](https://github.com/Automattic/Gravatar-SDK-Android/blob/8b26f6af48b97e4c8cc872de73d671d23906bac0/gravatar/openapi/api-gravatar.yaml) that we've created. (Obviously, it's the first version so it can contain errors or lack information).

For the code generation, we are using [OpenApi Generator](https://openapi-generator.tech/).

As we want to have the APIs and the models within the project, we've modified the Gradle task in order to generate all the files in "build" folder and move only the desired files to the project directories.

The generated files didn't look exactly as we wanted, so we've modified some of the OpenApi generator templates to match what we want.

The last step in the generation process consist on formatting the code using Ktlint.

# Testing

- [ ] Execute
```sh
./gradlew :gravatar:openApiGenerate
```
- [ ] Verify the files are re-generated
- [ ] Install the DemoApp and verify the profile tab works as expected.


gravatar/openapi/api-gravatar.yaml
```diff
@@ -0,0 +1,134 @@
+openapi: "3.0.0"
```
maxme: Latest is 3.1.0 - https://spec.openapis.org/oas/latest.html

gravatar/build.gradle.kts
```diff
@@ -110,3 +113,50 @@ project.afterEvaluate {
         }
     }
 }
+
+openApiGenerate {
+    generatorName = "kotlin"
+    inputSpec = "${projectDir.path}/openapi/api-gravatar.yaml"
+    outputDir = "${buildDir.path}/openapi"
```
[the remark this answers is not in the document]
maxme: Can't we use `${projectDir.path}/src/main/java/com/gravatar/api` directly? (not sure if the previous directory will be deleted or not)

hamorillo:
> Can't we use ${projectDir.path}/src/main/java/com/gravatar/api directly? (not sure if the previous directory will be deleted or not)

The OpenApi generator creates a whole project with a bunch of files. We don't need all those files, so we use an output directory outside the `src` folder and move the files we need.

maxme (approved):
- [x] Execute
```sh
./gradlew :gravatar:openApiGenerate
```
- [x] Verify the files are re-generated
- [x] Install the DemoApp and verify the profile tab works as expected.

Would be nice to have:
To track the changes we've made to the templates, we should commit a version of the templates that comes directly from OpenAPI generator and do another commit with only our changes.

hamorillo:
> To track the changes we've made to the templates, we should commit a version of the templates that comes directly from OpenAPI generator and do another commit with only our changes.

I've updated the PR and split the first commit into two to keep track of the changes.

gravatar/openapi/templates/libraries/jvm-retrofit2/api.mustache
```diff
@@ -65,7 +89,7 @@ import okhttp3.MultipartBody
 
 {{/isMultipart}}
 {{/x-kotlin-multipart-import}}
-interface {{classname}} {
+internal interface {{classname}} {
```
maxme, line 92: 👍

PR #78, opened 2024-03-20
Rename classes and methods 

### LINKED ISSUE #74: Refactor and rename existing class and methods to converge with a common architecture with the iOS counterpart

maxme:
Closes #74

### Description

- Rename classes and methods
- Rewrite GravatarUtils  (features are now in AvatarUrl and HashExtensions)
- Introduced new "types": Hash and Email to make sure third party dev uses the right data in the right methods

### Testing Steps

- Check tests are running and are correct
- Run the demo app, check basic features (get profile, get avatar, upload avatar)

Note that I've made another version of it without Hash and Email classes: https://github.com/Automattic/Gravatar-SDK-Android/tree/maxime/74-classes-methods-rename-STRINGS


gravatar/src/main/java/com/gravatar/types/Hash.kt
```diff
@@ -0,0 +1,5 @@
+package com.gravatar.types
+
+public class Hash(private val hash: String) {
```
maxme, line 24: Note: I made the hash param private to avoid code like: `email.hash().hash`, forcing `email.hash().toString()` seems better to me.

gravatar/src/main/java/com/gravatar/HashExtensions.kt
```diff
@@ -0,0 +1,19 @@
+package com.gravatar
+
+import java.security.MessageDigest
+
```
maxme: I wonder if we should move these extensions to Hash.kt

gravatar/src/main/java/com/gravatar/AvatarUrl.kt
```diff
@@ -0,0 +1,101 @@
...
+     * @param email Email address
+     */
+    public constructor(email: Email) : this(email.hash())
+
+    /**
+     * Create an avatar URL from an existing Gravatar URL.
+     *
+     * @param uri Gravatar URL
+     */
+    public constructor(uri: Uri) : this(hashFromUrl(uri)) {
```
hamorillo: ❓ Should we save, in this constructor, the `avatarQueryOptions` that can be contained in the `uri`? This can be useful if, for example, I have something like, `https://www.gravatar.com/avatar/31c5543c1734d25c..e84ff82f946a34fe?d=identicon&s=42` and I just want to change the default avatar. In that case, I can get `avatarQueryOptions`, create a copy and assign it again to the AvatarUrl.

I think that's what I would expect from this constructor. 
We can do it in a different PR; this one will grow fast.

gravatar/src/main/java/com/gravatar/AvatarUrl.kt
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Get the Gravatar URL with the correct query parameters set, depending the passed options.
+     *
+     * @param options Avatar query options
+     *
+     * @return Gravatar avatar URL
+     */
+    public fun uri(options: AvatarQueryOptions? = null): Uri {
+        return canonicalUrl.buildUpon().appendGravatarQueryParameters(options).build()
+    }
```
hamorillo: ⚠️ We should either remove the `options` param from this method and use the `avatarQueryOptions` declared on the class or remove the `avatarQueryOptions`. I would with the suggested one (We don't have a definition for this in the spreadsheet):

```suggestion
    public fun uri(): Uri {
        return canonicalUrl.buildUpon().appendGravatarQueryParameters(avatarQueryOptions).build()
    }
```

gravatar/src/main/java/com/gravatar/HashExtensions.kt
hamorillo: I don't have a strong opinion tbh. Maybe it looks cleaner in `Hash.kt`, yes.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -0,0 +1,60 @@
...
+import okhttp3.ResponseBody
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+import java.io.File
+import com.gravatar.di.container.GravatarSdkContainer.Companion.instance as GravatarSdkDI
+
+public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
+    private companion object {
+        const val LOG_TAG = "Gravatar"
```
hamorillo: ❓/💡
```suggestion
        const val LOG_TAG = "AvatarService"
```

gravatar/src/main/java/com/gravatar/services/ErrorUtils.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.coroutines.launch
+import java.net.SocketTimeoutException
+import java.net.UnknownHostException
+
+internal fun httpErrorToErrorType(code: Int): ErrorType = when (code) {
+    HttpResponseCode.HTTP_CLIENT_TIMEOUT -> ErrorType.TIMEOUT
+    HttpResponseCode.HTTP_NOT_FOUND -> ErrorType.NOT_FOUND
+    in HttpResponseCode.SERVER_ERRORS -> ErrorType.SERVER
+    else -> ErrorType.UNKNOWN
+}
```
hamorillo: 💡 Should we move this to `ErrorType.kt` and rename it to something like:

```
internal fun errorTypeFromHttpCode(code: Int): ErrorType = when (code) {
```

app/src/main/java/com/gravatar/demoapp/ui/components/ProfileCard.kt
```diff
@@ -36,7 +38,13 @@ fun ProfileCard(profile: UserProfile, modifier: Modifier = Modifier, avatarImage
         horizontalAlignment = Alignment.CenterHorizontally,
     ) {
         AsyncImage(
-            gravatarUrl(profile.hash ?: "", size = with(LocalDensity.current) { avatarImageSize.toPx().toInt() }),
+            AvatarUrl(Hash(profile.hash)).uri(
```
hamorillo: 💡 Just thinking aloud. It would be nice if I could get a `Hash` instance, or why not, an `AvatarUrl` instance directly from the `UserProfile` instance. Wdyt?

hamorillo (commented):
What a great job @maxme! 🎉

gravatar/src/main/java/com/gravatar/AvatarUrl.kt
maxme: Pull please, I already removed the `avatarQueryOptions` field of that class. I think this one will depend on the questions I asked on the spreadsheet

app/src/main/java/com/gravatar/demoapp/ui/components/ProfileCard.kt
maxme: Yes definitely, using extensions on top of the generated UserProfile code maybe?

gravatar/src/main/java/com/gravatar/HashExtensions.kt
maxme: done in 4958539c1ccd61f34c09ba164675bc63e1c9462b

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
maxme: Good 👀, I haven't check those

maxme: done in b6af742

gravatar/src/main/java/com/gravatar/services/ErrorUtils.kt
maxme: Done in 500110c

app/src/main/java/com/gravatar/demoapp/ui/components/ProfileCard.kt
hamorillo: Sounds good!

gravatar/src/main/java/com/gravatar/AvatarUrl.kt
hamorillo: In my opinion, it looks better the other way. Keep `avatarQueryOptions` as a field of the class and remove it from the `uri` method. I would say the `avatarQueryOptions` is a part on the AvatarUrl and keep it inside that instance will be easier to understand for them.

Also, if you consider what I wrote [here](https://github.com/Automattic/Gravatar-SDK-Android/pull/78#discussion_r1531726202) can make sense.

maxme:
Just a note, I've made another branch, same patch but I replaced Hash and Email classes by strings: https://github.com/Automattic/Gravatar-SDK-Android/tree/maxime/74-classes-methods-rename-STRINGS

app/src/main/java/com/gravatar/demoapp/ui/components/ProfileCard.kt
maxme: Logged here: https://github.com/Automattic/Gravatar-SDK-Android/issues/80

gravatar/src/main/java/com/gravatar/AvatarUrl.kt
maxme: Done in 5b038738935ff7c8f1abb0556f5d21ff868935d3

PR #82, opened 2024-03-22
Replace Uri by java.net.URL in the core module

### LINKED ISSUE #73: Replace android.net.Uri by java.net.URL
We'll need to add query manipulation functions

maxme:
Closes #73 

### Description

- I removed all references to `android.net.Uri` and replaced them by `java.net.URL`.
- I updated the tests and added a couple more (one to check for the `f=n` query parameter and another one to check that the AvatarUrl created with an URL has a non-empty hash path).
- Note: I only URL encoded the default avatar, as this is the only one to have user generated inputs

### Testing Steps

- [ ] Check that the updated tests are correct and run them
- [ ] Run the demo app

### DIFF
--- a/gravatar/src/main/java/com/gravatar/AvatarUrl.kt
+++ b/gravatar/src/main/java/com/gravatar/AvatarUrl.kt
@@ -92,11 +92,14 @@ public class AvatarUrl {
         } ?: false
     }
 
-    public fun url(): URL {
-        return URL(
-            canonicalUrl.protocol,
-            canonicalUrl.host,
-            canonicalUrl.path.plus(queryParameters(avatarQueryOptions)),
-        )
-    }
+    /**
+     * Get the Avatar URL including query parameters.
+     *
+     * @return URL
+     */
+    public fun url(): URL = URL(
+        canonicalUrl.protocol,
+        canonicalUrl.host,
+        canonicalUrl.path.plus(queryParameters(avatarQueryOptions)),
+    )
 }
--- a/gravatar/src/main/java/com/gravatar/AvatarUrl.kt
+++ b/gravatar/src/main/java/com/gravatar/AvatarUrl.kt
@@ -1,48 +1,48 @@
 package com.gravatar
 
-import android.net.Uri
 import com.gravatar.GravatarConstants.GRAVATAR_WWW_BASE_HOST
 import com.gravatar.types.Email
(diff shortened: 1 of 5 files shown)


gravatar/src/main/java/com/gravatar/AvatarUrl.kt
```diff
@@ -96,7 +92,11 @@ public class AvatarUrl {
         } ?: false
     }
 
-    public fun uri(): Uri {
-        return canonicalUrl.buildUpon().appendGravatarQueryParameters(avatarQueryOptions).build()
+    public fun url(): URL {
+        return URL(
```
hamorillo: ⛏ Nitpicking

```suggestion
    public fun url(): URL = URL(
```

gravatar/src/test/java/com/gravatar/AvatarUrlTest.kt
```diff
@@ -234,7 +245,19 @@ class AvatarUrlTest {
                 AvatarQueryOptions(
                     defaultAvatarOption = CustomUrl("https://example.com/?encoded=true&please=yes"),
                 ),
-            ).uri().toString(),
+            ).url().toString(),
+        )
+    }
+
+    @Test
+    fun `Force default avatar false must generate f=n`() {
```
hamorillo: ❓ Should we use lowercase for the first letter in the fun name? (Except in the other test where is the class name `AvatarUrl`)

```suggestion
    fun `force default avatar false must generate f=n`() {
```

hamorillo (approved):
❤️ Awesome job @maxme! 

:shipit:

gravatar/src/main/java/com/gravatar/AvatarUrl.kt
maxme: Fixed in [30a7c5f](https://github.com/Automattic/Gravatar-SDK-Android/pull/82/commits/30a7c5fc3b18440bbd1472e92b74adc875979655)

gravatar/src/test/java/com/gravatar/AvatarUrlTest.kt
maxme: Fixed in [30a7c5f](https://github.com/Automattic/Gravatar-SDK-Android/pull/82/commits/30a7c5fc3b18440bbd1472e92b74adc875979655)

PR #84, opened 2024-03-27
First set of UI components

### LINKED ISSUE #87: Create new atomic components and Full Profile Card using them

maxme:
Closes #87 

## What changed
- New Profile Card component (demonstrated in the demo app)
- New atomic components
- User profile extension: `formattedUserInfo` with some unit tests to format the string in the "personal information" section of the profile card.
- Added new fields to `api-gravatar.yaml`:
  - `job_title`, `company`, `pronunciation` used in the "personal information" section of the profile card.
  - `accounts` used in the "social media" section of the profile card.
- Added social media icons for all supported Gravatar services. All icons come directly from
- Sections "About Me" and "Personal Information" should be expandable. If and only if the text is longer that the maxLines defined, we should show an ellispis at the end of the text and the component should be clickable. Clicking on it should show a dialogto show the full content.
  
## What to check
- Atomic components: what parameter should we provide? Example the number of items in the SocialMediaRow component.
- Margins / Padding / Spacing: they should be the same as defined in the design document.

## Some details

*Social media*:
- Known services and icons are locally served but if we get an unknown service from the API response, we fetch and show the svg file from the provided url.


demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -147,7 +147,7 @@ private fun GravatarTabs(
 
 @Composable
 private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwable?) -> Unit) {
-    var email by remember { mutableStateOf("gravatar@automattic.com", neverEqualPolicy()) }
+    var email by remember { mutableStateOf("maxime.biais@automattic.com", neverEqualPolicy()) }
```
maxme, line 150: This helps a lot with the default card, but I think we should create a set of live test accounts and eventually have some kind of autofill in the email text field for quickly switching profiles

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/DisplayName.kt
```diff
@@ -0,0 +1,11 @@
+package com.gravatar.ui.components.atomic
+
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import com.gravatar.api.models.UserProfile
+
+@Composable
+public fun DisplayName(profile: UserProfile, modifier: Modifier = Modifier) {
+    Text(text = profile.displayName ?: "", modifier = modifier)
```
maxme: I know this is kind of useless and limiting, but there is one good thing about all these components: they take a UserProfile and nothing else. Also this would help someone who just need to know what's the logic (in this case using the "displayName" field or the empty string).

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/AboutMe.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.gravatar.api.models.UserProfile
+
+@Composable
+public fun AboutMe(profile: UserProfile, modifier: Modifier = Modifier, maxLines: Int = 2) {
+    EpandableText(profile.aboutMe ?: "", modifier, maxLines)
+}
+
+@Preview
+@Composable
+private fun AboutMe() {
```
hamorillo: ⛏

```suggestion
private fun AboutMePreview() {
```

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
```diff
@@ -0,0 +1,54 @@
...
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.window.Dialog
+
+@Composable
+internal fun EpandableText(text: String, modifier: Modifier = Modifier, maxLines: Int = 2) {
```
hamorillo: ⛏
```suggestion
internal fun ExpandableText(text: String, modifier: Modifier = Modifier, maxLines: Int = 2) {
```

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/AboutMe.kt
```diff
@@ -0,0 +1,25 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.tooling.preview.Preview
+import com.gravatar.api.models.UserProfile
+
+@Composable
+public fun AboutMe(profile: UserProfile, modifier: Modifier = Modifier, maxLines: Int = 2) {
+    EpandableText(profile.aboutMe ?: "", modifier, maxLines)
```
hamorillo: ⛏

```suggestion
    EpandableText(profile.aboutMe.orEmpty(), modifier, maxLines)
```

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/DisplayName.kt
hamorillo: To me, it makes complete sense.

⛏
```suggestion
    Text(text = profile.displayName.orEmpty(), modifier = modifier)
```

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
```diff
@@ -0,0 +1,54 @@
...
+            ) {
+                Text(
+                    text = text,
+                    modifier = Modifier
+                        .padding(16.dp)
+                        .fillMaxWidth()
+                        .wrapContentSize(),
+                    textAlign = TextAlign.Start,
+                )
+            }
```
hamorillo, line 52: 💡 Just thinking aloud for future improvements. We can add a new param to the `ExpandableText` in order to accept the composable that should be rendered inside the dialog. Something like:

```
@Composable
internal fun ExpandableText(
    text: String, modifier: Modifier = Modifier, maxLines: Int = 2,
    dialogContent: @Composable (text: String) -> Unit = { dialogText -> DefaultCard(dialogText) }
) {
```

With that approach, everything inside the dialog can be customized.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+}
```
hamorillo, line 55: 💡 I think the Preview is nice to have. It's helpful to launch the Composable isolated and try it.

```suggestion
}

@Preview
@Composable
private fun ExpandableTextPreview() {
    ExpandableText("Text that \n can be expanded \n by clicking on it.")
}
```

image

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/UserInfo.kt
```diff
@@ -0,0 +1,29 @@
...
+@Composable
+public fun UserInfo(profile: UserProfile, modifier: Modifier = Modifier, maxLines: Int = 2) {
+    // TODO this doesn't work with one Text field due. If the job_title and the company line is too long,
+    // it will to break the layout
+    EpandableText(profile.formattedUserInfo(), modifier, maxLines)
+}
+
+@Preview
+@Composable
+private fun UserInfo() {
```
hamorillo: ⛏
```suggestion
private fun UserInfoPreview() {
```

gravatar-ui/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="view_profile_button">View Profile</string>
```
hamorillo, line 3: 💡 Looks like, at some point, we'll need to think about i18n.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
```diff
@@ -0,0 +1,39 @@
...
+import androidx.compose.ui.unit.dp
+import coil.compose.AsyncImage
+import com.gravatar.AvatarQueryOptions
+import com.gravatar.api.models.UserProfile
+import com.gravatar.extensions.avatarUrl
+
+@Composable
+public fun Avatar(
+    profile: UserProfile,
+    size: Dp,
```
hamorillo, line 18: ⚠️ I don't think this is a good idea. As a developer, I wouldn't expect that param to change the modifier internally. 

Also, if I specify a size in the modifier externally that will be the size:

```
Avatar(
    profile = profile,
    size = 72.dp,
    modifier = Modifier.clip(CircleShape).size(140.dp),
)
```

image

Another point is that we are internally changing the `AvatarQueryOptions` that a user could have defined.

On the other hand, I'm trying to figure out how to achieve a good experience. We can merge and think about how to improve it.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
```diff
@@ -0,0 +1,117 @@
...
+    TripIt("tripit", R.drawable.tripit_icon),
+    Tumblr("tumblr", R.drawable.tumblr_icon),
+    Twitch("twitch", R.drawable.twitch_icon),
+    Twitter("twitter", R.drawable.twitter_icon),
+    Vimeo("vimeo", R.drawable.vimeo_icon),
+    WordPress("wordpress", R.drawable.wordpress_icon),
+    ;
+
+    companion object {
+        public val shortnames = entries.associateBy { it.shortname }
```
hamorillo: ❓
```suggestion
        private val shortnames = entries.associateBy { it.shortname }
```

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
```diff
@@ -0,0 +1,117 @@
...
+
+public fun mediaList(profile: UserProfile): List<SocialMedia> {
+    val mediaList = mutableListOf<SocialMedia>()
+    // Force the Gravatar icon
+    profile.profileUrl?.let {
+        mediaList.add(SocialMedia(URL(it), LocalIcon.Gravatar.name, icon = LocalIcon.Gravatar))
+    }
+    // List and filter the other accounts from the profile, keep the same order coming from UserProfile.accounts list
+    profile.accounts?.let {
+        for (account in it) {
```
hamorillo: ⛏
```suggestion
    profile.accounts?.forEach { account ->
```

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
```diff
@@ -0,0 +1,117 @@
...
+}
+
+@Composable
+fun SocialIconRow(socialMedia: List<SocialMedia>, modifier: Modifier = Modifier, maxIcons: Int = 4) {
+    var count = 0
+    Row(modifier = modifier) {
+        for (media in socialMedia) {
+            if (count++ >= maxIcons) {
+                break
+            }
```
hamorillo: ⛏
```suggestion
        socialMedia.take(maxIcons).forEach { media ->
```

gravatar/src/main/java/com/gravatar/extensions/UserProfileExtensions.kt
```diff
@@ -26,3 +26,14 @@ public fun UserProfile.avatarUrl(avatarQueryOptions: AvatarQueryOptions? = null)
 public fun UserProfile.profileUrl(): ProfileUrl {
     return ProfileUrl(this.hash())
 }
+
+/**
+ * Get formatted user info for a user profile.
+ */
+public fun UserProfile.formattedUserInfo(): String {
+    val line1 = arrayOf(this.jobTitle, this.company).filter { !it.isNullOrBlank() }.joinToString(", ")
```
hamorillo: ❓ Why do you use an `arrayOf ` instead of `listOf`?

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
```diff
@@ -0,0 +1,117 @@
...
+            SocialIcon(media = media, modifier = Modifier.size(32.dp))
+        }
+    }
+}
+
+@Composable
+fun SocialIconRow(profile: UserProfile, modifier: Modifier = Modifier, maxIcons: Int = 4) {
+    SocialIconRow(mediaList(profile), modifier, maxIcons)
+}
```
hamorillo, line 117:
```suggestion
}

@Preview(showBackground = true)
@Composable
fun SocialIconRowPreview() {
    SocialIconRow(
        socialMedia = LocalIcon.entries.map {
            SocialMedia(URL("https://${it.shortname}.com"), it.shortname, icon = it)
        },
        maxIcons = 5
    )
}
```

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
```diff
@@ -0,0 +1,65 @@
...
+                    tint = LocalTextStyle.current.color,
+                    contentDescription = "",
+                )
+            },
+        ),
+    )
+
+    TextButton(
+        onClick = {
+            uriHandler.openUri(profile.profileUrl.toString())
```
hamorillo: ⚠️ If `profileUrl` is `null` for some reason, the `uriHandler` will crash. 

```suggestion
            uriHandler.openUri(profile.profileUrl().url.toString())
```

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
```diff
@@ -0,0 +1,65 @@
...
+        },
+        modifier = modifier.padding(0.dp, 0.dp, 0.dp, 0.dp),
+        contentPadding = PaddingValues(start = 0.dp, end = 0.dp),
+    ) {
+        Text(text, inlineContent = inlineContent)
+    }
+}
```
hamorillo, line 70:
```suggestion
}

@Preview
@Preview(locale = "ar", showBackground = true) // Preview in RTL mode
@Composable
private fun ViewProfileButtonPreview() {
    ViewProfileButton(UserProfile("4539566a0223b11d28fc47c864336fa27b8fe49b5f85180178c9e3813e910d6a"))
}
```

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
```diff
@@ -0,0 +1,65 @@
...
+                )
+            },
+        ),
+    )
+
+    TextButton(
+        onClick = {
+            uriHandler.openUri(profile.profileUrl.toString())
+        },
+        modifier = modifier.padding(0.dp, 0.dp, 0.dp, 0.dp),
```
hamorillo: ❓Why do we need the `.padding(0.dp, 0.dp, 0.dp, 0.dp)` here?

gravatar-ui/src/main/java/com/gravatar/ui/components/ProfileCard.kt
```diff
@@ -0,0 +1,102 @@
...
+            ) {
+                ViewProfileButton(profile, Modifier.padding(0.dp))
+            }
+        }
+    }
+}
+
+@Preview
+@Composable
+fun PreviewUserProfileCard() {
```
hamorillo: ⛏
```suggestion
fun UserProfileCardPreview() {
```

gravatar-ui/src/main/java/com/gravatar/ui/components/ProfileCard.kt
```diff
@@ -0,0 +1,102 @@
...
+            currentLocation = "Crac'h, France",
+            pronouns = "They/Them",
+            accounts = listOf(
+                Account(name = "Mastodon", url = "https://mastodon.social/@ddoe"),
+                Account(name = "Tumblr", url = "https://ddoe.tumblr.com"),
+                Account(name = "WordPress", url = "https://ddoe.wordpress.com"),
+            ),
+            aboutMe = "I'm a farmer, I love to code. I ride my bicycle to work. One apple a day keeps the " +
+                "doctor away. This about me description is quite long, this is good for testing.",
+            emails = arrayListOf(Email(primary = true, value = "john@doe.com")),
```
hamorillo: ❓ Why not `listOf`?

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+    // List and filter the other accounts from the profile, keep the same order coming from UserProfile.accounts list
+    profile.accounts?.let {
+        for (account in it) {
+            if (LocalIcon.valueOf(account.shortname) != null) {
+                // Add local icon if the shortname exists in our predefined list
+                mediaList.add(SocialMedia(URL(account.url), account.name, icon = LocalIcon.valueOf(account.shortname)))
+            } else {
+                // Add a "remote" icon (using the url coming from the endpoint response)
+                mediaList.add(SocialMedia(URL(account.url), account.name, iconUrl = URL(account.iconUrl)))
```
hamorillo, line 67: 🛑 If `accout.iconUrl` is `null` the app will crash

```
Caused by: java.net.MalformedURLException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
                                                                                                    	at java.net.URL.<init>(URL.java:635)
                                                                                                    	at java.net.URL.<init>(URL.java:498)
                                                                                                    	at java.net.URL.<init>(URL.java:447)
                                                                                                    	at com.gravatar.ui.components.atomic.SocialMediaKt.mediaList(SocialMedia.kt:66)

```

gravatar-ui/src/main/java/com/gravatar/ui/components/ProfileCard.kt
```diff
@@ -0,0 +1,102 @@
...
+import com.gravatar.ui.components.atomic.ViewProfileButton
+
+@Composable
+fun ProfileCard(profile: UserProfile, modifier: Modifier = Modifier) {
+    Column(
+        modifier = modifier,
+        horizontalAlignment = Alignment.Start,
+        verticalArrangement = Arrangement.Top,
+    ) {
+        Row(modifier = Modifier.padding(0.dp)) {
```
hamorillo: ❓ We don't need the `.padding(0.dp)`, right?

gravatar-ui/src/main/java/com/gravatar/ui/components/ProfileCard.kt
```diff
@@ -0,0 +1,102 @@
...
+            )
+            Column(modifier = Modifier.padding(14.dp, 0.dp, 0.dp, 0.dp)) {
+                ProvideTextStyle(TextStyle(fontSize = 18.sp, fontWeight = FontWeight.Bold, lineHeight = 24.sp)) {
+                    DisplayName(profile)
+                }
+                ProvideTextStyle(
+                    TextStyle(
+                        fontSize = 14.sp,
+                        lineHeight = 20.sp,
+                        color = MaterialTheme.colorScheme.outline,
```
hamorillo, line 53: 💡I'm not an expert in handling style in composing. But maybe we can define some textStyle to share between component. As a future improvement! Not for this PR!

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
hamorillo: Sounds good!

hamorillo (commented):
Congrats @maxme! It's an extraordinary work!

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
hamorillo: I was talking with @mlumeau before and we are not sure if we should harcode private mails in the code. Wdyt?

maxme: This address is already crawlable from some public places (commits, pom files), so it's not a big deal. Would be better to have live test accounts though, I can create one before merging this.

gravatar-ui/src/main/java/com/gravatar/ui/components/ProfileCard.kt
maxme: removed in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/ProfileCard.kt
maxme: updated in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/UserInfo.kt
maxme: Updated in in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
maxme: updated in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
maxme: 👍 updated in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar/src/main/java/com/gravatar/extensions/UserProfileExtensions.kt
maxme: updated in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
maxme: updated in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
maxme: updated in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
maxme: not needed, removed in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/ProfileCard.kt
maxme: updated in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/ProfileCard.kt
maxme: Not sure what's the best way to deal with with this, I wonder if we should group all used styles somewhere and pass it to the main component. We should look at other libraries that expose Jetpack Components.

gravatar-ui/src/main/res/values/strings.xml
maxme: Yes, ideally we should provide translations for these strings. It's not a huge deal for third party devs, they can provide translations as well for the strings used in their dependencies (one thing we do in WPAndroid for example).

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
maxme: Yes, that's a good idea, we'll track this in https://github.com/Automattic/Gravatar-SDK-Android/issues/92

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/AboutMe.kt
maxme: updated in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
maxme: fixed in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/AboutMe.kt
maxme: updated in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/DisplayName.kt
maxme: fixed in 069e0eae065a0da28fc1a244f1c49788b2002ec0

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
maxme: New preview (a bit different from your proposable) in 7027888

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
maxme: Proposal added in 7027888

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
maxme: I agree, it's not ideal

> Another point is that we are internally changing the AvatarQueryOptions that a user could have defined.

The only reason of having this here is that we want to craft the right avatar url depending on the size.

I tried to measure the size of the image view (after modifiers were applied) to set the right size in `AvatarQueryOptions` but I couldn't make it work (delay to get the measurement)

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
maxme: Added in 7027888

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
maxme: Good catch, yes It will crash if the backend returns a malformed URL as well, eg. "example.com" (` if no protocol is specified, or an unknown protocol is found, or spec is null.`) so I guess the only thing to do here is to catch and ignore the exception

maxme: BTW, Same issue if the `account.url` is malformed or null

maxme: Fixed in 3af96216cdc157d8ec5d9316b89084c000e32bbc

maxme:
Ready for another pass

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
```diff
@@ -0,0 +1,134 @@
...
+    companion object {
+        private val shortnames = entries.associateBy { it.shortname }
+
+        public fun valueOf(shortname: String?): LocalIcon? {
+            return shortnames[shortname]
+        }
+    }
+}
+
+public class SocialMedia(val url: URL, val name: String, val iconUrl: URL? = null, val icon: LocalIcon? = null)
```
hamorillo, line 53: ⛏ While reviewing, I noticed that `iconUrl` and `icon` are nullable because they could be one or the other. A solution to prevent that could be something like: 

```suggestion
public sealed class SocialMedia{
    abstract val url: URL
    abstract val name: String

    data class LocalSocialMedia(override val url: URL, override val name: String, val icon: LocalIcon) : SocialMedia()
    data class RemoteSocialMedia(override val url: URL, override val name: String, val iconUrl: URL) : SocialMedia()
}
```

Anyway, not a big deal.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
```diff
@@ -0,0 +1,134 @@
...
+@Preview(showBackground = true)
+@Composable
+fun SocialIconRowPreview() {
+    val userProfile = UserProfile(
+        hash = "",
+        accounts = listOf(
+            Account(name = "Mastodon", url = "https://example.com", shortname = "mastodon"),
+            Account(name = "Tumblr", url = "https://example.com", shortname = "tumblr"),
+            // Invalid url, should be ignored:
+            Account(name = "TikTok", url = "example.com", shortname = "tiktok"),
```
hamorillo, line 128: 💡 We can create some unit test for `fun mediaList(profile: UserProfile)` to test that cases.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
```diff
@@ -0,0 +1,83 @@
...
+            },
+        ),
+    )
+
+    TextButton(
+        onClick = {
+            uriHandler.openUri(profile.profileUrl().url.toString())
+        },
+        contentPadding = PaddingValues(start = 0.dp, end = 0.dp),
+        modifier = modifier,
```
hamorillo, line 66: 💡/❓ Should we start taking care of the accesibility stuff?

```suggestion
        modifier = modifier
            .semantics(mergeDescendants = true) {}
            .clearAndSetSemantics { contentDescription = description },
```

hamorillo (approved):
Looks great! :shipit: 

I left some minor comments that we can integrate if necessary in a different PR.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
maxme: Hmmm yeah but we'll still need to watch for malformed urls and we shouldn't trust the backend anyway.

BTW, we have to be careful everytime we use the URL() constructor, maybe we can find a better way to prevent crashes.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
maxme: Good point, but maybe we can keep this one for documention purpose

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
maxme: Hmmm, I'm not sure how that help here, "View profile" button is correctly described. If we add the code you propose, we'll give third party devs an option to override the description, but we don't give them the option to change the text (even if they can override `R.string.view_profile_button`). IMO if we want to do this, we should also provide a param to pass a string for the text, but in that case we should just make a generic Button component that open a link on tap.

PR #89, opened 2024-04-03
Forcing a concrete version of "jackson-core"

### LINKED ISSUE #85: OpenAPI generation fails with current Dokka setup
There is an incompatibility with the current setup of Dokka and OpenAPI. Running `generateOpenApi` tasks fails with:

```
* What went wrong:
Execution failed for task ':gravatar:openApiGenerate'.
> 'com.fasterxml.jackson.core.io.ContentReference com.fasterxml.jackson.dataformat.yaml.YAMLFactory._createContentReference(java.lang.Object)'
```

`com.fasterxml.jackson.core` is a common dependency of Dokka and OpenAPI.

I think the issue was introduced when Dokka was added to the root build file in https://github.com/Automattic/Gravatar-SDK-Android/pull/76

Note: we should run OpenAPI generation in CI just to check generation works fine.

hamorillo:
Closes #85 

### Description

There is a conflict between the Jackson Core versions used by Dokka and the OpenAPI Generator. Forcing the last version to avoid that issue.

> Note: we should run OpenAPI generation in CI just to check generation works fine.

I'm not completely sure which is the best place to add. Buildkite? GitHub Actions? 
Maybe we can wait until we improve the REST API part and think about the best approach for the code generation. Wdyt?

### Testing Steps

- Clone the branch
- Execute and verify that works: `./gradlew :gravatar:openApiGenerate`
- Execute and verify that works: `./gradlew dokkaHtmlMultiModule`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -19,6 +19,11 @@ plugins {
 buildscript {
     dependencies {
         classpath("org.jetbrains.dokka:dokka-base:1.9.20")
+
+        /**
+         * Forcing the use of Jackson Core to avoid a conflict between the version used by Dokka and the one used by OpenAPI Generator.
+         */
+        classpath("com.fasterxml.jackson.core:jackson-core:2.15.3")
     }
 }
 


maxme:
> I'm not completely sure which is the best place to add. Buildkite? 

Since lint check, build and unit tests run on buildkite, I'd also put this on buildkite.

maxme (approved):
This fix the issue I had with openApi.

LGTM :shipit: - we can add the buildkite check later

PR #94, opened 2024-04-08
Let dev pass a default email to the local.properties file

maxme:
### Description

Let devs configure their own default email in the demo app

### Testing Steps

- Build and run the demo app, check the default email address is: "gravatar@automattic.com"
- Add a new property to your `local.properties` file: `demo-app.email=test@example.com`, then build and run the demo app, make sure the default email address is "test@example.com"

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,7 +21,8 @@ buildscript {
         classpath("org.jetbrains.dokka:dokka-base:1.9.20")
 
         /**
-         * Forcing the use of Jackson Core to avoid a conflict between the version used by Dokka and the one used by OpenAPI Generator.
+         * Forcing the use of Jackson Core to avoid a conflict between the version used by
+         * Dokka and the one used by OpenAPI Generator.
          */
         classpath("com.fasterxml.jackson.core:jackson-core:2.15.3")
     }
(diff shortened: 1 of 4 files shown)


hamorillo (commented):
Thanks for taking care of this @maxme 🙏🏼! 

⚠️ I think we should also use the `BuildConfig.DEMO_EMAIL `in the [AvatarUpdateTab](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/demo-app/src/main/java/com/gravatar/demoapp/ui/AvatarUpdateTab.kt)

❓ Not a big deal right now, but Can it make sense to include a `local.properties-example`? We can create a ticket for this.

maxme:
> ⚠️ I think we should also use the BuildConfig.DEMO_EMAIL in the [AvatarUpdateTab](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/demo-app/src/main/java/com/gravatar/demoapp/ui/AvatarUpdateTab.kt)

Yes thanks, I missed that one, updated in c7552cfe1949c0e693a0c0f0a96cf637ed746d03

>❓ Not a big deal right now, but Can it make sense to include a local.properties-example? We can create a ticket for this.

I think it's fine to keep this undocumented, it doesn't prevent anyone to test, it's just a shortcut for folks who know :) . If we do want an `-example` file, we shouldn't use `local.properties`as it's generated by AS.

hamorillo (approved):
:shipit:

PR #95, opened 2024-04-08
Publish the documentation on tags

### LINKED ISSUE #72: Generate and publish documentation for release tags
I think we should upload documentation for each release. For example:

https://automattic.github.io/Gravatar-SDK-Android/0.1.0/index.html

maxme:
Closes #72

### Description

Add a filter to run dokka action on tags push. Match semver version numbers: `[0-9]+.[0-9]+.[0-9]+`, so we can use tags without publishing docs for other stuff.

> Use the tags filter when you want to include tag name patterns or when you want to both include and exclude tag names patterns. 

from the [GitHub documentation
](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore)

### Testing Steps

I think we need to publish a test tag (0.0.1 maybe?) to test it, or just wait for the next release.

### DIFF
--- a/.github/workflows/dokka-github-pages.yml
+++ b/.github/workflows/dokka-github-pages.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - trunk
       - release/*
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+"
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:


hamorillo (approved):
> I think we need to publish a test tag (0.0.1 maybe?) to test it, or just wait for the next release.

What if we delete, for example, the `0.1.0` and create it again in the same commit?

maxme:
> What if we delete, for example, the 0.1.0 and create it again in the same commit?

0.1.0 won't have the updated `dokka-github-pages.yml` file, so it won't be triggered. I'll just try to tag a 0.2.1 and delete it after the test

maxme:
Running here: https://github.com/Automattic/Gravatar-SDK-Android/actions/runs/8602684881/job/23572794416

maxme:
Worked as expected: https://automattic.github.io/Gravatar-SDK-Android/0.2.1/index.html

I'll merge this one and delete the tag

hamorillo:
> 0.1.0 won't have the updated `dokka-github-pages.yml` file, so it won't be triggered. I'll just try to tag a 0.2.1 and delete it after the test

🤦🏼‍♂️🤦🏼‍♂️🤦🏼‍♂️

PR #98, opened 2024-04-10
Let ExpandableText to accept a custom dialog content

### LINKED ISSUE #92: Pass a composable function as a parameter to ExpandableText to allow for deeper customization

hamorillo:
Closes #92 

### Description

Let the `ExpandableText` composable to accept custom dialog content. That way, the content that should be shown when the `ExpandableText` is clicked can be customized. 

Also, it allows the SDK client to not show anything if they want to prevent the dialog.

Alt Text

**Open discussion**
With the actual changes, it will only be useful if someone uses the atomic components. If we decide to work on the customization of the components, adding some parameters to the ProfileCards could be interesting to let the dialogs customizations. 

### Testing Steps

- Code review
- Play with `ExpandableTextPreviewWithDialogContent`

### DIFF
--- a/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
@@ -1,11 +1,13 @@
 package com.gravatar.ui.components.atomic
 
 import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.wrapContentSize
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.Card
+import androidx.compose.material3.Icon
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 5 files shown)


gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
```diff
@@ -37,25 +46,53 @@ internal fun ExpandableText(text: String, modifier: Modifier = Modifier, maxLine
...
-                Text(
-                    text = text,
-                    modifier = Modifier
-                        .padding(16.dp)
-                        .fillMaxWidth()
-                        .wrapContentSize(),
-                    textAlign = TextAlign.Start,
-                )
+            if (dialogContent != null) {
+                dialogContent()
```
maxme: I know we can write something like:
```
    val text = "Text that \ncan be expanded \nby clicking on it."
    ExpandableText(text) {
        Card { Text(text) }
    }
```
But it seems more natural to me to change the composable signature to `((String) -> Unit)?` and pass the `text` here: `dialogContent(text)`? I would expect the "full string" to be displayed somewhere, and so we could just write this:

```
    ExpandableText("Text that \ncan be expanded \nby clicking on it.") {
        Card { Text(it) }
    }
```

maxme (approved):
LGTM, with one comment about the composable signature we'll pass to ExpandableText.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
hamorillo: Sounds good! I've updated the code. Let me know what you think.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
```diff
@@ -37,25 +46,54 @@ internal fun ExpandableText(text: String, modifier: Modifier = Modifier, maxLine
...
-                Text(
-                    text = text,
-                    modifier = Modifier
-                        .padding(16.dp)
-                        .fillMaxWidth()
-                        .wrapContentSize(),
-                    textAlign = TextAlign.Start,
-                )
-            }
+            dialogContent?.let { dialogContent(text) }
```
maxme: No need to update that but TIL we can write ` dialogContent?.invoke(text)`

maxme (approved):
:shipit:

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ExpandableText.kt
hamorillo: 🤦🏼‍♂️ thanks!

PR #102, opened 2024-04-11
Enable strict mode in `gravatar-ui`module 

### LINKED ISSUE #101: `gravatar-ui`module: Enable strict mode
We need to enable the strict mode for the `gravatar-ui` module.

hamorillo:
Closes #101 

### Description

When we created the `gravatar-ui` module, we forgot to set up the strict mode that forces us to specify the visibility in classes and methods. Similar to what we did in #57, we want to enable the strictApiMode in the UI module. 

In addition, I'm enabling the [OutdatedDocumentation](https://detekt.dev/docs/rules/comments/#outdateddocumentation) rule in Detekt in order to simplify the process of keeping the code documentation up-to-date.

### Testing Steps

- Remove a visibility identifier from a class and verify you receive an error when building the project ([Example](https://github.com/Automattic/Gravatar-SDK-Android/blob/5a229e61a056fb93c54fe9b7f2618326c4e11038/gravatar-ui/src/main/java/com/gravatar/ui/components/MiniProfileCard.kt#L32))

- Remove or modify any comment to make the documentation don't fit with the actual method signature ([Example](https://github.com/Automattic/Gravatar-SDK-Android/blob/5a229e61a056fb93c54fe9b7f2618326c4e11038/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt#L19)). Execute Detetk and verify you receive an error.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -62,3 +62,9 @@ comments:
     # TODO - Remove the explicit exclusion of Email.kt once we implement the new REST API
     excludes: [ '**/test/**', '**/androidTest/**', '**/demoapp/**', '**/models/Email.kt' ]
     searchProtectedProperty: false
+  OutdatedDocumentation:
+    active: true
+    excludes: [ '**/test/**', '**/androidTest/**', '**/demoapp/**', '**/models/**' ]
+    matchTypeParameters: true
+    matchDeclarationsOrder: true
+    allowParamOnConstructorProperties: false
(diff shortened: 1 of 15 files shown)


maxme (approved):
I did some changes to the ProfileCard to check if the error were caught and it works as expected:
```
ProfileCard.kt:40:1 Visibility must be specified in explicit API mode
ProfileCard.kt:39:12: Documentation of ProfileCard is outdated [OutdatedDocumentation]
```

LGTM, :shipit:

PR #103, opened 2024-04-11
Drop unused ProfileBackground field from the models

maxme:
### Description

The `profileBackground` field returns an object or an array depending on which account we get the account from (likely a bug on the backend). We don't use that field in the library yet and we'll likely switch to a new endpoint soon, so we'll just drop that field for now.

### Testing Steps

- Run the demo app, do some tests on the Profile tab, check the getProfile requests go through


hamorillo (approved):
:shipit:

PR #105, opened 2024-04-11
`gravatar-ui`- UI Component -> LargeSummaryProfile

### LINKED ISSUE #100: UI Component - LargeProfileSummaryView
image

hamorillo:
Closes #100 

### Description

Implementing the `LargeSummaryProfile` component using atomic components.

image

This component includes the `ViewProfile` atomic component but with a small change. As you can see in the design, in this case, we don't want to show the arrow. For that reason, I've modified the `ViewProfileButton.kt` to accept via param the `@Composable` we want to show using the arrow by default.

image

### Testing Steps

Check that the new component looks as per designs.


maxme (approved):
LGTM :shipit: 

Not related to this PR: I realized there is an issue with the `MiniProfileCard` when the location is empty, I'll open a ticket for it.

PR #106, opened 2024-04-11
Check that OpenAPI generation works as expected in CI

### LINKED ISSUE #90: Run OpenAPI generation in CI
Related to #85 

**Note: we should run OpenAPI generation in CI just to check generation works fine.**

maxme:
Closes #90

### Description

Check that OpenAPI generation works as expected in CI

### Testing Steps

Wait for CI and check openapi generation ran correctly

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -49,3 +49,10 @@ steps:
     notify:
       - github_commit_status:
           context: "Publish :gravatar"
+
+  # Run it after building step so the generated files are not bundled with the release
+  - label: "OpenAPI generation"
+    command: |
+      echo "--- ⚒️ Generating OpenAPI code"
+      ./gradlew :gravatar:openApiGenerate
+


hamorillo:
My only concern about this is that (If I'm right) we'll generate the code and, after that, the release, so for the release generation, we'll use the CI-generated code instead of the code we've committed. 

Indeed, it shouldn't be any problem because when we update the `yaml`, we regenerate the models. However, it may be better to avoid using the CI-generated code to ensure it is the same as what we've used on our machines.  🤷🏽‍♂️

maxme:
> My only concern about this is that (If I'm right) we'll generate the code and, after that, the release, so for the release generation, we'll use the CI-generated code instead of the code we've committed.

Yep, I can move the step after the build or even after the publishing step. This makes me wonder if we should make this new step fails in case the generated files are different than the one in the repo (that would mean we had changed the yaml without generating the files).

hamorillo:
> Yep, I can move the step after the build or even after the publishing step. 

Moving after the publishing step sounds good to me.

> This makes me wonder if we should make this new step fails in case the generated files are different than the one in the repo (that would mean we had changed the yaml without generating the files).

To be honest, I don't know what to say. 🤔 Maybe it looks more like a warning ⚠️? If we decide to pick the `yaml` from the public source instead of keeping it in the repo, it could make the build fail unexpectedly when they update the API. (Or not if we manually choose a concrete version of the API). I am not sure if it will bring us many benefits right now. 🤷🏽‍♂️  Anyway, I'm not saying it's a bad idea. I don't have a strong opinion. I'll try to think more about it.

maxme:
This is ready for review, I moved the step after publishing

hamorillo (approved):
:shipit:

PR #108, opened 2024-04-16
Don't show Location if it's empty

### LINKED ISSUE #107: Empty line on Mini Profile Card when the location is null or blank
Screenshot 2024-04-16 at 08 34 50

maxme:
Closes #107 

### Description

Don't show the Location component when the currentLocation is null or blank. I was thinking about moving the check in Location or ExpandableText components directly, but that might be unsettling for a third party developers to see the component disappear if it's empty.

Before | After
-|-
Screenshot 2024-04-16 at 08 34 50|Screenshot 2024-04-16 at 08 46 27

### Testing Steps

Run the demo app, go to the profile tab, check that a profile with current location unset shows the

### DIFF
--- a/gravatar-ui/src/main/java/com/gravatar/ui/components/MiniProfileCard.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/components/MiniProfileCard.kt
@@ -40,13 +40,15 @@ public fun MiniProfileCard(profile: UserProfile, modifier: Modifier = Modifier) 
             ProvideTextStyle(TextStyle(fontSize = 18.sp, fontWeight = FontWeight.Bold, lineHeight = 24.sp)) {
                 DisplayName(profile)
             }
-            ProvideTextStyle(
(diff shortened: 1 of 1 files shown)


hamorillo (approved):
I think we can face more problems like that for the other profile components, not only with the Location field but with all of them because they are all optional.

I wonder if we should open the discussion with the designer to evaluate how these components should work. I know it can be a pain because we use several fields, and most of them can be `null` (and I guess it will be the same with the new API). We also need to have the same behavior in the iOS SDK.

Anyway, it's not directly related to this PR so:

LGTM! :shipit:

maxme:
> I wonder if we should open the discussion with the designer to evaluate how these components should work.

I agree, and it's not just with null/empty states, same for too long/too large text or other limit cases. Example: I think we should limit the display name to one or 2 lines:
Screenshot 2024-04-17 at 09 02 38

PR #110, opened 2024-04-17
Demo - use coroutines instead of callbacks

maxme:
### Description

I'm tempted to keep both version of the "suspend" functions and the classic methods with the callbacks. If we just write the suspend versions, it's a huge pain for Java devs to integrate the library.

We could rename the callback versions `fun fetchWithListener(Email, Listenerer)` and set the suspend version as the default, eg. `suspend fun fetch(Email) : UserProfiles`

### Testing Steps

Run the demo app, make sure you can fetch profiles (and check the loading indicator and error works as expected)

### DIFF
--- a/gravatar/build.gradle.kts
+++ b/gravatar/build.gradle.kts
@@ -75,7 +75,7 @@ android {
 }
 
 dependencies {
-    implementation("com.squareup.okhttp3:okhttp:4.12.0")
+    api("com.squareup.okhttp3:okhttp:4.12.0")
     implementation("com.squareup.retrofit2:retrofit:2.9.0")
     implementation("com.squareup.retrofit2:converter-gson:2.9.0")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.7.3")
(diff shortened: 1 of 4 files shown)


maxme:
I'm not opening this one yet for merging, but let's open the discussion about doing this or not.

I started to write the "Get started" documentation and having suspend function makes things look much better, eg we could have this sample component that basically does everything (fetch and display a profile):

```kotlin

@Composable
fun SimpleGravatarProfileIntegration(emailAddress: String = "gravatar@automattic.com") {
    // Create a ProfileService instance
    val profileService = ProfileService()

    // Create a mutable state for the user profile
    var profile: UserProfile by remember { mutableStateOf(UserProfile(""), neverEqualPolicy()) }

    // We wrap the fetch call in a LaunchedEffect to fetch the profile when the composable is first launched, but this
    // could be triggered by a button click, a text field change, etc.
    LaunchedEffect(emailAddress) {
        try {
            // Fetch the user profile
            profile = profileService.fetchSuspend(Email(emailAddress).hash().toString()).entry.first()
        } catch (exception: ProfileService.FetchException) {
            // An error can occur when a profile doesn't exist, if the phone is in airplane mode, etc.
            // Here we log the error, but ideally we should show an error to the user.
            Log.e("Gravatar", exception.errorType.name)
        }
    }
    // Show the profile as a ProfileCard
    ProfileCard(profile)
}
```

gravatar/build.gradle.kts
```diff
@@ -75,7 +75,7 @@ android {
 }
 
 dependencies {
-    implementation("com.squareup.okhttp3:okhttp:4.12.0")
+    api("com.squareup.okhttp3:okhttp:4.12.0")
```
maxme, line 78: I have another PR for this: https://github.com/Automattic/Gravatar-SDK-Android/pull/128

maxme: Merged

maxme:
Closing this in favor of https://github.com/Automattic/Gravatar-SDK-Android/pull/130

PR #111, opened 2024-04-17
`gravatar-ui`- Refactor text styling

hamorillo:
Closes #

### Description

In this PR, I'm removing the use of `ProvideTextStyle` in our Profile components by adding a `textStyle` param in the atomic components.

I think this makes it easier to customize the atomic components and also has a `Default` style for the atomic ones. 

Another point about `ProvideTextStyle` is that it changes the `LocalTextStyle` for all the `@Composables` below it. This issue is easy to see in our `ExpandableText`:

image

The dialog text uses the style defined in the `ProvideTextStyle`.

Providing a `TextStyle` as a param with a default value (that should be a _GravatarStyle_) has fewer side effects and makes it more understandable to me. 

**Note:** This is a proposal to discuss; it may have some cons that I'm not seeing right now. 

### Testing Steps

- Run the DemoApp and verify everything in the Profile tab looks as expected.

| Before | After |
|----------|-----------|
|old|new|


maxme (approved):
LGTM, :shipit:

PR #112, opened 2024-04-17
`gravatar-ui` - Implement GravatarTheme

hamorillo:
### Description

This PR iterates over #111.

It would be nice to have our own `GravatarTheme` so that we can keep our Gravatar style in all UI components, at least for the default state.

Developers using the SDK could override our Theme if they don't want to use it. Using `CompositionLocal`, they'll be able to provide another implementation of GravatarTheme using whatever they want. [More info](https://developer.android.com/develop/ui/compose/compositionlocal)

Right now, we are using the Material 3 base styles, but we plan to update those with our colors, typographies, and shapes.

With this approach, we can create our `GravatarTheme` and use it along all the UI components without too much effort. (apart from defining the Theme). However, we won't limit the customization possibilities.

**Note:** As the first PR #111, I'm totally open to discussing this approach. Because I may be missing something.

### Testing Steps

- Code review
- Run DemoApp and verify everything looks good

### DIFF
--- a/gravatar-ui/src/main/java/com/gravatar/ui/GravatarTheme.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/GravatarTheme.kt
@@ -1,0 +1,57 @@
+package com.gravatar.ui
+
+import androidx.compose.material3.ColorScheme
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Shapes
+import androidx.compose.material3.Typography
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.staticCompositionLocalOf
+
+/**
+ * [GravatarTheme] is a composable that wraps the content of the application with the Gravatar theme.
+ */
+@Composable
+public fun GravatarTheme(content: @Composable () -> Unit) {
+    MaterialTheme(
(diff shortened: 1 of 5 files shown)


gravatar-ui/src/main/java/com/gravatar/ui/GravatarTheme.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    public val shapes: Shapes
+        @Composable
+        get() = MaterialTheme.shapes
+}
+
+/**
+ * [LocalGravatarTheme] is a CompositionLocal that provides the current [GravatarTheme].
+ */
+public val LocalGravatarTheme: ProvidableCompositionLocal<GravatarTheme> =
```
hamorillo, line 51: This allows any developer to replace the default `GravatarTheme` with their implementation. For example, they can replace everything with their own Theme, applying their own color, typographies, and shapes.

```
CompositionLocalProvider(LocalGravatarTheme provides object : GravatarTheme {
        @Composable
        get() = MaterialTheme.colorScheme.copy(outline = Color.Red)
}) {

```

image

maxme: Pretty cool!

maxme (approved):
LGTM, :shipit:

PR #116, opened 2024-04-17
Remove ExpandableText and replace it uses with a Text composable

### LINKED ISSUE #113: `gravatar-ui` - Remove expandable components from `Location` and `AboutMe`
Based on previous discussion:

> the user can see the entire text if they visit the profile from the link below.

The text should be truncated, but we shouldn't do anything more.

hamorillo:
Closes #113 

### Description

Based on discussion we don't want to expand any of the information inside the profile components, so we no longer need the ExpandableText. If you want to see more, you should tap the View More button.

| Before | After |
|----------|-----------|
|image|Alt Text|

### Testing Steps

- Launch the demo app and verify everything looks like expected.
- Verify that tapping on an overflowed `AboutMe`, `Location` or `UserInfo` don't do anything.


maxme (approved):
LGTM, :shipit:

PR #119, opened 2024-04-19
Fix #117: lower minsdkversion

### LINKED ISSUE #117: Lower minSdkVersion to 23 (or even lower)
Lower is better, we should check if we use any recent call (I doubt it).

maxme:
Closes #117

### Description

I picked 21 as minSdk for `gravatar-ui` because the maximum minSdk on our dependencies is 21 (`coil` minSdk is 21).

I wanted to set 14 as minimum to the `gravatar` module, but it's creating build errors with `mockk`. We could override the manifest merger (especially since it's only for test builds) but for now I think 21 as minSdk is fine.

I also updated the minSdk version of the demo-app as it was easier to test build limits.

Note: I also updated the demo app namespace in this PR as it was causing build warnings.

### DIFF
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -28,7 +28,7 @@ android {
 
     defaultConfig {
         applicationId = "com.gravatar.demoapp"
-        minSdk = 24
+        minSdk = 21
         targetSdk = 34
         versionCode = 1
         versionName = "1.0"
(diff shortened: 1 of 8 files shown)


maxme:
For reference: https://developer.android.com/build/publish-library/prep-lib-release#choose-minsdkversion

hamorillo (approved):
In my opinion, API 21 is enough. I don't think many apps use `minSdk` lower than that. Also, [Jetpack Compose requires](https://developer.android.com/develop/ui/compose/setup#create-new) at least minSDK 21.

:shipit:

PR #120, opened 2024-04-23
Trigger documentation generation on GitHub release publishing

maxme:
### Description

We noticed an issue on documentation generation for tags, `github.ref_name` is sometimes set to `trunk` on tags, eg. documentation for 0.2.0 was generated under `/trunk/` instead of `/0.2.0/`.

### Testing Steps

We'll need to merge this and then create and publish a test release (eg. 0.2.1) to check if the `github.ref_name` is correctly set.

### DIFF
--- a/.github/workflows/dokka-github-pages.yml
+++ b/.github/workflows/dokka-github-pages.yml
@@ -7,6 +7,8 @@ on:
       - release/*
     tags:
       - "[0-9]+.[0-9]+.[0-9]+"
+  release:
+    types: [published]
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:


.github/workflows/dokka-github-pages.yml
```diff
@@ -7,6 +7,8 @@ on:
       - release/*
     tags:
       - "[0-9]+.[0-9]+.[0-9]+"
```
hamorillo, line 9: I wonder if this is still necessary.

hamorillo (approved):

.github/workflows/dokka-github-pages.yml
maxme: Yep we can remove that if the release trigger works as expected. I think it's a bug on github side, I've found other project having issues with `ref_name` being unset or wrong.

I also found this issue https://github.com/actions/runner/issues/2788 which seems to affect the release trigger as well :/

PR #123, opened 2024-04-23
`gravatar-ui`- First iteration of the Ghost UI

### LINKED ISSUE #109: `gravatar-ui` - Shimmer effect
image

hamorillo:
Closes #109 

### Description

We want the Gravatar UI components to have a "loading" state. That state should have a skeleton animation of 800ms.

SkeletonEffect contains the base code to be used along the different components to create their "loading" state.

All modified components can be used in two different ways:

- Passing a profile as a parameter (Previous way of working)
- Passing a UserProfileLoadingState. This way, we'll show a skeleton to indicate when the profile is loading. When the state changes to "Loaded", the composable will be updated to show the already loaded state.

image

I'll apply the same pattern to the other components in the next PR.

### Testing Steps
- Run the different previews in a device
- Run the demo app and verify how the skeleton looks for the MiniProfileCard

https://github.com/Automattic/Gravatar-SDK-Android/assets/6974554/9bc63de8-523b-4ad7-9e06-26d4b5ecc3a7

### DIFF
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
@@ -156,7 +156,7 @@ private fun GravatarTabs(
 private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwable?) -> Unit) {
     var email by remember { mutableStateOf(BuildConfig.DEMO_EMAIL, neverEqualPolicy()) }
     var hash by remember { mutableStateOf("", neverEqualPolicy()) }
-    var profile: UserProfileState? by remember { mutableStateOf(null, neverEqualPolicy()) }
+    var profileState: UserProfileState? by remember { mutableStateOf(null, neverEqualPolicy()) }
     var error by remember { mutableStateOf("") }
     val profileService = ProfileService()
     val keyboardController = LocalSoftwareKeyboardController.current
@@ -179,13 +179,13 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
             Button(
                 onClick = {
                     keyboardController?.hide()
-                    profile = UserProfileState.Loading
+                    profileState = UserProfileState.Loading
                     error = ""
                     profileService.fetch(
                         Email(email),
                         object : GravatarListener<UserProfiles> {
                             override fun onSuccess(response: UserProfiles) {
-                                profile = response.entry.firstOrNull()?.let { UserProfileState.Loaded(it) }
+                                profileState = response.entry.firstOrNull()?.let { UserProfileState.Loaded(it) }
                             }
 
                             override fun onError(errorType: ErrorType) {
@@ -201,14 +201,14 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
                 GravatarDivider()
                 LabelledText(R.string.gravatar_generated_hash_label, text = hash)
                 GravatarDivider()
-                if ((profile is UserProfileState.Loading)) {
+                if ((profileState is UserProfileState.Loading)) {
                     CircularProgressIndicator()
                 }
             }
             Spacer(modifier = Modifier.height(16.dp))
             // Show the profile card if we got a result and there is no error and it's not loading
-            if ((profile is UserProfileState.Loaded) && error.isEmpty() && profile != null) {
-                (profile as? UserProfileState.Loaded)?.let {
+            if ((profileState is UserProfileState.Loaded) && error.isEmpty() && profileState != null) {
+                (profileState as? UserProfileState.Loaded)?.let {
                     ProfileCard(
                         it.userProfile,
                         Modifier
@@ -220,7 +220,7 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
                 }
             }
             if (error.isEmpty()) {
-                profile?.let {
+                profileState?.let {
                     MiniProfileCard(
                         it,
                         Modifier
@@ -232,8 +232,8 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
                     Spacer(modifier = Modifier.height(16.dp))
                 }
             }
-            if ((profile is UserProfileState.Loaded) && error.isEmpty() && profile != null) {
-                (profile as? UserProfileState.Loaded)?.let {
+            if ((profileState is UserProfileState.Loaded) && error.isEmpty() && profileState != null) {
+                (profileState as? UserProfileState.Loaded)?.let {
                     LargeProfile(
                         it.userProfile,
                         Modifier
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
@@ -61,6 +61,7 @@ import com.gravatar.ui.components.LargeProfile
 import com.gravatar.ui.components.LargeProfileSummary
 import com.gravatar.ui.components.MiniProfileCard
 import com.gravatar.ui.components.ProfileCard
+import com.gravatar.ui.components.UserProfileLoadingState
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 
@@ -155,8 +156,7 @@ private fun GravatarTabs(
 private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwable?) -> Unit) {
     var email by remember { mutableStateOf(BuildConfig.DEMO_EMAIL, neverEqualPolicy()) }
     var hash by remember { mutableStateOf("", neverEqualPolicy()) }
-    var profiles by remember { mutableStateOf(UserProfiles(emptyList()), neverEqualPolicy()) }
-    var loading by remember { mutableStateOf(false) }
+    var profile: UserProfileLoadingState? by remember { mutableStateOf(null, neverEqualPolicy()) }
     var error by remember { mutableStateOf("") }
     val profileService = ProfileService()
     val keyboardController = LocalSoftwareKeyboardController.current
@@ -179,20 +179,18 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
             Button(
                 onClick = {
                     keyboardController?.hide()
-                    loading = true
+                    profile = UserProfileLoadingState.Loading
                     error = ""
                     profileService.fetch(
                         Email(email),
                         object : GravatarListener<UserProfiles> {
                             override fun onSuccess(response: UserProfiles) {
-                                profiles = response
-                                loading = false
+                                profile = response.entry.firstOrNull()?.let { UserProfileLoadingState.Loaded(it) }
                             }
 
                             override fun onError(errorType: ErrorType) {
                                 onError(errorType.name, null)
                                 error = errorType.name
-                                loading = false
                             }
                         },
                     )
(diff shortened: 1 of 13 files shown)


gravatar-ui/src/main/java/com/gravatar/ui/GravatarTheme.kt
```diff
@@ -49,7 +56,13 @@ public interface GravatarTheme {
...
  */
 public val LocalGravatarTheme: ProvidableCompositionLocal<GravatarTheme> =
-    staticCompositionLocalOf { object : GravatarTheme {} }
+    staticCompositionLocalOf {
+        object : GravatarTheme {
+            override val colorScheme: ColorScheme
+                @Composable
+                get() = if (isSystemInDarkTheme()) DarkColorScheme else LightColorScheme
+        }
+    }
```
hamorillo, line 65: Supporting Dark and Light mode in GravatarTheme

gravatar-ui/build.gradle.kts
```diff
@@ -21,6 +21,8 @@ android {
 
     defaultConfig {
         minSdk = 21
+        // targetSdkVersion has no effect for libraries. This is only used for the test APK
+        targetSdk = 34
```
hamorillo, line 25: Without this, we can't install the Previews on a device.

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -203,46 +201,56 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
...
-                Spacer(modifier = Modifier.height(16.dp))
-                LargeProfileSummary(
-                    profiles.entry.first(),
-                    Modifier
-                        .padding(8.dp)
-                        .background(MaterialTheme.colorScheme.surfaceContainer)
-                        .padding(start = 24.dp, end = 24.dp, top = 24.dp, bottom = 8.dp)
-                        .fillMaxWidth(),
-                )
+            if ((profile is UserProfileLoadingState.Loaded) && error.isEmpty() && profile != null) {
```
hamorillo: I'll iterate over this part of the code when all of the components have the skeleton.

maxme:
I was looking at the iOS interface for showing the loading state. Here is the equivalent PR on iOS https://github.com/Automattic/Gravatar-SDK-iOS/pull/221

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
maxme: Sounds good, care to open ticket to create new skeletons and update this?

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -179,20 +179,18 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
             Button(
                 onClick = {
                     keyboardController?.hide()
-                    loading = true
+                    profile = UserProfileLoadingState.Loading
```
maxme: I'm wondering if we should also replace the `error` by the `UserProfileLoadingState`, and maybe rename it to `UserProfileState`, we could set all the states in there: Empty (before loading or if there is no profile), Loading, Loaded, Error.

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
maxme: If we had a `UserProfileState` we could simplify this and the following code.

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -155,8 +156,7 @@ private fun GravatarTabs(
 private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwable?) -> Unit) {
     var email by remember { mutableStateOf(BuildConfig.DEMO_EMAIL, neverEqualPolicy()) }
     var hash by remember { mutableStateOf("", neverEqualPolicy()) }
-    var profiles by remember { mutableStateOf(UserProfiles(emptyList()), neverEqualPolicy()) }
-    var loading by remember { mutableStateOf(false) }
+    var profile: UserProfileLoadingState? by remember { mutableStateOf(null, neverEqualPolicy()) }
```
maxme: Calling the state `profile` is a bit confusing to me, especially when reading the integration code with components that aren't ready yet (eg. `it.userProfile`). Should we rename it to `profileState`?

maxme: Another thing if we put all the states in a `UserProfileState`, we could make `profileState` non-null since we would always be in an "existing" state (Empty, Loading, Loaded, Error)

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -203,46 +201,56 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
...
+                            .background(MaterialTheme.colorScheme.surfaceContainer)
+                            .fillMaxWidth()
+                            .padding(24.dp),
+                    )
+                    Spacer(modifier = Modifier.height(16.dp))
+                }
+            }
+            if (error.isEmpty()) {
+                profile?.let {
+                    MiniProfileCard(
```
maxme, line 224: Would be great to have something like: `MiniProfileCard(profileState, modifier)` (related to my previous comments about non-null `profileState : UserProfileState`)

gravatar-ui/src/main/java/com/gravatar/ui/GravatarTheme.kt
maxme: Nice 👍

gravatar-ui/src/main/java/com/gravatar/ui/SkeletonEffect.kt
```diff
@@ -0,0 +1,71 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.rememberTextMeasurer
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+@Composable
+internal fun Modifier.skeletonEffect(): Modifier {
```
maxme, line 31: Great way of animating the component 👍

gravatar-ui/src/main/java/com/gravatar/ui/SkeletonEffect.kt
```diff
@@ -0,0 +1,71 @@
...
+    val infiniteTransition = rememberInfiniteTransition(label = "skeletonTransition")
+    val backgroundColor = infiniteTransition.animateColor(
+        // ⚠️ We should add the colors to the theme and use them here.
+        if (isSystemInDarkTheme()) Color(0xFF6B6B6B) else Color(0xFFE5E7E9),
```
maxme: Can we replace this (and the following line) with something like:
```
Color(LocalContext.current.resources.getColor(R.color.skeleton_color1, LocalContext.current.theme))
```

gravatar-ui/src/main/java/com/gravatar/ui/SkeletonEffect.kt
```diff
@@ -0,0 +1,71 @@
...
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.rememberTextMeasurer
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+@Composable
+internal fun Modifier.skeletonEffect(): Modifier {
+    val infiniteTransition = rememberInfiniteTransition(label = "skeletonTransition")
+    val backgroundColor = infiniteTransition.animateColor(
+        // ⚠️ We should add the colors to the theme and use them here.
```
maxme, line 35: Ok, what prevents us moving these to colors.xml?

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
hamorillo: I've created a couple tickets: #126 , #127

gravatar-ui/src/main/java/com/gravatar/ui/SkeletonEffect.kt
```diff
@@ -0,0 +1,71 @@
...
+        if (isSystemInDarkTheme()) Color(0xFF6B6B6B) else Color(0xFFE5E7E9),
+        if (isSystemInDarkTheme()) Color(0xFF979797) else Color(0xFFF2F2F2),
+        animationSpec = infiniteRepeatable(
+            animation = keyframes {
+                durationMillis = 800
```
maxme: We should also move this to a `values.xml` file, that will allow third party to override it.

maxme (approved):
Amazing work on the skeleton modifier.

To be honest, I don't know if I like or dislike the way to manage the loading state of components. Wrapping the profile in a state lets us have a very simple interface for the components, but makes the component quite special and non-intuitive.

Do you know if there are other Jetpack components that manage the different states like that, I'm curious about other API design for managing states.

Last thing, I think we should unify the different states even if it's a bit early, this would simplify things even more.

gravatar-ui/src/main/java/com/gravatar/ui/SkeletonEffect.kt
hamorillo: Done [6370c0a](https://github.com/Automattic/Gravatar-SDK-Android/pull/123/commits/6370c0a53212cd332a1ad41bc6ecb98fa4389047)

gravatar-ui/src/main/java/com/gravatar/ui/SkeletonEffect.kt
hamorillo: I've moved the colors to `Color.kt`. In my opinion, the ideal option would be to add those colors to the Gravatar themes. As we are still working on the Gravatar Theme with designers, we can keep it as it is and refactor it when possible. 

Moving the colors to the theme has the benefit of allowing developers to override the whole theme.

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
hamorillo: Renamed to `profileState`. 

> Another thing if we put all the states in a UserProfileState, we could make profileState non-null since we would always be in an "existing" state (Empty, Loading, Loaded, Error)

I totally agree with this. In fact, that was my original idea when passing a state to the composable. Allow all possible cases. 

The thing is that we don't have all the states right now, so I think we'll be there when we iterate over it with the designs for each state. 

So, I think that is the right approach, but we'll need more iterations until we have all the states within `UserProfileState`.

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
hamorillo: See [this comment.](https://github.com/Automattic/Gravatar-SDK-Android/pull/123#discussion_r1577551898)

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
hamorillo: I mentioned in another comment, we need to understand how each component should behave with the error state. However, this is the correct approach according to me.

hamorillo:
Thanks for your thoughtful review, @maxme !

> To be honest, I don't know if I like or dislike the way to manage the loading state of components. Wrapping the profile in a state lets us have a very simple interface for the components, but makes the component quite special and non-intuitive.

In my mind, it sounded like a good approach because we can keep independent composable for each state. Imagine you only want to show the Loaded state in your app because you always have the user profile. Or, you don't want our skeleton. But, on the other hand, you have a component that can handle everything.

Anyway, I'm totally open to different approaches. If you have another idea, we can talk about it.

maxme:
> Thanks for your thoughtful review, @maxme !
> 
> > To be honest, I don't know if I like or dislike the way to manage the loading state of components. Wrapping the profile in a state lets us have a very simple interface for the components, but makes the component quite special and non-intuitive.
> 
> In my mind, it sounded like a good approach because we can keep independent composable for each state. Imagine you only want to show the Loaded state in your app because you always have the user profile. Or, you don't want our skeleton. But, on the other hand, you have a component that can handle everything.
> 
> Anyway, I'm totally open to different approaches. If you have another idea, we can talk about it.

Another way to do it is to add an extra State param to the components. This could be generic and used for other components: `ProfileCard(profile, state)`, `Avatar(hash, state)`, etc. It seems a bit more straight forward to me. With the limitation of having the state taking precedence over the data we pass (eg. if we pass the loading state with full profile data, we should show a loading component)

```kotlin
    var profile: UserProfile by remember { mutableStateOf(UserProfile()) }
    var state: State by remember { mutableStateOf(State.EMPTY) }

    .... whatever OnClick() {
        try {
            state = State.LOADING
            profile = profileService.fetchSuspend(Email(emailAddress)).entry.first()
            state = State.LOADED
        } catch (exception: ProfileService.FetchException) {
            state = State.ERROR // or State.EMPTY
        }
    }
    ProfileCard(profile, state) // should work whatever the state is, no need to check for null or empty profile
```

hamorillo:
> Another way to do it is to add an extra State param to the components. This could be generic and used for other components: `ProfileCard(profile, state)`, `Avatar(hash, state)`, etc. It seems a bit more straight forward to me. With the limitation of having the state taking precedence over the data we pass (eg. if we pass the loading state with full profile data, we should show a loading component)

Mmm I don't see why we want to have two different params, one for the state and another for the profile, when only one state will use it. It look much simpler to me with the sealed class. For example, If I want to render a `Loaded` state, I need the profile, but, what sense make to use an `Error` state and also pass a userProfile as a parameter?

My idea is to add two new states in `UserProfileState`:

```
public data object Error : UserProfileState()
public data object Empty : UserProfileState()
```

UI components won't never need to check is userProfile is null, they will render based on state (same as you mention) but we don't need a fake UserProfile. I said fake, because if you want to create a UserProfile you'll need a hash and maybe you don't have the hash yet but the UI should be in a loading state.

maxme:
So at then end we should have something like this, right?

```kotlin
    var profileState: UserProfileState by remember { mutableStateOf(UserProfileState.Empty) }

    .... whatever OnClick() {
        try {
            profileState = UserProfileState.Loading
            val profile = profileService.fetchSuspend(Email(emailAddress).hash().toString()).entry.first()
            profileState = UserProfileState.Loaded(profile)
        } catch (exception: ProfileService.FetchException) {
            profileState = UserProfileState.Error
        }
    }
    ProfileCard(profileState)
```

---

One thing I prefer with the other version is that it seems more straight forward and it's easier to apply to any kind of component (not just component that can take a UserProfile)

hamorillo:
> So at then end we should have something like this, right?

Exactly.

> One thing I prefer with the other version is that it seems more straight forward and it's easier to apply to any kind of component (not just component that can take a UserProfile)

🤔 What do you think about the following approach?
If we make the State generic, it could be applied to any kind of component (right now, all our components work with a `UserProfile`).

```
public sealed class State<out T>   {
    public data object Loading : State<Nothing>()
    public data class Loaded<T>(val loadedValue: T) : State<T>()
}
```

It will work for both:

```
@Composable
public fun MiniProfileCard(state: State<UserProfile>, modifier: Modifier = Modifier) {
```

```
@Composable
public fun AvatarWithHash(state: State<String>) {
```

maxme:
> 🤔 What do you think about the following approach?

Yes that works, but TBH still feels weird to me to have a "profileState" or a "stringState" parameter. 

Anyway this is just a detail, the 3 solutions we described will work. Let's ship the current version and if we need to create more components, we'll extend it later. I'll add this to the list of things to review before we ship 1.0.0.

PR #128, opened 2024-04-24
ProfileService is taking an optional OkHttpClient parameter so we need to make okhttp an _api_ dependency

maxme:
I noticed the issue when testing the library in an empty app.


hamorillo (approved):
I was able to reproduce the issue

image


Nice catch! :shipit:

PR #130, opened 2024-04-24
Use suspend function in ProfileService

### LINKED ISSUE #129: Use suspend methods for networking instead of callbacks/listener
Make sure there is a way to integrate them in Java code

maxme:
Closes #129

### Description

Add another way to fetch profile information using suspend functions. Make them the defaults names (`fetch()`)

### Testing Steps

Run the demo app, make sure you can fetch profiles, check the ghost UI and check errors work as expected.

### DIFF
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
@@ -52,6 +52,7 @@ import com.gravatar.demoapp.theme.GravatarDemoAppTheme
 import com.gravatar.demoapp.ui.components.GravatarEmailInput
 import com.gravatar.demoapp.ui.model.SettingsState
 import com.gravatar.services.ProfileService
+import com.gravatar.services.Result
 import com.gravatar.types.Email
 import com.gravatar.ui.components.LargeProfile
 import com.gravatar.ui.components.LargeProfileSummary
@@ -155,7 +156,7 @@ private fun ProfileCards(profileState: UserProfileState?, error: String) {
         .fillMaxWidth()
         .padding(24.dp)
     // Show the profile card if we got a result and there is no error and it's not loading
-    if ((profileState is UserProfileState.Loaded) && error.isEmpty() && profileState != null) {
+    if ((profileState is UserProfileState.Loaded) && error.isEmpty()) {
         (profileState as? UserProfileState.Loaded)?.let {
             ProfileCard(it.userProfile, defaultModifier)
             Spacer(modifier = Modifier.height(16.dp))
@@ -167,7 +168,7 @@ private fun ProfileCards(profileState: UserProfileState?, error: String) {
             Spacer(modifier = Modifier.height(16.dp))
         }
     }
-    if ((profileState is UserProfileState.Loaded) && error.isEmpty() && profileState != null) {
+    if ((profileState is UserProfileState.Loaded) && error.isEmpty()) {
         (profileState as? UserProfileState.Loaded)?.let {
             LargeProfile(it.userProfile, defaultModifier)
             Spacer(modifier = Modifier.height(16.dp))
@@ -215,15 +216,18 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
                 onClick = {
                     keyboardController?.hide()
                     scope.launch {
-                        try {
-                            error = ""
-                            profileState = UserProfileState.Loading
-                            profileState = profileService.fetch(Email(email)).entry.firstOrNull()?.let {
-                                UserProfileState.Loaded(it)
+                        error = ""
+                        profileState = UserProfileState.Loading
+                        when (val result = profileService.fetch(Email(email))) {
+                            is Result.Success -> {
+                                result.value.entry.firstOrNull()?.let {
+                                    profileState = UserProfileState.Loaded(it)
+                                }
                             }
-                        } catch (exception: ProfileService.FetchException) {
-                            onError(exception.errorType.name, null)
-                            error = exception.errorType.name
+                            is Result.Failure -> {
+                                onError(result.error.name, null)
+                                error = result.error.name
+                            }
                         }
                     }
                 },
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
@@ -13,7 +13,6 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material3.Button
-import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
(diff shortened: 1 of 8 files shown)


demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -152,14 +148,53 @@ private fun GravatarTabs(
...
 }
 
+@Composable
+private fun ProfileCards(profileState: UserProfileState?, error: String) {
+    val defaultModifier = Modifier
+        .background(MaterialTheme.colorScheme.surfaceContainer)
+        .fillMaxWidth()
+        .padding(24.dp)
+    // Show the profile card if we got a result and there is no error and it's not loading
+    if ((profileState is UserProfileState.Loaded) && error.isEmpty() && profileState != null) {
```
hamorillo:
```suggestion
    if ((profileState is UserProfileState.Loaded) && error.isEmpty()) {
```

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -152,14 +148,53 @@ private fun GravatarTabs(
...
+            Spacer(modifier = Modifier.height(16.dp))
+        }
+    }
+    if (error.isEmpty()) {
+        profileState?.let {
+            MiniProfileCard(it, defaultModifier)
+            Spacer(modifier = Modifier.height(16.dp))
+        }
+    }
+    if ((profileState is UserProfileState.Loaded) && error.isEmpty() && profileState != null) {
```
hamorillo:
```suggestion
    if ((profileState is UserProfileState.Loaded) && error.isEmpty()) {
```

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -79,7 +81,67 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
...
+    public fun fetchWithListenerByUsername(username: String, getProfileListener: GravatarListener<UserProfiles>) {
+        fetchWithListener(username, getProfileListener = getProfileListener)
+    }
+
+    /**
+     * Exception thrown when a fetch operation fails.
+     *
+     * @property errorType The type of error that occurred
+     */
+    public class FetchException(public val errorType: ErrorType) : Exception()
```
hamorillo: ❓ Should we generalize the `FetchException` for other endpoints if necessary?

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -79,7 +81,67 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
...
+        return suspendCoroutine {
+            fetchWithListener(
+                hashOrUsername,
+                object : GravatarListener<UserProfiles> {
+                    override fun onSuccess(response: UserProfiles) {
+                        it.resumeWith(Result.success(response))
+                    }
+
+                    override fun onError(errorType: ErrorType) {
+                        it.resumeWithException(FetchException(errorType))
```
hamorillo: ❓Have you considered returning the error instead of throwing an exception? Using a Sealed class, for example. Throwing an exception can crash the app if you don't have a try/catch surrounding it. As far as I've read, there is no way in Kotlin to force the exception to be caught, so I wonder if we should use another approach.

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -69,8 +71,8 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
      * @param hash The hash to fetch the profile for
      * @param getProfileListener The listener to notify of the fetch result
      */
-    public fun fetch(hash: Hash, getProfileListener: GravatarListener<UserProfiles>) {
-        fetch(hash.toString(), getProfileListener = getProfileListener)
+    public fun fetchWithListener(hash: Hash, getProfileListener: GravatarListener<UserProfiles>) {
```
hamorillo: I think we can remove the `WithListener` from the function names; it looks cleaner, but I do not have a strong opinion.

hamorillo:
Nice work! I left some comments. Probably [this](https://github.com/Automattic/Gravatar-SDK-Android/pull/130#discussion_r1579020468) is the most important to me.

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
maxme: I want to keep these public obviously but I don't want to promote their usage, I feel like keeping the same name might not help with that.

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
maxme: I don't have strong opinion on this. I read code from folks using a sealed class that embeds the result or the error under a Success/Failure subclass. But they need to pack and unpack the error which results in another abstraction. 

With exceptions, we could just drop the `handleError` call and forward the throwable that was created in the underlying layer instead of packing them into an ErrorType, eg. we could do something like:

```kotlin
try {
    profile = fetch()
} catch (GravatarException e) {
    // specific to gravatar (eg. unknown hash?)
    showError()
} catch (SocketTimeoutException e) {
    // recoverable network error, show a retry button or auto retry later on connectivity change
    showErrorAndRetryButton()
}
```

maxme: If I'm not mistaken, same logic with a sealed class code will look like:

```kotlin
var error = null
var profile = null
val result = fetch()
when (result) {
    is Success -> profile = result.data
    is Error -> error = result.error
}
when (error) {
    HASH_NOT_FOUND -> showError()
    TIMEOUT -> showErrorAndRetryButton()
}
```

hamorillo: > If I'm not mistaken, same logic with a sealed class code will look like:
Yes

I don't like exceptions because they are more complicated to follow. To me, it is more linear to deal with two different types of responses. Also, as a third-party developer, if a call `profile = fetch()`, I won't have any warning from the IDE when using an exception, which could end in the app crashing. However, packing the result in a sealed class force me to think about both possibilities, success and failure.

For me, it is not something negative; the need to handle the Error; maybe the opposite, we are preventing errors.

maxme: >  as a third-party developer, if a call profile = fetch(), I won't have any warning from the IDE when using an exception, 

Fair enough, I'll drop the exception and replace it with a `Result` class

maxme: Updated in [ff53990](https://github.com/Automattic/Gravatar-SDK-Android/pull/130/commits/ff539903db402693f97dd023078d772b5ea35490)

gravatar/src/main/java/com/gravatar/services/Result.kt
```diff
@@ -0,0 +1,30 @@
+package com.gravatar.services
+
+/**
+ * Class representing the result of a network operation.
+ */
+public sealed class Result<T, E> {
```
hamorillo, line 6: 🤩 Looks great! As at the moment `E` is always `ErrorType`, we can remove that. But I'm totally fine with the current implementation.

hamorillo (approved):
:shipit:

gravatar/src/main/java/com/gravatar/services/Result.kt
maxme: Yes, the idea is to have a generic Result class that could be used for anything. That's also why I updated the `GravatarListener` class, so it's just the same kind of interface but with a different taste.

PR #137, opened 2024-04-26
Simple Full Featured MiniProfileCard

### LINKED ISSUE #131: Full featured profile component including networking / state management / display
We should provide a set of full featured profile component that takes only an email, and does the networking, state management and display of the profile.

maxme:
Closes #131 

I wrote this for documentation purpose, but I'm thinking about changing it to something usable when we have empty and error states.


mlumeau:
I'm closing this PR because it is too out of date with the current codebase. We'll maybe revisit this in the future.

PR #142, opened 2024-04-30
Closes 133: fix dokka publishing for latest release

### LINKED ISSUE #133: Fix dokka publishing on github pages - only store the latest released version
Drop the link attempt from https://github.com/Automattic/Gravatar-SDK-Android/blob/64505a5252106438437aeb872bc4427e6f6c1b0f/.github/workflows/dokka-github-pages.yml#L39

maxme:
Closes #133

### Description

Replace symbolic link by a recursive copy action. After doing a release we'll have 2 urls with the same content:

- https://automattic.github.io/Gravatar-SDK-Android/x.x.x/
- https://automattic.github.io/Gravatar-SDK-Android/latest/

I'll drop f9f53afbdd868eb5500a37c4e5a8945c33647984 and merge this after the review 

### Testing Steps

Visit https://automattic.github.io/Gravatar-SDK-Android/latest and  https://automattic.github.io/Gravatar-SDK-Android/maxme/133-fix-dokka-publishing-only-for-latest-release/

### DIFF
--- a/.github/workflows/dokka-github-pages.yml
+++ b/.github/workflows/dokka-github-pages.yml
@@ -1,9 +1,7 @@
 name: Deploy Dokka docs with GitHub Pages
 
 on:
-  push:
-    tags:
-      - "[0-9]+.[0-9]+.[0-9]+"
+  push
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:
--- a/.github/workflows/dokka-github-pages.yml
+++ b/.github/workflows/dokka-github-pages.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Generate dokka docs
         run: ./gradlew dokkaHtmlMultiModule
       - name: Move generated docs to branch directory
-        run: mkdir -p ./_site/${{ github.ref_name }} && mv ./docs/dokka/* ./_site/${{ github.ref_name }}/ && ln -s ./_site/${{ github.ref_name }} ./_site/latest
+        run: mkdir -p ./_site/${{ github.ref_name }} && mv ./docs/dokka/* ./_site/${{ github.ref_name }}/ && cp -r ./_site/${{ github.ref_name }} ./_site/latest
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v3
 


hamorillo:
⚠️❓ I'm thinking about this approach, and I would say we have a problem when we release a patch for a previous version. For example, imagine we've released `1.0.0`, after that `1.1.0`. Soon after releasing that version, we notice there is a bug in `1.0.0` and we release a patch under `1.0.1`. In that case, `latest` will be `1.0.1` instead of `1.1.0`,  right?

maxme:
> ⚠️❓ I'm thinking about this approach, and I would say we have a problem when we release a patch for a previous version. For example, imagine we've released `1.0.0`, after that `1.1.0`. Soon after releasing that version, we notice there is a bug in `1.0.0` and we release a patch under `1.0.1`. In that case, `latest` will be `1.0.1` instead of `1.1.0`, right?

That's a good point, but I don't know how we can get the list of released versions from the github actions context. And so I don't know how to compare current version with other existing versions.

Maybe the short term solution is to only have the `trunk` version in `latest` and the longer term solution is simply use the history of releases under docs.gravatar.com.

hamorillo:
> Maybe the short term solution is to only have the trunk version in latest and the longer term solution is simply use the history of releases under docs.gravatar.com.

Sound good! 

On the other hand, I saw [Dokka allows Versioning](https://kotlin.github.io/dokka/1.6.0/user_guide/versioning/versioning/). You can see an example [here](https://github.com/Kotlin/dokka/tree/master/examples/gradle/dokka-versioning-multimodule-example). We'll need to commit the generated doc to the repo, but that way we'll be able to keep everything in Github pages. Wdyt?

image

maxme:
> We'll need to commit the generated doc to the repo, but that way we'll be able to keep everything in Github pages. Wdyt?

Yes that's another solution, we could also put the docs in `gh-pages` branch (the old way of doing github pages).

hamorillo:
> we could also put the docs in gh-pages branch

👌🏼

maxme:
Superseded by https://github.com/Automattic/Gravatar-SDK-Android/pull/149

PR #143, opened 2024-05-02
Run onSuccess and onError callbacks on the main instead of the main thread

maxme:
Fix an issue reported on WPAndroid: https://github.com/wordpress-mobile/WordPress-Android/issues/20739#issuecomment-2089853261

Run the `onSuccess` and `onError` callbacks of the AvatarService and ProfileService in the main threads. We're not doing any logic in there so I think it's fine to wrap the full `onSucces/onFailure` methods we get from Retrofit directly in the main thread.

Note: this branch is forked from 0.2.0.

We can either release a 0.2.1 (but we should target `release/0.2.1` instead of `trunk`) or a `0.3.0` with all the recent changes in trunk. Since 0.2.0, we focused on UI components that aren't used in WPAndroid and  there is no incompatible changes in current trunk and WPAndroid, so I lean toward releasing a 0.3.0.

In the process, I removed `dispatcherDefault` and `dispatcherIO` from `GravatarSdkContainer` as I found them very confusing in the services. I think they were only used for injecting a different thread in testing environments. There is another way to override the main thread (`Dispatchers.setMain(containerRule.testDispatcher)`), feels better to me.

cc @thomashorta

### DIFF
--- a/gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
+++ b/gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
@@ -4,6 +4,7 @@ import com.google.gson.GsonBuilder
 import com.gravatar.GravatarApiService
 import com.gravatar.GravatarConstants.GRAVATAR_API_BASE_URL
 import com.gravatar.GravatarConstants.GRAVATAR_BASE_URL
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import okhttp3.OkHttpClient
 import retrofit2.Retrofit
@@ -20,8 +21,10 @@ internal class GravatarSdkContainer private constructor() {
 
     private fun getRetrofitBaseBuilder() = Retrofit.Builder().baseUrl(GRAVATAR_BASE_URL)
 
+    val dispatcherMain: CoroutineDispatcher = Dispatchers.Main
     val dispatcherDefault = Dispatchers.Default
     val dispatcherIO = Dispatchers.IO
+
     private val gson = GsonBuilder().setLenient().create()
 
     /**
--- a/gravatar/src/main/java/com/gravatar/services/AvatarService.kt
+++ b/gravatar/src/main/java/com/gravatar/services/AvatarService.kt
@@ -22,7 +22,8 @@ public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
         const val LOG_TAG = "AvatarService"
     }
 
-    private val coroutineScope = CoroutineScope(GravatarSdkDI.dispatcherDefault)
+    // Run onResponse and onError callbacks on the main thread
+    private val coroutineScope = CoroutineScope(GravatarSdkDI.dispatcherMain)
 
     /**
      * Uploads a Gravatar image for the given email address.
@@ -61,7 +62,9 @@ public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
                 }
 
                 override fun onFailure(call: Call<ResponseBody>, t: Throwable) {
-                    handleError(t, gravatarUploadListener, coroutineScope)
+                    coroutineScope.launch {
+                        gravatarUploadListener.onError(t.error())
+                    }
                 }
             },
         )
--- a/gravatar/src/test/java/com/gravatar/GravatarSdkContainerRule.kt
+++ b/gravatar/src/test/java/com/gravatar/GravatarSdkContainerRule.kt
@@ -23,8 +23,10 @@ class GravatarSdkContainerRule : TestRule {
                 gravatarSdkContainerMock = mockk<GravatarSdkContainer>()
                 gravatarApiServiceMock = mockk<GravatarApiService>(relaxed = true)
                 mockkObject(GravatarSdkContainer)
-                every { GravatarSdkContainer.instance } returns gravatarSdkContainerMock
+                every { gravatarSdkContainerMock.dispatcherMain } returns testDispatcher
                 every { gravatarSdkContainerMock.dispatcherDefault } returns testDispatcher
+                every { gravatarSdkContainerMock.dispatcherIO } returns testDispatcher
+                every { GravatarSdkContainer.instance } returns gravatarSdkContainerMock
                 every { gravatarSdkContainerMock.getGravatarApiService(any()) } returns gravatarApiServiceMock
 
                 base.evaluate()
--- a/gravatar/src/main/java/com/gravatar/services/ErrorUtils.kt
+++ b/gravatar/src/main/java/com/gravatar/services/ErrorUtils.kt
@@ -1,18 +1,12 @@
 package com.gravatar.services
 
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.launch
 import java.net.SocketTimeoutException
 import java.net.UnknownHostException
 
-internal fun handleError(t: Throwable, listener: GravatarListener<*, ErrorType>, coroutineScope: CoroutineScope) {
-    val error: ErrorType =
-        when (t) {
-            is SocketTimeoutException -> ErrorType.TIMEOUT
-            is UnknownHostException -> ErrorType.NETWORK
-            else -> ErrorType.UNKNOWN
-        }
-    coroutineScope.launch {
-        listener.onError(error)
+internal fun Throwable.error(): ErrorType {
+    return when (this) {
+        is SocketTimeoutException -> ErrorType.TIMEOUT
+        is UnknownHostException -> ErrorType.NETWORK
+        else -> ErrorType.UNKNOWN
     }
 }
--- a/gravatar/src/main/java/com/gravatar/services/ProfileService.kt
+++ b/gravatar/src/main/java/com/gravatar/services/ProfileService.kt
@@ -22,7 +22,8 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
         const val LOG_TAG = "ProfileService"
     }
 
-    private val coroutineScope = CoroutineScope(GravatarSdkDI.dispatcherDefault)
+    // Run onResponse and onError callbacks on the main thread
+    private val coroutineScope = CoroutineScope(GravatarSdkDI.dispatcherMain)
 
     private fun fetchWithListener(
         hashOrUsername: String,
@@ -52,7 +53,9 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
                 }
 
                 override fun onFailure(call: Call<UserProfiles>, t: Throwable) {
-                    handleError(t, getProfileListener, coroutineScope)
+                    coroutineScope.launch {
+                        getProfileListener.onError(t.error())
+                    }
                 }
             },
         )


gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
```diff
@@ -20,8 +19,6 @@ internal class GravatarSdkContainer private constructor() {
 
     private fun getRetrofitBaseBuilder() = Retrofit.Builder().baseUrl(GRAVATAR_BASE_URL)
 
-    val dispatcherDefault = Dispatchers.Default
-    val dispatcherIO = Dispatchers.IO
```
thomashorta, line 24: I think the approach here was fine for the reason you mentioned before (testing) and making it easy to inject the correct Dispatcher at runtime.

Moreover,  injecting Dispatchers instead of hardcoding them is actually a [best practice for coroutines](https://developer.android.com/kotlin/coroutines/coroutines-best-practices) so I'm against removing the injection that existed before this PR. I suggest reverting this change and creating another dependency provider here for `dispatcherMain = Dispatchers.Main`.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -22,7 +23,8 @@ public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
         const val LOG_TAG = "AvatarService"
     }
 
-    private val coroutineScope = CoroutineScope(GravatarSdkDI.dispatcherDefault)
+    // Run onResponse and onError callbacks on the main thread
+    private val coroutineScope = CoroutineScope(Dispatchers.Main)
```
thomashorta: I'm not 100% sure what's the best default Dispatcher to use in this Coroutine Scope, but maybe you're fine with using what you had before (`GravatarSdkDI.dispatcherDefault`) and just making sure to switch context whenever it makes sense (see comments below for more details/examples).

thomashorta: Since it looks like I can't add review comments in code that was not changed I will add some stuff related to coroutine context switching here:

### Asynchronous work
I just noticed that `service.uploadImage` is not being called in a coroutine, which looks like an error, as it's probably the longest-running operation in this function. If that should indeed be running inside a coroutine, I would change this function to actually launch a coroutine ASAP, before line 38:

```kotlin
coroutineScope.launch {
    // content from lines 38-64 here (aka the whole function body)
}
```

### IO work
In this class it looks like the `service.uploadImage` call highly relies on network communication, so it's suited for running on the IO Dispatcher (a dispatcher that is optimized for disk/network operations). BTW 

So your code could have this just before line 43, closing the block after the function call, at line 64 (considering the code is now running inside a coroutine):
```kotlin
withContext(GravatarSdkDI.dispatcherIO) {
    service.uploadImage(...).enqueue(...)
}
```

### Main-thread work
Likewise, operations that should run in the Main thread should also use `withContext` where applicable around the blocks that make sense, such as callback calls, which are _usually_ expected to run in the main thread, especially if we don't have information about where the calling code is running.

Based on that, the code for the `onSuccess` should launch a coroutine in the Main dispatcher, so line 46 should be:
```kotlin
coroutineScope.launch(GravatarSdkDI.dispatcherMain) {
    // code here for checking success and calling the appropriate callback
}
```

### Something odd about the `onFailure` code
I noticed that for the failure there's a `handleError` function that receives a `coroutineScope` as parameter, not sure exactly why. If the idea here is that `handleError` should _always_ be called from a coroutine, the best way you can do so is to convert it into a `suspend fun` and remove the `coroutineScope` argument and [the coroutine launch inside it](https://github.com/Automattic/Gravatar-SDK-Android/blob/d43fab05636329659b124b49f525b739200f415c/gravatar/src/main/java/com/gravatar/services/ErrorUtils.kt#L15), to enforce the caller having to launch a coroutine before calling `handleError`.

Then inside `handleError` you can change lines 15-17 to switch to the main context, so any code using this function will automatically use the correct Dispatcher for calling the `listener.onError`.

```kotlin
// inside ErrorUtils.kt, lines 15-17
withContext(GravatarSdkDI.dispatcherMain) {
    listener.onError(error)
}
```

You will also likely need to launch a coroutine around line 61:
```kotlin
coroutineScope.launch {
    // not really important to use the correct Dispatcher since the suspending handleError should take care
    handleError(t, gravatarUploadListener)
}
```

### More about using Dispatchers
You can find more about using Dispatchers and their differences [here](https://developer.android.com/kotlin/coroutines/coroutines-adv#main-safety).

gravatar/src/test/java/com/gravatar/GravatarSdkContainerRule.kt
```diff
@@ -24,7 +24,6 @@ class GravatarSdkContainerRule : TestRule {
                 gravatarApiServiceMock = mockk<GravatarApiService>(relaxed = true)
                 mockkObject(GravatarSdkContainer)
                 every { GravatarSdkContainer.instance } returns gravatarSdkContainerMock
-                every { gravatarSdkContainerMock.dispatcherDefault } returns testDispatcher
```
thomashorta, line 27: Nice! It's probably worth also mocking the other dispatchers to also use the `testDispatcher`, which is usually what you want in most tests.

thomashorta (changes requested):
Thanks for the changes! I reviewed it and found some things that go against some coroutine best practices and concepts so I took some time to add thorough comments about it and requested some changes.

Feel free to ping me if you have questions or if something doesn't make sense.

🙇🏼

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
maxme: > In this class it looks like the service.uploadImage call highly relies on network communication, so it's suited for running on the IO Dispatcher (a dispatcher that is optimized for disk/network operations). BTW

We're using the Retrofit library, in this case the call to `enqueue()` will run the network request asynchronously and return results in the `Callback<ResponseBody>` block when ready. File reading is also done asynchronously (via the `RequestBody` object).

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
maxme: From the document you linked:

> This dependency injection pattern makes testing easier as you can replace those dispatchers in unit and instrumentation tests with a [test dispatcher](https://developer.android.com/kotlin/coroutines/coroutines-best-practices#test-coroutine-dispatcher) to make your tests more deterministic.

Injecting dispatchers is a good practice for making testing easier, but using `Dispatchers.setMain` seems even easier to me. I wonder if they're other reasons though.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
maxme: We're only doing a comparison here, that's why I wrapped everything in under `Dispatchers.Main`, IMO it's not worth switching context in there.

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
eyedol: Just adding my 2 cents: Ignore if they don't help in anyway :)

> Injecting dispatchers is a good practice for making testing easier, but using Dispatchers.setMain seems even easier to me. I wonder if they're other reasons though.

I believe it's nice to have a central place to configure dispatchers. Perhaps in the event you might need a custom dispatcher. Probably not relevant to this project to define custom dispatchers but generally its nice to have a central place to do this kind of thing.

Also, as stated in the linked documentation, having your dispatchers centrally managed can make it easier to swap them out with a test dispatcher thus making the test more deterministic - Yeah, I know nothing new here, just wanted to rehash it again :)

gravatar/src/test/java/com/gravatar/AvatarServiceTest.kt
```diff
@@ -29,8 +32,10 @@ class AvatarServiceTest {
 
     private lateinit var avatarService: AvatarService
 
+    @OptIn(ExperimentalCoroutinesApi::class)
     @Before
     fun setUp() {
+        Dispatchers.setMain(containerRule.testDispatcher)
```
eyedol: Not sure if it makes sense to set this in `GravatarSdkContainerRule` Also it's a good practice to do a `Dispatchers.resetMain()` call after the test finishes running.

```kotlin
...
override fun evaluate() {
    try {
        gravatarSdkContainerMock = mockk<GravatarSdkContainer>()
        gravatarApiServiceMock = mockk<GravatarApiService>(relaxed = true)
        mockkObject(GravatarSdkContainer)
        every { GravatarSdkContainer.instance } returns gravatarSdkContainerMock
        every { gravatarSdkContainerMock.getGravatarApiService(any()) } returns gravatarApiServiceMock
        Dispatchers.setMain(testDispatcher)
        
        base.evaluate()
    } finally {
        Dispatchers.resetMain()
    }
}
...            
```

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
eyedol: Maybe you don't even need to launch a coroutine for the response callbacks as they seem to be ran in the correct thread. Per the [documentation](https://square.github.io/retrofit/2.x/retrofit/retrofit2/Callback.html) from retrofit's `Callback` interface:

> Android: Callbacks are executed on the application's main (UI) thread.
JVM: Callbacks are executed on the background thread which performed the request.

thomashorta: I didn't realize Retrofit was being used, my bad. In that case I agree with @eyedol, probably there's no need to launch a coroutine for any response callbacks (and maybe anything else in this class, since the thread management for the UI part will already be done by Retrofit).

gravatar/src/test/java/com/gravatar/AvatarServiceTest.kt
eyedol: Alternatively you can make a new reusable JUnit4 TestRule to set and reset the main dispatcher. See: https://developer.android.com/kotlin/coroutines/test#setting-main-dispatcher Specifically the sample class `MainDispatcherRule` in the link.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
hamorillo: Great discussion here! 

I see two options:

- Let Retrofit handle the thread management using callbacks and `enqueue()` calls and remove everything related to Dispatchers because we won't need it anymore with that implementation. 
- Use sync Retrofit implementation with `execute()` and handle everything internaly in the GravatarSDK. For example:
```
CoroutineScope(GravatarSdkDI.dispatcherIO).launch {
    service.uploadImage("Bearer $accessToken", identity, filePart).execute().let { response ->
        if (response.isSuccessful) {
            withContext(GravatarSdkDI.dispatcherMain) {
                gravatarUploadListener.onSuccess(Unit)
            }
        } else {
            // Log the response body for debugging purposes if the response is not successful
            Logger.w(
                LOG_TAG,
                "Network call unsuccessful trying to upload Gravatar: $response.body",
            )
            withContext(GravatarSdkDI.dispatcherMain) {
                gravatarUploadListener.onError(errorTypeFromHttpCode(response.code()))
            }
        }
    }
}
```

I would go with the second option because it gives us control over where and how things are executed. Also, we probably need to use coroutines for more things than network requests, so I think we'll need to handle the Dispatcher injection anyway.

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
thomashorta: > Injecting dispatchers is a good practice for making testing easier, but using Dispatchers.setMain seems even easier to me. I wonder if they're other reasons though.

Besides the nice central place for configuration like @eyedol mentioned, `setMain` only replaces the `Main` dispatcher and the function exists because many built-in coroutine functions and helpers use the main dispatcher implicitly (so we are not able to inject `testDispatcher` for example). 

If you are using other Dispatchers (such as `IO` and `Default`) `setMain` doesn't help you and you need a way of replacing it with a different dispatcher, mostly for testing, since using background threads during tests can lead to wrong behavior or just add unnecessary difficulty. That's where injecting via constructor or even DI frameworks is useful.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
thomashorta: Good point. Handling thread/dispatcher switching ourselves definitely gives us more flexibility and ways of optimizing the performance of some stuff. For instance, doing most processing in the `Default` dispatcher, network calls in `IO` dispatcher, and calling the callbacks in the `Main` dispatcher.

But if the idea is to use Coroutines for everything with Retrofit then I would suggest also using the coroutine-compatible Retrofit API. For instance, instead of having Services with functions that use `Call`s, replace them with `suspend` functions.

An example of how that would look like in the `GravatarApiService` (if I'm not mistaken):

```kotlin
// BEFORE
@Multipart
@POST("upload-image")
fun uploadImage(
    @Header("Authorization") authorization: String,
    @Part identity: MultipartBody.Part,
    @Part data: MultipartBody.Part,
): Call<ResponseBody>

// AFTER (with coroutines)
@Multipart
@POST("upload-image")
suspend fun uploadImage(
    @Header("Authorization") authorization: String,
    @Part identity: MultipartBody.Part,
    @Part data: MultipartBody.Part,
): Response<ResponseBody>
```

But now I think we might be going way beyond the scope of this PR and talking about general improvements to the codebase, which should be tackled separately.

thomashorta: Still regarding the usage of coroutines for more control over how things are executed, I just wanted to voice out a concern/opinion:

My main concern (and I might be being nitpicky) is about which Coroutine `Context` is being used, since the `CoroutineScope` is currently being created and managed by the class, without being attached to any lifecycle-aware component nor the calling code (e.g.: via a `suspend` function). With that, we lose the advantage of [structured concurrency](https://kotlinlang.org/docs/coroutines-basics.html#structured-concurrency).

This means that if external callers end up "dying" while the request is ongoing, we have a few issues:
1. the request is not dropped/ignored/canceled, using resources unnecessarily;
2. the callbacks might still be called and try to access a stale reference, since the Context is not inherited from the caller and therefore any launched `Job` will not be part of the coroutine job structure.

eyedol: Great inputs @hamorillo and @thomashorta. Makes sense to go all in with using coroutines as it gives control and yes better if this is handled as a separate PR as it will require a bit of rework.

I don't mind taking it as my first real coding contribution though :)

hamorillo: > I would suggest also using the coroutine-compatible Retrofit API. For instance, instead of having Services with functions that use Calls, replace them with suspend functions.

Yeah, it makes sense. In fact, as we are using the [OpenApi generator](https://github.com/Automattic/Gravatar-SDK-Android?tab=readme-ov-file#generating-api-code-from-openapi-definitions) is just a matter of changing one [configuration](https://openapi-generator.tech/docs/generators/kotlin#config-options).

image

> My main concern (and I might be being nitpicky) is about which Coroutine Context is being used, since the CoroutineScope is currently being created and managed by the class, without being attached to any lifecycle-aware component nor the calling code (e.g.: via a suspend function). 

I don't think you are nitpicking. That's an important point. We probably should declare our functions as `suspend`.

Anyway, I would use this PR to fix the issue detected in WPAndroid in a simple way (for example, by adding the Main dispatcher), and we can work on the refactor later.

maxme:
This PR diverged into an architecture discussion. I opened a new ticket  https://github.com/Automattic/Gravatar-SDK-Android/issues/144 to update OpenAPI generation to use the coroutines adapter and change the Service code accordingly. I'll re-introduce dispatcher injection and then we can merge this one to fix the WPAndroid issue.

hamorillo (commented):
LGTM! We need to solve the conflicts, and I'll go with 0.3.0. At this moment, I don't see the need to maintain two versions. When we release 1.0.0, we'll do it if needed.

hamorillo (approved):
:shipit:

PR #145, opened 2024-05-08
`gravatar-ui` - Empty state in UI components

### LINKED ISSUE #132: UserProfile State management: add support for empty
More on the PR discussion here: https://github.com/Automattic/Gravatar-SDK-Android/pull/123#issuecomment-2074893737

hamorillo:
Closes #132 

### Description

This PR adds the `empty state` in all the UI components.

| Light | Dark |
|----------|-----------|
|image|image|

### Testing Steps

Using the demo app, in the `Profile` tab, look for an email without a Gravatar account, (ex: `gravatarr@automattic.com`) and verify that the empty views look as expected. 

While reviewing the views switch between `light` / `dark` and `system` theme.

Verify that tapping over the `Gravatar` icon brings you to the `www.gravatar.com`.
Verify that tapping over the `Claim profile` or any other part of the profile components brings you to the `Sign-up` page.

Smoke test that loading and loaded states also looks as expected.

### DIFF
--- a/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
@@ -13,8 +13,10 @@ import coil.compose.AsyncImage
 import com.gravatar.AvatarQueryOptions
 import com.gravatar.api.models.UserProfile
 import com.gravatar.extensions.avatarUrl
+import com.gravatar.ui.R
 import com.gravatar.ui.components.LoadingToLoadedStatePreview
 import com.gravatar.ui.components.UserProfileState
+import com.gravatar.ui.components.isNightModeEnabled
 import com.gravatar.ui.skeletonEffect
 
 /**
@@ -33,13 +35,22 @@ public fun Avatar(
     avatarQueryOptions: AvatarQueryOptions? = null,
 ) {
     val preferredSize = with(LocalDensity.current) { size.roundToPx() }
-    AsyncImage(
-        profile.avatarUrl(
+    Avatar(
+        model = profile.avatarUrl(
             // Override the preferredSize
             avatarQueryOptions?.copy(
                 preferredSize = preferredSize,
             ) ?: AvatarQueryOptions(preferredSize = preferredSize),
         ).url().toString(),
+        size = size,
+        modifier = modifier,
+    )
+}
+
+@Composable
+private fun Avatar(model: Any?, size: Dp, modifier: Modifier) {
+    AsyncImage(
+        model = model,
         contentDescription = "User profile image",
         modifier = modifier.size(size),
     )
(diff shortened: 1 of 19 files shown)


gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
```diff
@@ -77,6 +88,16 @@ public fun Avatar(
                 avatarQueryOptions = avatarQueryOptions,
             )
         }
+
+        UserProfileState.Empty -> Avatar(
+            model = if (isNightModeEnabled()) {
+                R.drawable.empty_profile_avatar_dark
+            } else {
+                R.drawable.empty_profile_avatar
+            },
```
hamorillo, line 97: At first I tried using the folder `drawable-night` and it works well when you switch between light and dark modo in the device. However, we you override the `uiMode` as [we do in the demo app](https://github.com/Automattic/Gravatar-SDK-Android/blob/d43fab05636329659b124b49f525b739200f415c/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt#L172), it still shows the drawable for the device theme. 

For that reason I've written this workaround.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
```diff
@@ -50,11 +51,28 @@ public fun ViewProfileButton(
...
+        stringResource(R.string.view_profile_button),
+        profile.profileUrl().url.toString(),
+        textStyle,
+        modifier,
+        inlineContent,
+    )
+}
+
+@Composable
+private fun ViewProfileButton(
```
maxme, line 65: I think we can rename that to something more generic, since we can change the text and the url, this button can do other actions.

That reminded me, we should create an "EditProfileButton" component as well (probably needed for Pocket Casts integration).

maxme (approved):
LGTM :shipit: 

One comment though:

> Verify that tapping over the `Gravatar` icon and `Claim profile` brings you to the `www.gravatar.com`.

This works fine, but from the design document, I see:

> Clicking on the Gravatar logo would take the user to Gravatar.com.
> Clicking anywhere else on the Profile Card will take the user directly to the Sign-up flow.

I'm not sure if this was discussed elsewhere and updated. If we have to, we can create another ticket for this.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
hamorillo: When using the SDK, you can modify the text but not the URL. What about `ProfileButton` which, can include `ClaimProfile`, `EditProfile` and `NavigateToProfile`. I guess we can iterate over this.

hamorillo:
> This works fine, but from the design document, I see:
> 
> > Clicking on the Gravatar logo would take the user to Gravatar.com.
> > Clicking anywhere else on the Profile Card will take the user directly to the Sign-up flow.
> 
> I'm not sure if this was discussed elsewhere and updated. If we have to, we can create another ticket for this.

I've updated the behavior when clicking the empty profile components. I'm not completely sure If I like that empty / loaded states have different behaviors, but I guess we can discuss and iterate over it.

@maxme Could you give another review round, please? 🙏🏼

maxme (approved):
LGTM :shipit:

PR #146, opened 2024-05-10
`gravatar-ui`- Screenshot testing - Roborazzi

### LINKED ISSUE #114: `gravatar-ui`- Screenshot testing
Review and try to implement **a stable** set of screenshot tests.

hamorillo:
Closes #114 

### Description

This PR adds screenshot testing using [Roborazzi](https://github.com/takahirom/roborazzi). At the moment, screenshots have been added to the `gravatar-ui` module to verify that we don't break anything between iterations of the UI components.

If you execute the verification task, you'll find the report in:

`gravatar-ui/build/reports/roborazzi/index.html`

**Note:** We should add a placeholder to the `Avatar` component in order to prevent it from showing nothing. Another option could be adding the skeleton while the image is being loaded. I've tried but encountered some issues with the screenshots, so I'll leave it for another PR.

### Testing Steps

- Review CI and verify tests have been executed
- Execute locally the `record` and `verify` commands:
```
- Record -> ./gradlew recordRoborazziDebug
- Verify -> ./gradlew verifyRoborazziDebug
```
- Play with the test, trying to break something on purpose to verify it works.


.buildkite/pipeline.yml
```diff
@@ -26,10 +26,10 @@ steps:
       # `build` task would also run tasks, which are executed in a separate job
       ./gradlew assemble
 
-  - label: "Unit tests"
+  - label: "Unit and screenshot tests"
     command: |
-      echo "--- 🧪 Testing"
-      ./gradlew test
+      echo "--- 🧪📸 Testing"
+      ./gradlew verifyRoborazziDebug
```
hamorillo: All unit tests will be executed by the `verifyRoborazziDebug` task, not only screenshot ones.

gravatar-ui/src/test/java/com/gravatar/gravatar/ui/RoborazziTest.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.gravatar.ui.GravatarTheme
+import org.junit.Rule
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.GraphicsMode
+
+@RunWith(AndroidJUnit4::class)
+@GraphicsMode(GraphicsMode.Mode.NATIVE)
+@Config(qualifiers = RobolectricDeviceQualifiers.Pixel5)
+abstract class RoborazziTest {
```
hamorillo, line 21: All screenshot test classes should extend from `RoborazziTest`.

README.md
```diff
@@ -43,6 +43,20 @@ Run unit tests on your machine via the following command:
 ./gradlew test
```
maxme, line 43: Roborazzi tests fail when I run `./gradlew test`:

> com.gravatar.gravatar.ui.components.LargeProfileSummaryTest > profileLoadingLight FAILED
>    java.lang.RuntimeException at RoboMonitoringInstrumentation.java:105

We should change the README or exclude Roborazzi tests from being ran by `./gradlew test`. I think the second option is better.

.buildkite/pipeline.yml
maxme: I tried to break one of the test in `AvatarUrlTest` and then I ran `./gradlew verifyRoborazziDebug`, I got a `BUILD SUCCESSFUL in 6s`, so I think our unit tests are not run by the `verifyRoborazziDebug` task.

Note that `AvatarUrlTest` and other unit tests are under the `gravatar` module, not `gravatar-ui`.

README.md
maxme: We only have Roborazzi tests under `gravatar-ui`, so we could make a difference and tell users to run `:gravatar:test` for other unit test and `./gradlew :gravatar-ui:verifyRoborazziDebug` for checking screenshot tests.

maxme (changes requested):
We can't change the `.buildkite/pipeline.yml` that way as the unit tests from the `gravatar` module won't be run, we can eventually add a new task to run `./gradlew :gravatar:test` only

.buildkite/pipeline.yml
maxme: One way to update it is to change this to `./gradlew :gravatar:test ` and add a new task to run `./gradlew :gravatar-ui:verifyRoborazziDebug`

gradle.properties
```diff
@@ -24,3 +24,6 @@ android.nonTransitiveRClass=true
 
 org.gradle.caching=true
 org.gradle.configuration-cache=true
+
+# Roborazzi - https://github.com/takahirom/roborazzi?tab=readme-ov-file#q-how-can-i-execute-screenshot-tests-using-android-studios-run-button
+roborazzi.test.verify=true
```
hamorillo, line 29: This allows us to run the screenshot tests from the ID by tapping the `run` button.

hamorillo:
Great catch @maxme! Thank you for your in-depth review.

I've excluded all the screenshot tests from the default tests task. Now, when executing `./gradlew test` only non-screenshot tests will be executed. To be more specific, tests that are not annotated with `@ScreenshotTest`. 

If we want to execute screenshot tests, we need to add to the gradlew task the `-Pscreenshot` property. 

Ideally, I would like to remove the necessity of the `-Pscreenshot` property. I'll iterate over it, but I don't think it's a blocker right now.

Wdty?

maxme:
> Ideally, I would like to remove the necessity of the `-Pscreenshot` property. I'll iterate over it, but I don't think it's a blocker right now.

That's fine, if we use them mostly for catching regressions, it will be fine if it's correctly set up in CI.

maxme (approved):
- `./gradlew :gravatar-ui:verifyRoborazziDebug -Pscreenshot` work as expected and run the screenshot tests.
- `./gradle test` work as expected, run the unit test and doesn't run the screenshot tests.


LGTM - :shipit:

PR #151, opened 2024-05-16
Refactor GravatarServices to expose suspend functions

### LINKED ISSUE #144: Update OpenAPI generation to use the coroutine adapter for retrofit2
We'll also need to update how the network functions are invoked in `*Service` classes.

hamorillo:
Closes #144 

### Description

This PR tries to improve our approach with Kotlin Coroutines. As we discussed in #143 ([here](https://github.com/Automattic/Gravatar-SDK-Android/pull/143#discussion_r1588417398)), we should use retrofit with its coroutines compatible API. 

We've modified the OpenApi generator config to use this when generating the `Service` classes. Now, they will be `suspend` functions returning `Response<*>`. That way, the caller can handle the `CoroutineScope`.

### Testing Steps

- Run the demo-app
- Try to retrieve different profiles
- Try to upload a new avatar _(You'll need your Access Token)_

### DIFF
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
@@ -73,6 +73,7 @@ import com.gravatar.ui.components.ProfileSummary
 import com.gravatar.ui.components.UserProfileState
 import com.gravatar.ui.gravatarTheme
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 
 @Composable
@@ -248,7 +249,7 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
                 Button(
                     onClick = {
                         keyboardController?.hide()
-                        scope.launch {
+                        scope.launch(Dispatchers.IO) {
                             error = ""
                             profileState = UserProfileState.Loading
                             when (val result = profileService.fetch(Email(email))) {
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
@@ -73,7 +73,6 @@ import com.gravatar.ui.components.ProfileSummary
 import com.gravatar.ui.components.UserProfileState
 import com.gravatar.ui.gravatarTheme
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 
 @Composable
@@ -249,7 +248,7 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
                 Button(
                     onClick = {
                         keyboardController?.hide()
-                        scope.launch(Dispatchers.IO) {
+                        scope.launch {
                             error = ""
                             profileState = UserProfileState.Loading
                             when (val result = profileService.fetch(Email(email))) {
--- a/gravatar-ui/src/main/java/com/gravatar/ui/GravatarImagePickerWrapper.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/GravatarImagePickerWrapper.kt
@@ -13,6 +13,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.neverEqualPolicy
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
@@ -20,10 +21,14 @@ import androidx.core.net.toFile
 import com.gravatar.services.AvatarService
 import com.gravatar.services.ErrorType
 import com.gravatar.services.GravatarListener
+import com.gravatar.services.Result
 import com.gravatar.types.Email
 import com.yalantis.ucrop.UCrop
 import com.yalantis.ucrop.UCropActivity
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import java.io.File
+import com.gravatar.di.container.GravatarSdkContainer.Companion.instance as GravatarSdkDI
 
 /**
  * UI component that wraps the received [@Composable], allowing the user to pick an image from the
@@ -49,11 +54,20 @@ public fun GravatarImagePickerWrapper(
     val imagePickerLauncher = rememberLauncherForActivityResult(ActivityResultContracts.GetContent()) {
         selectedImageUri = it
     }
+    val coroutineScope = rememberCoroutineScope()
     val uCropLauncher = rememberLauncherForActivityResult(ActivityResultContracts.StartActivityForResult()) {
         it.data?.let { intentData ->
             UCrop.getOutput(intentData)?.let { croppedImageUri ->
                 listener.onAvatarUploadStarted()
-                AvatarService().upload(croppedImageUri.toFile(), Email(email), wordpressBearerToken, listener)
+                coroutineScope.launch(GravatarSdkDI.dispatcherIO) {
+                    val response = AvatarService().upload(croppedImageUri.toFile(), Email(email), wordpressBearerToken)
+                    withContext(GravatarSdkDI.dispatcherMain) {
+                        when (response) {
+                            is Result.Success -> listener.onSuccess(Unit)
+                            is Result.Failure -> listener.onError(response.error)
+                        }
+                    }
+                }
             }
         }
     }
--- a/gravatar-ui/src/main/java/com/gravatar/ui/GravatarImagePickerWrapper.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/GravatarImagePickerWrapper.kt
@@ -26,9 +26,7 @@ import com.gravatar.types.Email
 import com.yalantis.ucrop.UCrop
 import com.yalantis.ucrop.UCropActivity
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
 import java.io.File
-import com.gravatar.di.container.GravatarSdkContainer.Companion.instance as GravatarSdkDI
 
 /**
  * UI component that wraps the received [@Composable], allowing the user to pick an image from the
@@ -59,13 +57,16 @@ public fun GravatarImagePickerWrapper(
         it.data?.let { intentData ->
             UCrop.getOutput(intentData)?.let { croppedImageUri ->
                 listener.onAvatarUploadStarted()
-                coroutineScope.launch(GravatarSdkDI.dispatcherIO) {
-                    val response = AvatarService().upload(croppedImageUri.toFile(), Email(email), wordpressBearerToken)
-                    withContext(GravatarSdkDI.dispatcherMain) {
-                        when (response) {
-                            is Result.Success -> listener.onSuccess(Unit)
-                            is Result.Failure -> listener.onError(response.error)
-                        }
+                coroutineScope.launch {
+                    when (
+                        val response = AvatarService().upload(
+                            croppedImageUri.toFile(),
+                            Email(email),
+                            wordpressBearerToken,
+                        )
+                    ) {
+                        is Result.Success -> listener.onSuccess(Unit)
+                        is Result.Failure -> listener.onError(response.error)
                     }
                 }
             }
--- a/gravatar/src/main/java/com/gravatar/GravatarApiService.kt
+++ b/gravatar/src/main/java/com/gravatar/GravatarApiService.kt
@@ -3,7 +3,7 @@ package com.gravatar
 import com.gravatar.api.apis.ProfilesApi
 import okhttp3.MultipartBody
 import okhttp3.ResponseBody
-import retrofit2.Call
+import retrofit2.Response
 import retrofit2.http.Header
 import retrofit2.http.Multipart
 import retrofit2.http.POST
@@ -12,9 +12,9 @@ import retrofit2.http.Part
 internal interface GravatarApiService : ProfilesApi {
     @Multipart
     @POST("upload-image")
-    fun uploadImage(
+    suspend fun uploadImage(
         @Header("Authorization") authorization: String,
         @Part identity: MultipartBody.Part,
         @Part data: MultipartBody.Part,
-    ): Call<ResponseBody>
+    ): Response<ResponseBody>
 }
--- a/gravatar/src/main/java/com/gravatar/api/apis/ProfilesApi.kt
+++ b/gravatar/src/main/java/com/gravatar/api/apis/ProfilesApi.kt
@@ -8,7 +8,7 @@
 package com.gravatar.api.apis
 
 import com.gravatar.api.models.Profile
-import retrofit2.Call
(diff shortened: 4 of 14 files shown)


hamorillo:
@maxme I haven't finished this PR yet. I need to do a second pass about the code, but it would be nice to have your opinion before your AFK. (If possible, I know there's only one day left, so if you can't, no worries at all.)

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
```diff
@@ -10,9 +10,15 @@ import okhttp3.OkHttpClient
 import retrofit2.Retrofit
 import retrofit2.converter.gson.GsonConverterFactory
 
-internal class GravatarSdkContainer private constructor() {
-    companion object {
-        val instance: GravatarSdkContainer by lazy {
+/**
+ * Container that handle the DI for the Gravatar SDK.
+ */
+public class GravatarSdkContainer private constructor() {
```
hamorillo: My main concert is about making this class public so that it can be used in the `gravatar-ui ` module. I wanted to use the dispatcher in the [GravatarImagePickerWrapper.kt](https://github.com/Automattic/Gravatar-SDK-Android/pull/151/files#diff-30259c0c1479d08a6f6951482b4f25dee6b030d4e33db9ea7ccc01ffefbbe529), but maybe I should look for another option. Wdty?

Note that not everything will be public, but still not sure if that's the best way.

gravatar/src/main/java/com/gravatar/api/apis/ProfilesApi.kt
```diff
@@ -25,11 +25,11 @@ internal interface ProfilesApi {
      *
      * @param profileIdentifier This can either be an email address, SHA256 hash of an email address, or profile URL slug.
      * @param authorization Bearer token to authenticate the request. Full profile information is only available in authenticated requests. (optional)
-     * @return [Call]<[Profile]>
+     * @return [Profile]
      */
     @GET("profiles/{profileIdentifier}")
-    fun getProfileById(
+    suspend fun getProfileById(
```
maxme, line 31: Reading the auto generated documentation made me realize we don't have an error defined in`ErrorType` for 429 (Rate limit exceeded). We should add it asap (either in this PR or create a new ticket).

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
maxme: What are our options?
- what you've done: make this class and fields public.
- have another internal container for `gravatar-ui`. We'll need to duplicate a few things (dispatchers for now, maybe more later). This would make sense if someone could use `gravatar-ui` without `gravatar`, but I don't think we'll ever do this. 

I think it's fine making these fields public, I doubt we'll need to override them in other context than prod/testing environments so even if a third party use them by mistake, it shouldn't have a big impact.

hamorillo:
The changes are ready for a first review. 

@thomashorta @eyedol, as you participated in the original discussion, feel free to express your opinions.

maxme (approved):
LGTM :shipit: 

👍 Thanks for adding new tests to the ProfileService.

Error management is the same as before (http code or exceptions are converted into an ErrorType and forwarded to the listener via the `Result` class). If you don't want to deal with the 429 error in this PR, please open a new ticket.

gravatar/src/main/java/com/gravatar/api/apis/ProfilesApi.kt
hamorillo: [Ticket created](https://github.com/Automattic/Gravatar-SDK-Android/issues/153)

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -248,7 +249,7 @@ private fun ProfileTab(modifier: Modifier = Modifier, onError: (String?, Throwab
                 Button(
                     onClick = {
                         keyboardController?.hide()
-                        scope.launch {
+                        scope.launch(Dispatchers.IO) {
```
eyedol: If the suspend function is `main-safe` the caller won't have to provide a dispatcher for it. Similar to how Retrofit callback calls work. In general better to design the `suspend` function to define the dispatcher it should run in

gravatar-ui/src/main/java/com/gravatar/ui/GravatarImagePickerWrapper.kt
```diff
@@ -49,11 +54,20 @@ public fun GravatarImagePickerWrapper(
     val imagePickerLauncher = rememberLauncherForActivityResult(ActivityResultContracts.GetContent()) {
         selectedImageUri = it
     }
+    val coroutineScope = rememberCoroutineScope()
     val uCropLauncher = rememberLauncherForActivityResult(ActivityResultContracts.StartActivityForResult()) {
         it.data?.let { intentData ->
             UCrop.getOutput(intentData)?.let { croppedImageUri ->
                 listener.onAvatarUploadStarted()
-                AvatarService().upload(croppedImageUri.toFile(), Email(email), wordpressBearerToken, listener)
+                coroutineScope.launch(GravatarSdkDI.dispatcherIO) {
```
eyedol: Same comment as earlier

gravatar/src/main/java/com/gravatar/GravatarApiService.kt
```diff
@@ -12,9 +12,9 @@ import retrofit2.http.Part
 internal interface GravatarApiService : ProfilesApi {
     @Multipart
     @POST("upload-image")
-    fun uploadImage(
+    suspend fun uploadImage(
```
eyedol, line 15: 👍🏿

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
eyedol: If you make the suspend functions main safe you might not need to expose it but if you want to make the dispatchers defined in it for general use, then I guess relaxing its visibility is the way to go.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -22,51 +16,35 @@ public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
...
-     * @param gravatarUploadListener The listener to notify of the upload result
      */
-    public fun upload(
-        file: File,
-        email: Email,
-        wordpressBearerToken: String,
-        gravatarUploadListener: GravatarListener<Unit, ErrorType>,
-    ) {
+    public suspend fun upload(file: File, email: Email, wordpressBearerToken: String): Result<Unit, ErrorType> {
         val service = GravatarSdkDI.getGravatarApiV1Service(okHttpClient)
```
eyedol: I know this is not part of your changes but I think it's best if this is defined as a class property instead of a local val as it seems for every time the `upload` function is called, it  setups the underlying retrofit service. I guess we could avoid doing that as I don't see anything changing upon call of the `upload` function that would require recreating the retrofit service. But yeah I stand corrected if I'm missing something and hence it was designed this way.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -22,51 +16,35 @@ public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
...
      * @param wordpressBearerToken The bearer token for the user's WordPress/Gravatar account
-     * @param gravatarUploadListener The listener to notify of the upload result
      */
-    public fun upload(
-        file: File,
-        email: Email,
-        wordpressBearerToken: String,
-        gravatarUploadListener: GravatarListener<Unit, ErrorType>,
-    ) {
+    public suspend fun upload(file: File, email: Email, wordpressBearerToken: String): Result<Unit, ErrorType> {
```
eyedol, line 29: I would define this main safe as recommended in the [coroutines best practices doc](https://developer.android.com/kotlin/coroutines/coroutines-best-practices#main-safe)

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -22,51 +16,35 @@ public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
...
+                // Log the response body for debugging purposes if the response is not successful
+                Logger.w(
+                    LOG_TAG,
+                    "Network call unsuccessful trying to upload Gravatar: $response.body",
+                )
+                Result.Failure(errorTypeFromHttpCode(response.code()))
+            }
+        } catch (ex: Exception) {
+            Result.Failure(ex.error())
+        }
```
eyedol: It seems you are catching all exceptions which isn't safe as you're also catching `CancellationException`. Conventional advice is you re-throw cancellation exceptions.

```kotlin
...
return try {
    val response = service.uploadImage("Bearer $wordpressBearerToken", identity, filePart)

    if (response.isSuccessful) {
        Result.Success(Unit)
    } else {
        // Log the response body for debugging purposes if the response is not successful
        Logger.w(
            LOG_TAG,
            "Network call unsuccessful trying to upload Gravatar: $response.body",
        )
        Result.Failure(errorTypeFromHttpCode(response.code()))
    }
} catch( ce: CancelationException) {
    throw ce
} catch (ex: Throwable) {
    Result.Failure(ex.error())
}

```

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -22,95 +15,34 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
...
+            } else {
+                // Log the response body for debugging purposes if the response is not successful
+                Logger.w(
+                    LOG_TAG,
+                    "Network call unsuccessful trying to get Gravatar profile: $response.body",
+                )
+                Result.Failure(errorTypeFromHttpCode(response.code()))
+            }
+        } catch (ex: Exception) {
+            Result.Failure(ex.error())
```
eyedol: Same as comment above, re-throw the `CancelationException`

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -22,95 +15,34 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
...
-        fetchWithListener(username, getProfileListener = getProfileListener)
-    }
-
     /**
      * Fetches a Gravatar profile for the given hash or username.
      *
      * @param hashOrUsername The hash or username to fetch the profile for
      * @return The fetched profile
      */
     public suspend fun fetch(hashOrUsername: String): Result<Profile, ErrorType> {
```
eyedol: Same as above, make this main-safe if possible

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -22,95 +15,34 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
...
-                    override fun onSuccess(response: Profile) {
-                        it.resume(Result.Success(response))
-                    }
-
-                    override fun onError(errorType: ErrorType) {
-                        it.resume(Result.Failure(errorType))
-                    }
-                },
-            )
+        val service = GravatarSdkDI.getGravatarApiV3Service(okHttpClient)
```
eyedol: Also this `service` local variable can be defined as a property to avoid recreating the retrofit service like commented above

eyedol (commented):
Nice seeing these changes coming to live. In general it looks good. Passed a few comments worth looking into. Thanks for picking this up so fast.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
hamorillo: Nice suggestion! 🚀

hamorillo:
Thanks for your comments @eyedol! I've improved the code with some of your comments.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -2,71 +2,51 @@ package com.gravatar.services
...
-                            // Log the response body for debugging purposes if the response is not successful
-                            Logger.w(
-                                LOG_TAG,
-                                "Network call unsuccessful trying to upload Gravatar: $response.body",
-                            )
-                            gravatarUploadListener.onError(errorTypeFromHttpCode(response.code()))
-                        }
-                    }
-                }
+        return runCatchingService {
```
eyedol, line 34: 👍🏿

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -2,71 +2,51 @@ package com.gravatar.services
...
+            withContext(Dispatchers.IO) {
```
eyedol: For consistency perhaps use this one `GravatarSdkDI.dispatcherIO`? As it seems to be the one managed by us.

gravatar/src/main/java/com/gravatar/services/ServiceUtils.kt
```diff
@@ -0,0 +1,14 @@
+package com.gravatar.services
+
+import kotlinx.coroutines.CancellationException
+
+internal inline fun <T> runCatchingService(block: () -> Result<T, ErrorType>): Result<T, ErrorType> {
```
eyedol, line 5: 👍🏿

gravatar/src/test/java/com/gravatar/services/ProfileServiceTests.kt
```diff
@@ -0,0 +1,121 @@
...
+        coEvery { containerRule.gravatarApiServiceMock.getProfileById(username, any()) } returns mockResponse
+
+        val loadProfileResponse = profileService.fetchByUsername(username)
+
+        coVerify(exactly = 1) { containerRule.gravatarApiServiceMock.getProfileById(username) }
+        assertTrue(loadProfileResponse is Result.Success)
+    }
+
+    @Test
+    fun `given an username when loading its profile but date is NOT returned then result is UNKNOWN failure`() =
```
eyedol: Small typo ~date~ `data`

gravatar/src/test/java/com/gravatar/services/ProfileServiceTests.kt
```diff
@@ -0,0 +1,121 @@
...
+            val mockResponse = mockk<Response<Profile>> {
+                every { isSuccessful } returns true
+                every { body() } returns null
+            }
+            coEvery { containerRule.gravatarApiServiceMock.getProfileById(username, any()) } returns mockResponse
+
+            val loadProfileResponse = profileService.fetchByUsername(username)
+
+            coVerify(exactly = 1) { containerRule.gravatarApiServiceMock.getProfileById(username) }
+            assertTrue((loadProfileResponse as Result.Failure).error == ErrorType.UNKNOWN)
```
eyedol, line 60: Not sure if we should test for the other error types like `TIMEOUT` similar to the test cases in `AvatarServiceTest`

eyedol (approved):
Thanks for making the changes. It looks good. Passed a few minor comments worth checking other than that LGTM.

gravatar/src/test/java/com/gravatar/services/ProfileServiceTests.kt
hamorillo: I'll probably refactor that in another PR.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
hamorillo: Good catch! Thanks.

PR #152, opened 2024-05-17
`demo-app` - Load AccessToken from `local.properties`

hamorillo:
### Description

To simplify the development process, it would be nice to load the accessToken from the `local.properties`, as we do with the Email.

As a developer you just need to add the following line in your local.properties with your accessToken:

`demo-app.accessToken=YOUR_TOKEN_HERE`

### Testing Steps

- Verify that adding your `accessToken` in `local.properties` loads it in the Avatar Update Tab.

### DIFF
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -43,8 +43,8 @@ android {
             )
             buildConfigField(
                 "String",
-                "DEMO_ACCESS_TOKEN",
-                "\"${properties["demo-app.accessToken"]?.toString() ?: ""}\"",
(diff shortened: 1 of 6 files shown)


maxme (approved):
LGTM, but I'd use this occasion to rename `accessToken` to `wordPressBearerToken` or similar just to avoid any confusion with the future apiKey or future gravatar accessToken.

hamorillo:
> LGTM, but I'd use this occasion to rename `accessToken` to `wordPressBearerToken` or similar just to avoid any confusion with the future apiKey or future gravatar accessToken.

Great suggestion! Done!

PR #155, opened 2024-05-21
`gravatar`- 429 - Rate Limit Exceeded

### LINKED ISSUE #153: Add ErrorType for 429 (Rate limit exceeded).
See https://github.com/Automattic/Gravatar-SDK-Android/pull/151#discussion_r1604639197

hamorillo:
Closes #153 

### Description

With the integration of the new REST API, we'll use the code 429 when a user reaches its limit. In this PR, we add the `429` error code as a handled one. 

### Testing Steps

- Smoke test the app, review new tests, CI.


gravatar/src/test/java/com/gravatar/services/ErrorUtilsTest.kt
mlumeau, line 1: Couldn't that test be merged with `ErrorTypeTest` (and `ErrorUtils` with `ErrorType`)?

mlumeau (approved):
LGTM, found no way to trigger the 429 response from the server, but it's not what we should test here anyway. One small note regarding classes that can be merged IMO, but not blocking for this PR. Good job with the new tests.

gravatar/src/test/java/com/gravatar/services/ErrorUtilsTest.kt
hamorillo: Yeah, it totally makes sense. I've merged both classes and the associated test files. Thanks!

mlumeau (approved):
That's perfect now!

PR #159, opened 2024-05-23
`gravatar`- Update date/time fields

### LINKED ISSUE #157: `gravatar` - Update endpoint definition
As discussed, `registration_date` will follow the `date-time` format. We need to update the local file with the definition and regenerate de model.

hamorillo:
Closes #157 

### Description

We'll need to update [api-gravatar.json](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/gravatar/openapi/api-gravatar.json) with the final version of the definition. But in order to keep moving, updating it manually based on previous discussion.

On the other hand, [Instant](https://developer.android.com/reference/java/time/Instant) class is a better representation of the [DateTime](https://www.baeldung.com/openapi-map-date-types#2-date-time) format in our SDK. So, we are modifying the mapping configuration.

We also need to add an adapter to deserialize and string with the proper format to an Instant instance.

### Testing Steps

- Smoke test the demoApp, code review, and CI should be enough. (The change will be more important with a following PR that will enable the API Key so we'll receive those values).

### DIFF
--- a/gravatar/build.gradle.kts
+++ b/gravatar/build.gradle.kts
@@ -37,6 +37,8 @@ android {
         }
     }
     compileOptions {
+        isCoreLibraryDesugaringEnabled = true
+
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
     }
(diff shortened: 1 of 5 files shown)


gravatar/build.gradle.kts
```diff
@@ -72,6 +74,8 @@ android {
 }
 
 dependencies {
+    coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:2.0.4")
```
hamorillo, line 77: Some `java.time` classes are not available by default on API < 26. For that reason, we need to add [desugaring](https://developer.android.com/studio/write/java8-support) on the core module.

mlumeau (approved):
Code changes LGTM, should work as intended, I noted no difference while testing.

PR #161, opened 2024-05-23
`gravatar`- Add API Key initialization to the SDK

### LINKED ISSUE #156: `gravatar`- Setting the Gravatar API Key in the SDK
We must provide some method/s to set the API Key in the SDK.

hamorillo:
Closes #156 

### Description

Adding the method to initialize the SDK with the API Key. At the moment, that API Key is required by the new REST API when you want to get a full profile. Without the API  Key, some fields are not returned by the backend because you need to be authenticated.

Also adding in the DemoApp an expandable section that allows to review the profile raw data:

image  image

### Testing Steps

- Add your ApiKey in `local.properties` as follows:
```
demo-app.gravatar.api.key=YOUR_API_KEY_HERE
```
- Launch the DemoApp
- Go to the `Profile` tab
- Get any profile
- Expand the `Raw Profile`
- Verify it contains fields only present within authenticated requests like: `registrationDate`, `lastProfileEdit`, `contactInfo` etc.

- You can also verify that you don't receive those fields when the ApiKey is removed from `local.properties`.

### DIFF
--- a/gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
+++ b/gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
@@ -26,10 +26,6 @@ internal class GravatarSdkContainer private constructor() {
 
     private fun getRetrofitApiV3Builder() = Retrofit.Builder().baseUrl(GRAVATAR_API_BASE_URL_V3)
 
-    val dispatcherMain: CoroutineDispatcher = Dispatchers.Main
-    val dispatcherDefault = Dispatchers.Default
-    val dispatcherIO = Dispatchers.IO
-
     private val gson = GsonBuilder().setLenient()
         .registerTypeAdapter(
             Instant::class.java,
@@ -38,6 +34,12 @@ internal class GravatarSdkContainer private constructor() {
             },
         )
         .create()
+
+    val dispatcherMain: CoroutineDispatcher = Dispatchers.Main
+    val dispatcherDefault = Dispatchers.Default
+    val dispatcherIO = Dispatchers.IO
+
+    var apiKey: String? = null
 
     /**
      * Get Gravatar API service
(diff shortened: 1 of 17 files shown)


hamorillo:
I'll keep the PR as a draft because we need some backend changes. But feel free if you want to review the code or run the app without the ApiKey (with the API Key will crash until the changes are done). @mlumeau

hamorillo:
Backend changes have been deployed, so the PR is ready to review.

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
```diff
@@ -53,7 +56,7 @@ internal class GravatarSdkContainer private constructor() {
 
     fun getGravatarApiV3Service(okHttpClient: OkHttpClient? = null): GravatarApiService {
         return getRetrofitApiV3Builder().apply {
-            okHttpClient?.let { client(it) }
+            client(okHttpClient ?: OkHttpClient().newBuilder().addInterceptor(AuthenticationInterceptor()).build())
```
mlumeau: We're not adding our interceptor to any dev-supplied HTTP client? Not necessarily saying we should, but since we made the AuthenticationInterceptor `internal`, devs will have to re-write it in order to support an API key...

Here's what this change would look like:

```suggestion
            client((okHttpClient ?: OkHttpClient()).newBuilder().addInterceptor(AuthenticationInterceptor()).build())
```

gravatar/src/main/java/com/gravatar/services/AuthenticationInterceptor.kt
```diff
@@ -0,0 +1,20 @@
+package com.gravatar.services
+
+import okhttp3.Interceptor
+import okhttp3.Request
+import okhttp3.Response
+import com.gravatar.di.container.GravatarSdkContainer.Companion.instance as GravatarSdkDI
+
+internal class AuthenticationInterceptor : Interceptor {
```
mlumeau, line 8: I pushed a quick modification in order to improve readability and facilitate adding other headers if needed in the future. Tell me what you think.

mlumeau (commented):
I just have a question regarding third party HTTP clients. Once we agree I'll approve the PR.

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
hamorillo: It's a good point. Tbh, I had the same doubt. I've committed the suggested change.

mlumeau (approved):
LGTM 🚀

PR #168, opened 2024-06-03
Refactor `UserProfileState` to a generic state

### LINKED ISSUE #167: `gravatar` -  Generic state for Gravatar Composables
https://github.com/Automattic/Gravatar-SDK-Android/pull/123#issuecomment-2074978761

hamorillo:
Closes #167 

### Description

Replacing the concrete `UserProfileState` with a generic state that can be used with different types of data. For example, image a component that receives a `String`; with that generic state, we can implement it. 

### Testing Steps

- Smoke test the app. Everything should work as before. Pay special attention to the `Profile` tab.

### DIFF
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
@@ -56,6 +56,7 @@ import com.gravatar.AvatarQueryOptions
 import com.gravatar.AvatarUrl
 import com.gravatar.DefaultAvatarOption
 import com.gravatar.ImageRating
+import com.gravatar.api.models.Profile
 import com.gravatar.demoapp.BuildConfig
 import com.gravatar.demoapp.R
 import com.gravatar.demoapp.theme.GravatarDemoAppTheme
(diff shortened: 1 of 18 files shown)


gravatar-ui/src/main/java/com/gravatar/ui/components/State.kt
```diff
@@ -14,38 +14,39 @@ import com.gravatar.api.models.Profile
...
 import kotlinx.coroutines.delay
 import java.net.URI
 
 /**
- * [UserProfileState] represents the state of a user profile loading.
+ * [State] represents the state of a user profile loading.
  * It can be in a [Loading] state or a [Loaded] state.
  */
-public sealed class UserProfileState {
+public sealed class State<out T> {
```
hamorillo: I'm doubting about `State` or `GravatarState`. Wdyt?

mlumeau: I'm worried `State` will conflict with user-defined classes. Let's be honest, we all have these generically-named classes in our projects. On the other hand `GravatarState` is misleading because in my mind it would refer to the state of the whole SDK. `GravatarComponentState` maybe?

hamorillo: > I'm worried State will conflict with user-defined classes.

With the auto-complete of the modern IDE, I wouldn't say it's a big deal. But yeah, I see the point.

> `GravatarComponentState` maybe?

I'm not 100% done with it, but given that, I don't have a better option 😅 let's change it.

mlumeau: Or just `ComponentState` for short? 🤔

hamorillo: Done!

mlumeau (approved):
LGTM!

PR #171, opened 2024-06-04
Release 1.0.0

hamorillo:
### Description

The new REST API is integrated, and the UI components are stable. We are in good shape to release `1.0.0`.

We can improve our documentation, but, in my opinion, that shouldn't be a blocker.

**Note:** When this PR is approved and merged, I'll create the `1.0.0` tag and the release.

### Testing Steps

- [x] Think about whether we are missing something for the release
- [x] Review PR changed files
image


mlumeau (approved):
We should be good to go. 🚀

PR #173, opened 2024-06-05
Bump kotlin to v2.0.0

eyedol:
Closes #147 

### Description

- Bumped Kotlin to v2.0.0
- Applied compose compiler's Gradle plugin as it's now required for use with kotlin 2.0.0
- Enable skipping mode for compose compiler 

### Testing Steps

- Built and run the demo app to make sure it runs
- Ran `./gradlew ktlintcheck detekt` to make sure it was successful
- Ran `./gradlew assemble` to make sure it was successful
- Ran `./gradlew test` to make sure it was successful
- Ran `./gradlew :gravatar-ui:verifyRoborazziDebug -Pscreenshot`

PR #175, opened 2024-06-11
Bump AGP version + enabling desugaring in other modules

hamorillo:
### Description

After bumping Kotlin to 2.0.0 #173, some CI tasks [have started failing](https://github.com/Automattic/Gravatar-SDK-Android/actions/runs/9461375953/job/26062126862). You can replicate it locally by executing:

```
./gradlew :gravatar:dependencies :gravatar-ui:dependencies
```

Bumping the AGP solves the issue.

**Note:** The latest version of AGP requires your modules to enable desugaring if they depend on another module using desugaring ([see](
https://issuetracker.google.com/issues/329346764)). In our case, `gravatar-ui` and `demo-app` depend on `core,` which has desugaring enabled. It shouldn't be a problem, but feel free to open a discussion if you think of another way.

### Testing Steps

- Checkout locally the code and execute:

```
./gradlew :gravatar:dependencies :gravatar-ui:dependencies
```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,7 +5,7 @@ import org.jetbrains.dokka.versioning.VersioningPlugin
 
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 plugins {
-    id("com.android.application") version "8.1.0" apply false
+    id("com.android.application") version "8.4.1" apply false
     id("org.jetbrains.kotlin.android") version "2.0.0" apply false
     id("org.jetbrains.kotlin.plugin.compose") version "2.0.0" apply false
 
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -62,6 +62,8 @@ android {
         }
     }
     compileOptions {
+        isCoreLibraryDesugaringEnabled = true
+
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
     }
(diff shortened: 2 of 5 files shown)


mlumeau:
I'm not sure I understand the fix. Will that force our users to enable this feature as well?

hamorillo:
> Will that force our users to enable this feature as well?

Yes. Apps using our library will need to enable desugaring support.

> I'm not sure I understand the fix.

We've bumped the Kotlin version to 2.0.0. That broke this [CI job](https://github.com/Automattic/Gravatar-SDK-Android/actions/runs/9461375953/job/26062126862), and bumping algo AGP solves the issue. As a side effect of the AGP bump, modules depending on others with desugaring support enabled [also need to enable it](https://issuetracker.google.com/issues/329346764). 

I found a comment there very helpful:

> Re #4: This is indeed working as intended. In AGP 8.3.0 there are more validations that protect the application from runtime crashes. One such validation is to ensure that if any library/subproject that the module depends uses Java 8+ API desugaring, the consumer must have that feature enabled. Otherwise, you may hit runtime crashes.
>
> Re #5: https://developer.android.com/studio/write/java8-support#library-desugaring has more details on this. Note that you must enable core library desugaring on every module, but you'll get only 1 copy of the desugar_jdk_lib in your final APK. In the release variant, AGP will ensure that any unused APIs from desugar_jdk_lib are removed, so your APK will increase only because of APIs that are actually used. Hope that helps.


If we don't want apps to enable it, as I can see, we have two options:

- Revert the Kotlin bump and keep everything as it was (I don't think it makes sense because at some point we'll need to increase versions)
- Remove the need for desugaring in our library. We are using it for `java.time` classes, and it's pretty convenient but...

In my opinion, it should be fine to describe it in the documentation and keep it. My second option will be to try to remove the desugaring need. Wdyt?

mlumeau (approved):
You're right, let's embrace Kotlin 2.0. I'm fine with keeping desugaring as long as we update our documentation to make it clear this is a requirement. 👍

hamorillo:
Added to the docs.

mlumeau (approved):

PR #178, opened 2024-06-18
`gravatar-ui` - Profile components over a `surface`

hamorillo:
### Description

In order to use the proper theme values, we should show the profile UI component over a surface element. That way, client apps can override the themes as needed.

When overriding `GravatarTheme` to customize the styles, we could obtain some unexpected results due to components not being on a surface.

The client app could add the Gravatar `@Composable` over a surface, but I would expect the SDK to work out of the box just by coding something like:

```kotlin
CompositionLocalProvider(
    LocalGravatarTheme provides object : GravatarTheme {
        // Override theme colors
            @Composable
            get() = colorScheme
    },
) {
    LargeProfileSummary(state)
}
```

### Testing Steps

- Code review and screenshot tests should be enough

### DIFF
--- a/gravatar-ui/src/main/java/com/gravatar/ui/components/LargeProfile.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/components/LargeProfile.kt
@@ -28,7 +28,7 @@ import java.net.URI
 
 /**
  * [LargeProfile] is a composable that displays a user's profile card.
- * Given a [UserProfile], it displays a [LargeProfile] using the other atomic components provided within the SDK.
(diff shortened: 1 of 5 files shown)


mlumeau (approved):
LGTM, works as expected.

AdamGrzybkowski (approved):
Looks good to me 👍  
I can see that we already have an example of how to override the theme so all good here as well. 

I thought about whether this change negatively affects the behavior and my understanding is that it does not. The default behavior stays the same and will only be modified when explicitly overriding the theme.

hamorillo:
Thanks to both for your reviews!

PR #181, opened 2024-06-19
`gravatar-ui`- Allow custom avatar and view profile button

### LINKED ISSUE #180: `gravatar-ui` - Allow custom avatar and view profile button
Some apps decorate their avatars to, for example, give information about the account status. We have a clear example with the [PocketCast app](https://github.com/Automattic/pocket-casts-android).

image

Allowing custom avatars will make our UI components more flexible.

hamorillo:
Closes #180 

### Description

Apps usually add some decoration to their avatars, so allowing pass as a parameter for a custom avatar makes our profile UI components more flexible.

A similar thing happens with the view profile button.

We can take PocketCast as a good example. They decorate the avatar to indicate whether a user has an active subscription.

Alt Text

Profile UI components will continue to be exactly the same as before by default, but now, developers using the SDK will have the possibility to pass their composable for the avatar or the view profile button. 

**Note:** This PR doesn't break any previous implementation because, by default, it is the same behavior and methods declaration.

### Testing Steps

- Code review and CI should be enough.


hamorillo:
This PR only allows the customization of the `avatar` and `view profile` buttons. However, should we iterate over this and make all the components customizable? 

It would be nice to have, and it should be quick for us to implement. We only need to follow the same approach as we've done.

What do you think @mlumeau @AdamGrzybkowski ?

AdamGrzybkowski:
>This PR only allows the customization of the avatar and view profile buttons. However, should we iterate over this and make all the components customizable?

I think that's something to discuss with the design. There's probably a line where we will lose the Gravatar feel when we allow too much customization.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
```diff
@@ -52,9 +54,10 @@ public fun ViewProfileButton(
     modifier: Modifier = Modifier,
     textStyle: TextStyle = MaterialTheme.typography.titleSmall.copy(color = MaterialTheme.colorScheme.onBackground),
     inlineContent: @Composable ((String) -> Unit)? = { DefaultInlineContent(textStyle.color) },
+    buttonText: String = stringResource(R.string.view_profile_button),
```
AdamGrzybkowski: I believe we should keep the `@Composable inlineContent` param as the last one to keep the trailing lambda funtionality. 
Reordering will not be backward compatible so I guess we need to add a new `ViewProfileButton` with new params, thoughts?

AdamGrzybkowski (commented):
Looks good 👍 
One comment left that I think needs to be addressed.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/ViewProfileButton.kt
hamorillo: Nice suggestion. Yeah, I agree it's better to keep the `inlineContent` as the last parameter. 

I've pushed some changes. Let me know what you think.

AdamGrzybkowski (approved):
LGTM 👍 Thanks for the fix ✋

PR #184, opened 2024-06-20
Update REST API with interest

### LINKED ISSUE #182: `gravatar`- Update REST API with `Interest`

hamorillo:
Closes #182 

### Description

Gravatar has recently added interest to its REST API, so this PR updates the models to retrieve the new field.

image

### Testing Steps

**Note:** The `interest` field is only provided in authenticated requests.

1. Run the demoApp with a valid API key
2. Retrieve a profile with interests and verify you can see them in the `Raw Profile`
3. Retrieve a profile without interest and verify you can see an empty array
4. Run the demoApp without an API key
5. Retrieve a profile with interests and verify you can't see them. You'll see the field as `null`.
6. Retrieve a profile without interests and verify you can't see them. You'll see the field as `null`.


gravatar/src/main/java/com/gravatar/api/models/Profile.kt
```diff
@@ -73,6 +74,9 @@ public data class Profile(
     // A list of links the user has added to their profile. This is only provided in authenticated API requests.
     @SerializedName("links")
     val links: kotlin.collections.List<Link>? = null,
+    // A list of interests the user has added to their profile. This is only provided in authenticated API requests.
+    @SerializedName("interests")
+    val interests: kotlin.collections.List<Interest>? = null,
```
AdamGrzybkowski, line 79: ⚠️ I think we have to be careful with such modifications. This would break the project compilation if the `Profile` constructor was used without the named params (would be broken regardless if used in Java). 
Because these are data classes there would also be some behavioral changes to the destructing as per [this](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api).

hamorillo: > Because these are data classes there would also be some behavioral changes to the destructing as per [this](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api).

Great catch and awesome stuff in that link! 

The result of using the [Binary compatibility validator﻿](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#use-the-binary-compatibility-validator) between `trunk` and this PR: 

image

I bet this is not the last time we'll add some fields to the REST API, so we need to come up with a process/rules about how to handle this. Some options:

- This is nice to have in order to keep the SDK up-to-date with the current REST API, but there is no hurry at all so we can **wait for a major release**. However, when will we release a new major? 🤷🏽‍♂️
- This can cause some inconveniences to the SDK users, but I would say those will be minor. We can document that with the release as best as we can and merge it. (This is far from ideal but can be a practical solution at the moment)
- We can create another constructor for Profile that keeps the same previous signature. However, we'll still have a problem if someone is using the `copy` method. (We can try to get rid of `data class`, maybe in a major release?)
image
This approach doesn't scale well if we add several fields (and we'll need to figure it out what to do with the OpenApi generator...) but it could work...

What do you think?

cc: @mlumeau

AdamGrzybkowski: >This is nice to have in order to keep the SDK up-to-date with the current REST API, but there is no hurry at all so we can wait for a major release. However, when will we release a new major? 🤷🏽‍♂️

I think that is the least ideal solution. We can't bump to major every time we add some new fields from the API. REST API is backward compatible, the new fields are extra additions but the old SDK will work without them, so should our new version of the SDK. Having new fields should be an optional extra, not a requirement. 

>This can cause some inconveniences to the SDK users, but I would say those will be minor. We can document that with the release as best as we can and merge it. (This is far from ideal but can be a practical solution at the moment)

Although not ideal, if we are able to verify the number of library downloads we could consider that. But if we go that route I think we should also rethink our approach to those API-generated classes and all other `data classes` as they will be a pain for us in the long term.

hamorillo: > I think that is the least ideal solution. We can't bump to major every time we add some new fields from the API. REST API is backward compatible, the new fields are extra additions but the old SDK will work without them, so should our new version of the SDK. Having new fields should be an optional extra, not a requirement.

That's what I was trying to say. We can't bump the major just for this, but waiting for a major release will keep the SDK outdated for an undefined period of time.

> But if we go that route I think we should also rethink our approach to those API-generated classes and all other data classes as they will be a pain for us in the long term.

I don't know if I'm following you. I see why we need to get rid of `data classes` but do you mean something more with that?

AdamGrzybkowski: >I don't know if I'm following you. I see why we need to get rid of data classes but do you mean something more with that?

We have data classes generated with OpenAPI generator but also some that are fully created by us - modifying those will have the same issues.

eyedol: Just chipping in with a post that I saw that details this issue and some possible workarounds for it: https://jakewharton.com/public-api-challenges-in-kotlin/ -- There's also a library that is aiming to solve it: https://github.com/tobrun/kotlin-data-compat. In case you are not aware of them and they may help...

hamorillo: Great info! Thanks @eyedol

hamorillo:
I'm closing this PR. I'll open a new one when #204, #207, and #206 are merged.

PR #186, opened 2024-06-20
Bump retrofit and and gson converter to `2.11.0`

wzieba:
### Description

To address https://github.com/Automattic/Gravatar-SDK-Android/security/dependabot/23

### Testing Steps

- Browse the changelog to see if some changes affected this project
  - I see one candidate: 
    > Eagerly reject suspend fun functions that return Call<Body>. These are never correct, and should declare a return type of Body directly.
  
  but it looks that this SDK is not using `Call<*>` anymore
- Smoke test the SDK


hamorillo (approved):
LGTM! Thanks! 🚀

PR #187, opened 2024-06-21
Integrate Binary Compatibility Validator

### LINKED ISSUE #179: Integrate Binary Compatibility Validator
There's a tool that validates exposed public APIs - https://github.com/Kotlin/binary-compatibility-validator.
Adding it to the Gravatar library will help us make sure the public APIs are not changed accidentally.

AdamGrzybkowski:
Closes #179 

### Description

This PR adds the Binary Compatibility Validator plugin that will help us ensure backward compatibility with each release. 

There's an extra step added to the `.pipeline.yml` file that will run `./gradlew apiCheck` command so any changes to the API will have to be reviewed and approved.

### Testing Steps
- Review the public API, maybe there are some internal APIs that we want to exclude from the check that I'm not aware of. 
- CI step test to make sure it's being run when it should

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -40,6 +40,12 @@ steps:
     artifact_paths:
       - "**/build/test-results/*/*.xml"
 
+  - label: "Binary compatibility check"
+    if: build.pull_request.base_branch == "trunk"
+    command: |
+      echo "--- Validating binary compatibility"
+      ./gradlew apiCheck
+
   - wait: ~
     if: build.branch == "trunk" || build.tag != null
 
--- a/gravatar-ui/api/gravatar-ui.api
+++ b/gravatar-ui/api/gravatar-ui.api
@@ -1,0 +1,299 @@
+public final class com/gravatar/ui/GravatarImagePickerWrapperKt {
+	public static final fun GravatarImagePickerWrapper (Lkotlin/jvm/functions/Function2;Ljava/lang/String;Ljava/lang/String;Lcom/gravatar/ui/GravatarImagePickerWrapperListener;Landroidx/compose/ui/Modifier;Lcom/gravatar/ui/ImageEditionStyling;Landroidx/compose/runtime/Composer;II)V
+}
+
+public abstract interface class com/gravatar/ui/GravatarImagePickerWrapperListener : com/gravatar/services/GravatarListener {
+	public abstract fun onAvatarUploadStarted ()V
+}
+
+public abstract interface class com/gravatar/ui/GravatarTheme {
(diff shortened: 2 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -40,6 +40,12 @@ steps:
     artifact_paths:
       - "**/build/test-results/*/*.xml"
 
+  - label: "Binary compatibility check"
+    if: build.pull_request.base_branch == "trunk"
```
AdamGrzybkowski, line 44: Thoughts about this? I think we don't need to run the API ckeck on PRs that are not targeting the `trunk` branch. 

I'm not sure yet what process we would follow when pushing a fix to an older version - whether we would target a tag or a release branch, but we should cover that scenario as well.

hamorillo: > Thoughts about this? I think we don't need to run the API ckeck on PRs that are not targeting the trunk branch.

I agree; I don't think we need to cover other branches than `trunk` right now.

hamorillo (approved):
:shipit:

gravatar-ui/api/gravatar-ui.api
```diff
@@ -0,0 +1,299 @@
...
+
+public final class com/gravatar/ui/components/ComponentStateKt {
+	public static final fun LoadingToLoadedStatePreview (Lkotlin/jvm/functions/Function3;Landroidx/compose/runtime/Composer;II)V
+}
+
+public final class com/gravatar/ui/components/ComposableSingletons$ComponentStateKt {
+	public static final field INSTANCE Lcom/gravatar/ui/components/ComposableSingletons$ComponentStateKt;
+	public static field lambda-1 Lkotlin/jvm/functions/Function3;
+	public fun <init> ()V
+	public final fun getLambda-1$gravatar_ui_release ()Lkotlin/jvm/functions/Function3;
```
wzieba, line 85: Do you know what's `gravatar_ui_release`? This look suspicious for a public contract.

wzieba (approved):
Configuration of the plugin looks good! I have one question/concern about the public contract itself, but I guess it can/should be addressed in a different PR (if needed).

gravatar-ui/api/gravatar-ui.api
hamorillo: 🤔 There's an open issue in the binary compatibility repository: https://github.com/Kotlin/binary-compatibility-validator/issues/74

wzieba: Thanks @hamorillo , TIL!

AdamGrzybkowski: Good catch 👍 Let's keep an eye on that one

PR #188, opened 2024-06-21
Add .aiexclude config file

### LINKED ISSUE #150: Research and prevent sharing private code/API keys with Gemini
https://www.droidcon.com/2024/05/07/protect-your-code-from-gemini-in-android-studio/

AdamGrzybkowski:
### Description

Initial file to exclude some of our files/directories from Gemini as per this [documentation](https://developer.android.com/studio/preview/gemini/aiexclude). 

### Testing Steps

Think if all potential files with secrets are covered

### DIFF
--- a/.aiexclude
+++ b/.aiexclude
@@ -1,0 +1,6 @@
+# Configuration files
+local.properties
+gradle.properties
+
+# Build files
+build/
--- a/.aiexclude
+++ b/.aiexclude
@@ -1,6 +1,5 @@
 # Configuration files
-local.properties
-gradle.properties
+.properties
 
 # Build files
 build/


mlumeau (commented):
Why not `*.properties` ?

hamorillo (approved):
[WordPress file](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/.aiexclude) contains more files, but I don't see a real need to add them to our case.

AdamGrzybkowski:
>Why not *.properties ?

Good question. We only have two and I didn't want to exclude something in the future unintentionally.  Though... given what we are doing here, I think it's completely fine to exclude all `.properties` files.

PR #192, opened 2024-06-25
Remove Robolectric from the :gravatar module

### LINKED ISSUE #191: Remove Robolectric dependency from the :gravatar module
It's not needed and it adds unnecessary overhead to the tests.

AdamGrzybkowski:
Closes #191 

### Description

We do have a dependency on Robolectric in the `:gravatar` module despite the fact that we don't need it to run the tests there. This may be a leftover from before the `:-ui` module was created. 

Robolectrc adds additional overhead to the tests and makes them significantly slower so there's no reason to use it unless necessary. 

### Testing Steps

Passing tests on CI should be enough.


hamorillo (approved):
Thanks for fixing this. 🚀

PR #196, opened 2024-06-28
Fix: Remove empty spaces in Profile component when profile description is empty

### LINKED ISSUE #193: Fix: Make the Profile components responsive to the content
The height of the component should change depending on the content. For example, if there's no description the whole text area for it should be hidden leaving no empty spaces.

AdamGrzybkowski:
Closes #193

### Description

The Profile component should adjust its height depending on the content. The current implementation was leaving both the  `AboutMe` and the `Spacer` components when the description was empty leaving a lot of empty space in the middle of the component. 

Here's the comparison from Roborazzi:
Screenshot 2024-06-28 at 08 32 43

Before making changes, I've added one screenshot test for LargeProfile to ensure there won't be a regression and one new test for Profile to ensure the new changes are visible. These are committed in a separate commit. 

### Testing Steps

1. Run the demo app
2. Go to the Profile tab
3. Fetch the profile with the description 
4. Confirm no changes are visible
5. Fetch the profile without a description
6. Confirm the Profile component takes less space vertically


demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -186,7 +186,7 @@ private fun ProfileComponents(profileState: ComponentState<Profile>?, theme: The
         val defaultModifier = Modifier
             .background(gravatarTheme.colorScheme.surfaceContainer)
             .fillMaxWidth()
-            .padding(24.dp)
+            .padding(start = 24.dp, bottom = 10.dp, end = 24.dp, top = 21.dp)
```
AdamGrzybkowski, line 189: This adjusts the paddings to match the Figma. I think these should be included in our components - WDYT?

gravatar-ui/src/main/java/com/gravatar/ui/components/Profile.kt
```diff
@@ -84,8 +84,10 @@ public fun Profile(
                             UserInfo(state)
                         }
                     }
-                    Spacer(modifier = Modifier.height(16.dp))
-                    AboutMe(state)
+                    if (state.isAboutMeVisible()) {
+                        Spacer(modifier = Modifier.height(16.dp))
+                        AboutMe(state)
+                    }
```
AdamGrzybkowski, line 90: Originally, I wanted to have this check inside `AboutMe` component but having the `Spacer` made that impossible. Also, there's different spacing in the `LargeComponent` that didn't help.

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
hamorillo: > I think these should be included in our components

I'm not completely sure about adding paddings to our components. We've tried to make them as customizable as possible, so we didn't add them internally to allow SDK users to set whatever they want and keep the component as simple as possible. 🤔

hamorillo (approved):
LGTM! 🚀

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
AdamGrzybkowski: I think these could be the defaults unless others were specified. This way by default we end up with what our design shows and not no padding at all.

hamorillo: I guess you are talking about initializing the modifier with the default padding. However, that's not a recommended practice:

image

[Here](https://stackoverflow.com/questions/73415055/why-was-i-warned-optional-modifier-parameter-should-have-a-default-value-of-mod) they give a good explanation about why it is not recommended.

AdamGrzybkowski: I see! Let's leave it for now as it's not part of this PR anyway but I think this might be worth discussing with the design.

PR #197, opened 2024-06-28
Run the binary compatibility check on each code push

AdamGrzybkowski:
### Description

The CI checks are started on each code push. That means the check to run only for PRs doesn't work because that's rarely the case (only when you push updats to an already opened PR). 

I'm removing the check to always run the step.

### Testing Steps
Make sure you see the `validating binary compatibility` check in CI steps below


hamorillo (approved):
image

Awesome! :shipit:

Issue #199, opened 2024-07-01
Fix: Skeleton height is not correctly set based on the text measure funtion


AdamGrzybkowski:
There seems to be a bug somewhere [here](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/gravatar-ui/src/main/java/com/gravatar/ui/SkeletonEffect.kt#L51). The height of the skeleton is not always correctly set - it's obvious on the skeleton where the User name is. This one should be taller but it's not always the case.

PR #200, opened 2024-07-01
Fix: Loading state of the profile component is not matching the design

### LINKED ISSUE #195: Fix: Loading UI for the Profile card
| Design | How it looks |
|---|---|
| Screenshot 2024-06-26 at 15 27 29 | Screenshot 2024-06-26 at 15 27 38 |

This should bring it closer to iOS implementation as well.

AdamGrzybkowski:
Closes #195

### Description

The loading state for all the profile cards did not correctly represent what the design created. This PR aims to bring them closer to the "what we want" state. It won't be 100% correct yet, because I have found other issues (https://github.com/Automattic/Gravatar-SDK-Android/issues/199) and I didn't want to push all the changes in one PR. 

This one focuses on the Skeleton object width and padding. 

Here's the comparison using designs from Figma and screenshots from our test showing the changes:

| Design | Before - After |
|---|---|
| Screenshot 2024-07-01 at 10 50 41 | Screenshot 2024-07-01 at 10 50 17 |
| Screenshot 2024-07-01 at 10 50 38 | Screenshot 2024-07-01 at 10 50 31 |
| Screenshot 2024-07-01 at 10 50 52 | Screenshot 2024-07-01 at 10 49 52 |
| Screenshot 2024-07-01 at 10 50 57 | Screenshot 2024-07-01 at 10 48 59 |

**Why and what**

To achieve that I had to add a separate `skeletonModifier` to control the size and styling of the skeleton object ([as per this](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier#:~:text=Composables%20that%20accept%20modifiers%20to%20be%20applied%20to%20a%20specific%20subcomponent%20foo%20should%20name%20the%20parameter%20fooModifier%20and%20follow%20the%20same%20guidelines%20above%20for%20default%20values%20and%20behavior.%20Subcomponent%20modifiers%20should%20be%20grouped%20together%20and%20follow%20the%20parent%20composable%27s%20modifier.%20For%20example%3A)). It is required because of how all the components, that take in the `ComponentState`, are designed. There's no single parent composable - based on the state we create different top composables so sharing the modifier is not an option.

Alternatively, we could add that missing top composable, but this would still require us to pass in the modifier conditionally as those would differ depending on the `ComponentState`.  

### Testing Steps
1. Run the demo app
2. Go to the Profile tab
3. Tap the `Skeleton` button
4. Confirm the UI looks properly and is closer to the designs


AdamGrzybkowski:
I'm expecting the binary compatibility and screenshots tests to fail. I will upload the new "base" for both in separate commits.

AdamGrzybkowski:
Here are the public API changes - only new additions. 
Screenshot 2024-07-01 at 11 26 05

AdamGrzybkowski:
Screenshots test failing - only those in loading state. Other states are unaffected. 
Screenshot 2024-07-01 at 11 27 11

hamorillo (approved):
Looks beautiful. Thanks for improving this @AdamGrzybkowski. :shipit:

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/DisplayName.kt
```diff
@@ -48,10 +49,32 @@ public fun DisplayName(
...
+ * @param state The user's profile state
+ * @param modifier Composable modifier
+ * @param skeletonModifier Composable modifier for the loading skeleton component
+ * @param textStyle The style to apply to the text
+ */
+@Composable
+public fun DisplayName(
+    state: ComponentState<Profile>,
+    modifier: Modifier = Modifier,
+    skeletonModifier: Modifier,
```
hamorillo, line 68: ⚠️ I just noticed we miss the default value here.

```
skeletonModifier: Modifier = Modifier,
```

PR #201, opened 2024-07-02
Align left side of the Gravatar icon with other composables in the pr…

### LINKED ISSUE #194: Fix: Avatar and Gravatar Icon aren't vertically aligned
The avatar and the Gravatar icon should be vertically aligned. Currently, the Gravatar icon has a bigger left margin (including the icon padding) than the avatar.

AdamGrzybkowski:
Closes #194

### Description

**Note: I'm targeting another branch that is in a PR state because this PR is also affecting the loading state**

This PR aims to align the left side of the Gravatar icon with other composables in the profile cards (Profile and LargeProfile).
The Gravatar icon has padding so to align it with other views we have to set a negative offset to it.
Additionally, the loading skeleton view was also updated not to have the padding at the start and end of the row. Instead, there's a spacing set only between the items using `Arrangement.spacedBy(4.dp)`.

| Before | After |
|---|---|
| Screenshot 2024-07-02 at 10 32 31 | Screenshot 2024-07-01 at 16 42 13 |
| Screenshot 2024-07-02 at 10 32 25 | Screenshot 2024-07-02 at 10 27 31 |

There should be some failures in `Profile` and `LargeProfile` screenshots tests. I will record and push new screenshots. 

### Testing Steps

1. Open the Demo app
2. Go to the profile tab
3. Load the profile
4. Enable the layout inspector
5. Confirm the Gravatar icon is aligned with the avatar and other composables that are aligned to the left of the profile card.


gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
```diff
@@ -183,7 +184,7 @@ public fun SocialIconRow(socialMedia: List<SocialMedia>, modifier: Modifier = Mo
  */
 @Composable
 public fun SocialIconRow(profile: Profile, modifier: Modifier = Modifier, maxIcons: Int = 4) {
-    SocialIconRow(mediaList(profile), modifier, maxIcons)
+    SocialIconRow(mediaList(profile), modifier.offset((-5).dp), maxIcons)
```
hamorillo: 🤔 This helps with the icon alignment in our profile components but hardcodes the offset inside `SocialIconRow`. If someone wants to use this component isolated (or include us in a different design), they can't do anything to avoid the offset. Have you considered that?

Also, we left some "empty" space on the right of the component. I don't think this is a big deal for our Profile components, but as before, not sure if it looks good if we use it as an isolated component.

image

AdamGrzybkowski: >If someone wants to use this component isolated (or include us in a different design), they can't do anything to avoid the offset. Have you considered that?

I did. I've decided to include it here because the Gravatar icon is also hardcoded in that component. It's always the first element as part of of this composable (added in the `mediaList()` function). So, if the Gravatar icon is removed we can remove that offset with it. The offset can also be neutralized by setting the same positive value, so it's still possible to achieve the same effect.

Other solutions I tried:
1. Add `offset` param to `SocialMedia` class and apply it based on the icon type. This didn't work because offsetting one composable in a row didn't move others. 
2. Applying `offset` to the modifier in each profile card to the `SocialIconRow()` - the main problem here is that the same modifier is used inside this composable for the skeleton that has different paddings. So if we go this route we would have to fake the padding for the skeleton in order to match the alignment. 

I'm happy to modify this though this solution made the most sense to me. Do you have other ideas? 


>not sure if it looks good if we use it as an isolated component.

I haven't been here in the beginning so I'm not sure what's exposed on purpose and what composables are exposed as a side effect, but I thought the main goal was to provide ready-to-use, plug-and-play components - and that was my focus. I can try and modify the code to make the alternative solution number 2 work.

AdamGrzybkowski: oh and it's also possible to use the one function above
```public fun SocialIconRow(socialMedia: List<SocialMedia>, modifier: Modifier = Modifier, maxIcons: Int = 4)```
to have the raw list without both: the hardcoded Gravatar icon and the offset.

hamorillo: > I haven't been here in the beginning so I'm not sure what's exposed on purpose and what composables are exposed as a side effectt, but I thought the main goal was to provide ready-to-use, plug-and-play components

The original idea was to provide both complex, ready-to-use components like the profile ones and the atomic components that we use to render those complex composables. That way, clients can use our profile components or create and display the information they want wherever they want, with their customized design. For that reason, the `atomic` components are exposed.

>  I can try and modify the code to make the alternative solution number 2 work

In my opinion, number 2 looks better. We'll have a skeleton identical to the view, and the offset will be set outside the component, in our case from the profile composables. It sounds good to me. 

image

image

At the same time, if we use `SocialIconRow` in a different place, we don't need to neutralize the offset; it works as expected.

Wdty?

AdamGrzybkowski: Agree! Update the code. Let me know what you think now.

hamorillo (approved):
Looks great! 🚀

PR #204, opened 2024-07-02
#198/1 - Add new model classes to support REST API updates without breaking changes

### LINKED ISSUE #193: Fix: Make the Profile components responsive to the content

### LINKED ISSUE #198: `gravatar`- Improve SDK to allow REST API updates without breaking changes
See #184

hamorillo:
Closes #198 

### Description
This is the first PR of a set of three (reviewing them in order will help), whose aim is to refactor all the necessary stuff to allow us to update the REST API without breaking changes or releasing a new major version. This conversation started in #184.

The idea is to create model classes that allow us to add new fields with binary or source compatibility. This PR is highly inspired by [this post](https://jakewharton.com/public-api-challenges-in-kotlin/).

As we were already using OpenApi templates, this PR adds and refactors templates to generate model classes that can easily updated.

This PR only updates templates and generates classes that are not used yet. That will be the next PR.

### Testing Steps
- Code review and CI should be enough.


gravatar/build.gradle.kts
```diff
@@ -119,7 +119,7 @@ openApiGenerate {
             "library" to "jvm-retrofit2",
             "serializationLibrary" to "gson",
             "groupId" to "com.gravatar",
-            "packageName" to "com.gravatar.api",
+            "packageName" to "com.gravatar.restapi",
```
hamorillo, line 122: What do you think about `restapi` as the name for the new package?

gravatar/openapi/templates/data_class.mustache
```diff
@@ -144,6 +145,34 @@ import {{packageName}}.infrastructure.ITransformForStorage
 {{/isEnum}}
 {{/vars}}
 {{/hasEnums}}
-{{#vendorExtensions.x-has-data-class-body}}
+
+{{^isEnum}}
+	override fun toString() : String = "{{#allVars}}{{className}}{{name}}=${{name}}{{^-last}}, {{/-last}}{{/allVars}}"
```
AdamGrzybkowski: I think this needs to be modified. The generated value of the `toString` function doesn't have the class name. The end result is "param1=value1" instead of "ClassName(param1=value1)"

gravatar/openapi/templates/data_class.mustache
```diff
@@ -60,14 +61,14 @@ import {{packageName}}.infrastructure.ITransformForStorage
 @Deprecated(message = "This schema is deprecated.")
 {{/isDeprecated}}
 {{>additionalModelTypeAnnotations}}
-{{#nonPublicApi}}internal {{/nonPublicApi}}{{#discriminator}}interface{{/discriminator}}{{^discriminator}}public data class{{/discriminator}} {{classname}}{{^discriminator}} (
+{{#nonPublicApi}}internal {{/nonPublicApi}}{{#discriminator}}interface{{/discriminator}}{{^discriminator}}public class{{/discriminator}} {{classname}}{{^discriminator}} private constructor(
```
AdamGrzybkowski: Curious but the `public` keyword has no condition so if we use `nonPublicApi` we will end up with `internal public class`, right?

gravatar/build.gradle.kts
AdamGrzybkowski: I like it 👍

AdamGrzybkowski (approved):
Great work with those templates! I left some comments though nothing critical.

gravatar/openapi/templates/data_class.mustache
hamorillo: Good catch! Fixed!

gravatar/openapi/templates/data_class.mustache
hamorillo: Good point. 

I've added the condition, but to be honest, I am not 100% sure that the kotlin generator supports it. I don't see it on the [kotlin docs](https://openapi-generator.tech/docs/generators/kotlin), and I couldn't use it🤷🏽‍♂️. Anyway, it looks better with the condition, just in case.

PR #207, opened 2024-07-03
#198/2 - Deprecate old model classes and update services to use the new ones

### LINKED ISSUE #198: `gravatar`- Improve SDK to allow REST API updates without breaking changes

hamorillo:
Closes #198 

### Description

This PR deprecates old model classes and updates services to use and return new model classes while keeping compatibility with previous versions. 

We are only refactoring `gravatar-core` classes. This approach allows us to verify that `gravatar-ui` elements and `demo-app` continue working as they should without any changes in the rest of the modules. 

### Testing Steps

- Run `demo-app` and verify everything works as expected

### DIFF
--- a/gravatar/src/main/java/com/gravatar/services/ProfileService.kt
+++ b/gravatar/src/main/java/com/gravatar/services/ProfileService.kt
@@ -103,7 +103,7 @@ public class ProfileService(okHttpClient: OkHttpClient? = null) {
      * @param hashOrUsername The hash or username to fetch the profile for
      * @return The fetched profile
      */
-    public suspend fun retrieve(hashOrUsername: String): Result<Profile, ErrorType> = runCatchingService {
+    public suspend fun retrieveCatching(hashOrUsername: String): Result<Profile, ErrorType> = runCatchingService {
         withContext(GravatarSdkDI.dispatcherIO) {
             val response = service.getProfileById(hashOrUsername)
             if (response.isSuccessful) {
@@ -130,8 +130,8 @@ public class ProfileService(okHttpClient: OkHttpClient? = null) {
      * @param email The email address to fetch the profile for
      * @return The fetched profiles
      */
-    public suspend fun retrieve(email: Email): Result<Profile, ErrorType> {
-        return retrieve(email.hash())
+    public suspend fun retrieveCatching(email: Email): Result<Profile, ErrorType> {
+        return retrieveCatching(email.hash())
     }
 
     /**
@@ -140,8 +140,8 @@ public class ProfileService(okHttpClient: OkHttpClient? = null) {
      * @param hash The hash to fetch the profile for
      * @return The fetched profiles
      */
-    public suspend fun retrieve(hash: Hash): Result<Profile, ErrorType> {
-        return retrieve(hash.toString())
+    public suspend fun retrieveCatching(hash: Hash): Result<Profile, ErrorType> {
+        return retrieveCatching(hash.toString())
     }
 
     /**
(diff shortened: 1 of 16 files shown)


gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -51,7 +58,7 @@ public class ProfileService(okHttpClient: OkHttpClient? = null) {
      * @param email The email address to fetch the profile for
      * @return The fetched profiles
      */
-    public suspend fun fetch(email: Email): Result<Profile, ErrorType> {
+    public suspend fun fetch(email: Email): Result<LegacyProfile, ErrorType> {
```
AdamGrzybkowski, line 66: Shouldn't we deprecate all the `fetch` functions?

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -71,7 +78,64 @@ public class ProfileService(okHttpClient: OkHttpClient? = null) {
...
         return fetch(username)
     }
+
+    /**
+     * Fetches a Gravatar profile for the given hash or username.
+     *
+     * @param hashOrUsername The hash or username to fetch the profile for
+     * @return The fetched profile
+     */
+    public suspend fun retrieve(hashOrUsername: String): Result<Profile, ErrorType> = runCatchingService {
```
AdamGrzybkowski: If we are about to change the names how do you feel about using [this naming convention](https://kotlinlang.org/docs/api-guidelines-consistency.html#choose-the-appropriate-error-handling-mechanism:~:text=Consider%20providing%20overloads,methods%20for%20channels.)?
>Consider providing overloads of functions, where one throws an exception, while the other wraps it in a result type instead. In these cases, use the Catching suffix to indicate that exceptions are caught in the function. For example, the standard library has the [run](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/run.html) and [runCatching](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/run-catching.html) functions using this convention, and the coroutines library has [receive](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-receive-channel/receive.html) and [receiveCatching](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-receive-channel/receive-catching.html) methods for channels.

If not now, something to consider in the future.

AdamGrzybkowski (commented):
Tested the demo apps and it built successfully. I think we need to deprecate more functions though - details in the comment below.

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
hamorillo: Good catch! Fixed!

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
hamorillo: I like the idea! I've renamed the `retrieve` methods to `retrieveCatching` in order to use the naming convention. I don't want to complicate this PR, so I would create a ticket to add the other methods later. Wdyt?

hamorillo: I've renamed the methods now because later will be more complicated if, for some reason, we release a new version in between. (I don't think that will happen, but just in case).

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -151,6 +151,6 @@ public class ProfileService(okHttpClient: OkHttpClient? = null) {
      * @return The fetched profiles
      */
     public suspend fun retrieveByUsername(username: String): Result<Profile, ErrorType> {
```
AdamGrzybkowski: I think this one would have to be renamed as well, right?

hamorillo: 🤦🏼‍♂️ Thanks @AdamGrzybkowski! Fixed.

AdamGrzybkowski (approved):
LGTM 👍

PR #208, opened 2024-07-04
Change the skeleton rounder corners value to 50 percent

### LINKED ISSUE #199: Fix: Skeleton height is not correctly set based on the text measure funtion

AdamGrzybkowski:
Closes #199

### Description

So the TextMeasurer works fine but we do use slightly different sizing compared to Figma that results in those visual differences of the Skeleton objects. 

While at it I've noticed the rounded corners aren't flexible enough to accomodate taller Skeleton objects so I've updated them to a percentage value. 

Here's the comparison from the screenshots test
Screenshot 2024-07-04 at 12 41 52

### Testing Steps

1. Start the demo app
2. Go to the profile tab
3. Tap the Skeleton button
4. Verify the corner radius works for all skeleton objects


hamorillo (approved):
Looks nice! 🚀

PR #209, opened 2024-07-04
`gravatar`- Update REST API with interest

### LINKED ISSUE #182: `gravatar`- Update REST API with `Interest`

hamorillo:
Closes #182 

### Description

This PR is an iteration over #184, where we discussed a better approach to avoiding breaking changes in the SDK when updating the REST API. Now, we shouldn't break anything here. 🤞🏼

Gravatar has recently added interest to its REST API, so this PR updates the models to retrieve the new field.

image

### Testing Steps

**Note:** The `interest` field is only provided in authenticated requests.

1. Run the demoApp with a valid API key
2. Retrieve a profile with interests and verify you can see them in the `Raw Profile`
3. Retrieve a profile without interest and verify you can see an empty array
4. Run the demoApp without an API key
5. Retrieve a profile with interests and verify you can't see them. You'll see the field as `null`.
6. Retrieve a profile without interests and verify you can't see them. You'll see the field as `null`.


gravatar/api/gravatar.api
```diff
@@ -322,8 +322,9 @@ public final class com/gravatar/restapi/models/CryptoWalletAddressKt {
 }
 
 public final class com/gravatar/restapi/models/GalleryImage {
-	public synthetic fun <init> (Ljava/net/URI;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public synthetic fun <init> (Ljava/net/URI;Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
```
hamorillo, line 325: **TIL** - In spite of the fact that we don't have a public constructor, I was worried about this change in the method so I've asked Gemini:

> A constructor declared as public synthetic fun <init> in a .api file would not be directly accessible from your code.
>
>The keyword **synthetic** indicates that this constructor is not a real part of the original code. It's generated by the compiler for internal purposes, often related to how Kotlin interacts with Java bytecode.

Similar to `@set:JvmSynthetic`.

AdamGrzybkowski:
I will review this one once we finish all the other PRs related to the model changes!

gravatar/api/gravatar.api
AdamGrzybkowski: Thanks for sharing! TIL

AdamGrzybkowski (approved):
LGTM 👍

PR #211, opened 2024-07-04
Fix: Fix blank space in LargeProfile without description

### LINKED ISSUE #210: Fix: Make LargeProfile responsive
There's a big gap when the description is empty same as it was with the [Profile component](https://github.com/Automattic/Gravatar-SDK-Android/issues/193). The LargeProfile should adjust its height properly.

AdamGrzybkowski:
Closes #210 

### Description

Same as with the `Profile` component the `LargeProfile` has an extra amount of empty space when the profile has no description. This PR fixes that.

Here's the screenshot comparison:
Screenshot 2024-07-04 at 14 37 51

On top of that, now that we have the extra `skeletonModifier` I was able to modify the previously added check in the `Profile` component. There's no need to do an extra if check. There are no UI changes - tests are passing. 

### Testing Steps

1. Remove description from your Gravatar profile
2. Open the demo app
3. Go to the profile tab
4. Fetch the profile
5. Confirm there's less blank space without description
6. Confirm the skeleton still looks ok and has proper spacing


hamorillo (commented):
Looks nice. We need to solve the conflict but apart from that looks good to me.

AdamGrzybkowski:
@hamorillo  done!

gravatar-ui/src/main/java/com/gravatar/ui/components/LargeProfile.kt
```diff
@@ -189,11 +189,42 @@ private fun LargeProfilePreview() {
...
+            ),
+        ),
+    )
+}
+
 // TODO Make this preview internal in a future major release
-@Deprecated(
-    "This function is deprecated and will be removed in a future release.",
-    level = DeprecationLevel.WARNING,
-)
```
AdamGrzybkowski: oh wait this is wrong!

AdamGrzybkowski: fixed

hamorillo (approved):
Looks great! 🚀

PR #206, opened 2024-07-02
#198/3 - Update UI components to use new model classes

hamorillo:
Closes #198

### Description

This PR updates the `gravatar-ui` module to use the new model classes by default while keeping retro-compatibility. All UI elements should be usable with the legacy models and should look identical. (I've added some screenshot tests to ensure that.)

**Note:** We need to review and update our documentation if needed.

You can see the diffs between the new SDK APIs and the initial one (which is the 1.0.0 release). We are only adding, so we should be binary-compatible.

<details>
<summary> gravatar.api  - diff </summary>

```diff
diff --git a/gravatar/api/gravatar.api b/gravatar/api/gravatar.api
index dbf2780..7c02af9 100644
--- a/gravatar/api/gravatar.api
+++ b/gravatar/api/gravatar.api
@@ -282,12 +282,275 @@ public final class com/gravatar/api/models/VerifiedAccount {
 
 public final class com/gravatar/extensions/UserProfileExtensionsKt {
 	public static final fun avatarUrl (Lcom/gravatar/api/models/Profile;Lcom/gravatar/AvatarQueryOptions;)Lcom/gravatar/AvatarUrl;
+	public static final fun avatarUrl (Lcom/gravatar/restapi/models/Profile;Lcom/gravatar/AvatarQueryOptions;)Lcom/gravatar/AvatarUrl;
 	public static synthetic fun avatarUrl$default (Lcom/gravatar/api/models/Profile;Lcom/gravatar/AvatarQueryOptions;ILjava/lang/Object;)Lcom/gravatar/AvatarUrl;
+	public static synthetic fun avatarUrl$default (Lcom/gravatar/restapi/models/Profile;Lcom/gravatar/AvatarQueryOptions;ILjava/lang/Object;)Lcom/gravatar/AvatarUrl;
+	public static final fun defaultProfile (Ljava/lang/String;Ljava/lang/String;Ljava/net/URI;Ljava/net/URI;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Lcom/gravatar/restapi/models/ProfilePayments;Lcom/gravatar/restapi/models/ProfileContactInfo;Ljava/util/List;Ljava/lang/Integer;Ljava/time/Instant;Ljava/time/Instant;)Lcom/gravatar/restapi/models/Profile;
[...]

PR #213, opened 2024-07-05
Fix broken screenshot tests

hamorillo:
### Description

Some screenshot tests are broken because when merging #204, #207, and #206 the trunk had been changed with #208 in between.

### Testing Steps

- CI should be enough


AdamGrzybkowski (approved):

AdamGrzybkowski:
BTW. We're a small team so maybe we could enable this setting
Screenshot 2024-07-05 at 10 26 49

It's not like we have a ton of PRs getting merged in a day.

hamorillo:
> BTW. We're a small team so maybe we could enable this setting 

Sounds good to me!

PR #216, opened 2024-07-09
Add non-catching `retrieve` methods

### LINKED ISSUE #212: Create retrieve methods throwing exception instead of catching them
https://github.com/Automattic/Gravatar-SDK-Android/pull/207#discussion_r1664026712

hamorillo:
Closes #212 

### Description

In a previous commit, we added the retrieveCatching methods, which prevent them from throwing exceptions. Now, we are adding the throwing exception version of them based on [this discussion](https://github.com/Automattic/Gravatar-SDK-Android/pull/207#discussion_r1664026712).

### Testing Steps
- CI and code review should be enough

### DIFF
--- a/gravatar/src/main/java/com/gravatar/services/ProfileService.kt
+++ b/gravatar/src/main/java/com/gravatar/services/ProfileService.kt
@@ -99,11 +99,12 @@ public class ProfileService(okHttpClient: OkHttpClient? = null) {
 
     /**
      * Fetches a Gravatar profile for the given hash or username.
+     * This method throws any exception that occurs during the execution.
      *
      * @param hashOrUsername The hash or username to fetch the profile for
      * @return The fetched profile
      */
-    public suspend fun retrieveCatching(hashOrUsername: String): Result<Profile, ErrorType> = runCatchingService {
+    public suspend fun retrieve(hashOrUsername: String): Result<Profile, ErrorType> =
         withContext(GravatarSdkDI.dispatcherIO) {
             val response = service.getProfileById(hashOrUsername)
             if (response.isSuccessful) {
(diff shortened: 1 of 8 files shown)


gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -122,35 +123,81 @@ public class ProfileService(okHttpClient: OkHttpClient? = null) {
...
+    }
+
+    /**
+     * Fetches a Gravatar profile for the given email address.
+     * This method throws any exception that occurs during the execution.
+     *
+     * @param email The email address to fetch the profile for
+     * @return The fetched profiles
+     */
+    public suspend fun retrieve(email: Email): Result<Profile, ErrorType> {
```
AdamGrzybkowski:
```suggestion
    public suspend fun retrieve(email: Email): Profile {
```

I think that we should return only when the operation is successful thus we don't need the `Result` return type. All errors should be thrown. Otherwise, users will have to implement two error handling for this method. Thoughts on this?

hamorillo: > users will have to implement two error handling for this method

Yeah, that's a good point. We'll need to figure out what exceptions to throw and whether we need to declare our own exceptions for each of the current ErrorTypes. Anyway, I think it's a great suggestion. I'll iterate over it. Thanks @AdamGrzybkowski .

hamorillo: I've pushed some changes. It should be ready for another round. 🙏🏼

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -99,58 +99,99 @@ public class ProfileService(okHttpClient: OkHttpClient? = null) {
...
-                Logger.w(
-                    LOG_TAG,
-                    "Network call unsuccessful trying to get Gravatar profile: $response.body",
-                )
-                Result.Failure(errorTypeFromHttpCode(response.code()))
-            }
+    public suspend fun retrieve(hashOrUsername: String): Profile = withContext(GravatarSdkDI.dispatcherIO) {
+        val response = service.getProfileById(hashOrUsername)
+        if (response.isSuccessful) {
+            response.body() ?: error("Response body is null")
```
AdamGrzybkowski, line 111: Quick question: 

Does `service.getProfileById(hashOrUsername)` always return a value even if the hash is not matched?  It will return an empty profile, right?

gravatar/src/main/java/com/gravatar/services/ServiceUtils.kt
```diff
@@ -3,6 +3,11 @@ package com.gravatar.services
 import kotlinx.coroutines.CancellationException
 import com.gravatar.di.container.GravatarSdkContainer.Companion.instance as GravatarSdkDI
 
+@Deprecated(
+    "This method is deprecated and should be removed in a future release.",
+    replaceWith = ReplaceWith("com.gravatar.services.runCatchingRequest"),
+    level = DeprecationLevel.WARNING,
+)
 internal inline fun <T> runCatchingService(block: () -> Result<T, ErrorType>): Result<T, ErrorType> {
```
AdamGrzybkowski, line 11: This is internal - do we have to deprecate it? Can't we remove it?

gravatar/src/test/java/com/gravatar/services/ProfileServiceTests.kt
```diff
@@ -212,4 +203,99 @@ class ProfileServiceTests {
         coVerify(exactly = 1) { containerRule.gravatarApiMock.getProfileById(usernameEmail.hash().toString()) }
         assertTrue(loadProfileResponse is Result.Success)
     }
+
+    @Test
+    fun `given an username when retrieving its profile and an exception is thrown then result is failure`() = runTest {
```
AdamGrzybkowski:
```suggestion
    fun `given a username when retrieving its profile and an exception is thrown then result is failure`() = runTest {
```

gravatar/src/test/java/com/gravatar/services/ProfileServiceTests.kt
```diff
@@ -212,4 +203,99 @@ class ProfileServiceTests {
...
+        } throws Exception()
+
+        val loadProfileResponse = profileService.retrieveCatching(usernameHash)
+        coVerify(exactly = 1) { containerRule.gravatarApiMock.getProfileById(usernameHash.toString()) }
+        assertTrue((loadProfileResponse as Result.Failure).error == ErrorType.UNKNOWN)
+    }
+
+    // Throwing Exception Version of the methods
+    @Test(expected = TimeoutException::class)
+    fun `given an username when retrieving its profile and a timeout occurs then exception is thrown`() = runTest {
```
AdamGrzybkowski:
```suggestion
    fun `given a username when retrieving its profile and a timeout occurs then exception is thrown`() = runTest {
```

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
hamorillo: The endpoint returns a `404` if the hash is not matched. We match that `404` to an empty state

image

gravatar/src/main/java/com/gravatar/services/ServiceUtils.kt
hamorillo: I didn't remove it because is still in use [here](https://github.com/Automattic/Gravatar-SDK-Android/blob/1f754d9e778df05266e9810d432cc26695f44c07/gravatar/src/main/java/com/gravatar/services/ProfileService.kt#L34) so I think we can remove it when we remove the other deprecated methods.

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
AdamGrzybkowski: >We match that 404 to an empty state

Do we match that in the SDK? I can only see that in the demo app. 


I think we shouldn't throw an error if it is expected not to get a profile. It's probably better to return null instead.

hamorillo: Updated. Let me know what you think.

AdamGrzybkowski (approved):
LGTM! 👍 Thanks for all the changes.

PR #221, opened 2024-07-16
Migrate dependency management to version catalogs

AdamGrzybkowski:
### Description
I wanted to update libraries to the latest versions but I figured it would be beneficial to migrate to the version catalogs first. 

We had all of the dependencies specified in each grade file and with each new module the pain of updating the libraries will increase. We also had some discrepancies so that should fix it. 

### Testing Steps
CI, changes review, run the demo app

### DIFF
--- a/gravatar-ui/build.gradle.kts
+++ b/gravatar-ui/build.gradle.kts
@@ -1,19 +1,16 @@
 import org.jetbrains.dokka.gradle.DokkaTaskPartial
 
 plugins {
-    id("com.android.library")
-    id("org.jetbrains.kotlin.android")
-    id("org.jetbrains.kotlin.plugin.compose")
-    id("com.automattic.android.publish-to-s3")
-
+    id(libs.plugins.android.library.get().pluginId)
+    alias(libs.plugins.kotlin.android)
+    alias(libs.plugins.kotlin.plugin.compose)
+    alias(libs.plugins.publish.to.s3)
     // Ktlint
-    id("org.jlleitschuh.gradle.ktlint")
-
+    alias(libs.plugins.ktlint)
     // Detekt
-    id("io.gitlab.arturbosch.detekt")
-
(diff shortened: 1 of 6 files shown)


gravatar-ui/build.gradle.kts
```diff
@@ -1,19 +1,16 @@
 import org.jetbrains.dokka.gradle.DokkaTaskPartial
 
 plugins {
-    id("com.android.library")
-    id("org.jetbrains.kotlin.android")
-    id("org.jetbrains.kotlin.plugin.compose")
-    id("com.automattic.android.publish-to-s3")
-
+    id(libs.plugins.android.library.get().pluginId)
```
AdamGrzybkowski, line 4: I could not use the `alias()` here, the Gradle sync would error out with [this issue](https://github.com/gradle/gradle/issues/20084).

hamorillo: TIL

hamorillo (approved):
Great improvement! I've been thinking about doing this for some time but never found a good moment. Thanks!

gravatar-ui/build.gradle.kts
davidburstrom: I reckon you could have used `alias` here if you'd added

```
alias(libs.plugins.android.library) apply false
```

to the root `build.gradle.kts` too, just like it's done for `alias(libs.plugins.android.application) apply false`.

hamorillo: You are right. I've opened a ticket (#432) and PR (#433), so feel free to leave any comments there.

Thanks for your suggestions, @davidburstrom; we are delighted to receive external feedback and contributions!

davidburstrom: My pleasure! I happened to come across it as @AdamGrzybkowski linked to the Gradle issue, and I also had a solution for it.

PR #228, opened 2024-07-16
Update dependencies

AdamGrzybkowski:
### Description

This PR updates Gradle, AGP, and all dependencies.

Additionally, I've cleaned the unused Android Test dependencies. 

There's an issue when I try to bump the robolectric. The tests on CI fail but it works for me locally. I'm skipping it for now to update the rest of the dependencies. 

### Testing Steps
Smoke test with the demo app


.idea/misc.xml
```diff
@@ -1,9 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ExternalStorageConfigurationManager" enabled="true" />
   <component name="PDMPlugin">
     <option name="skipTestSources" value="false" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" default="true" project-jdk-name="jbr-17" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" default="true" project-jdk-name="17" project-jdk-type="JavaSDK">
```
hamorillo: ❓I know this comes from before, but I think it would be better to remove this file from the repo and not track it.

gradle/libs.versions.toml
```diff
@@ -1,30 +1,28 @@
...
 desugarJdk = "2.0.4"
 junit = "4.13.2"
 mockk = "1.13.9"
-espresso = "3.5.1"
-androidXJunit = "1.1.5"
-appcompat = "1.6.1"
-material = "1.11.0"
+appcompat = "1.7.0"
+material = "1.12.0"
 ucrop = "2.2.8"
```
hamorillo:
```suggestion
ucrop = "2.2.9"
```

AdamGrzybkowski: Yeah, I've noticed AS doesn't always suggest the latest available version 🤔 not sure why it lags but it didn't highlight `ucrop` at all for me.

.idea/misc.xml
AdamGrzybkowski: agree 👍

gradle/libs.versions.toml
hamorillo: Yeah, I don't know why that happens sometimes. In this case, it was highlighted for me.

hamorillo (approved):
Thanks for the changes! 🚀

PR #229, opened 2024-07-16
Extract screenshot test utils to an external module

### LINKED ISSUE #227: QuickEditor: Extract `RoborazziTest` to a module to share it between different modules

hamorillo:
Closes #227 

### Description

As we'll add new modules with UI features, it would be nice to have a shared module between them to access screenshot test utils, like `RoborazziTest.kt`, which helps us create a screenshot test for `@Composable`.

This PR creates the new module and migrates the screenshot test utils to that module. 

Everything should work as before.

**Note:** The run button in the IDE is not working as expected. I'm not sure why yet. As the gradlew task works fine, I think it's not blocking this PR. I'll try to dig more into that.

### Testing Steps

- CI and code review should be enough. (You can also execute the screenshot tests locally)


uitestutils/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
+plugins {
+    id("com.android.library")
+    id("org.jetbrains.kotlin.android")
+    id("org.jetbrains.kotlin.plugin.compose")
+
+    // Ktlint
+    id("org.jlleitschuh.gradle.ktlint")
+
+    // Detekt
+    id("io.gitlab.arturbosch.detekt")
```
AdamGrzybkowski: We can use the `libs.plugins....` from the version catalog here as well

uitestutils/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+
+dependencies {
+
+    // Jetpack Compose
+    implementation(platform(libs.compose.bom))
+    implementation(libs.compose.ui)
+    implementation(libs.compose.ui.tooling.preview)
+    implementation(libs.compose.material3)
+    implementation(libs.androidx.activity.compose)
+    debugImplementation(libs.androidx.compose.ui.tooling)
```
AdamGrzybkowski: Do we need all of those here?

hamorillo: Nop. Removed.

uitestutils/build.gradle.kts
hamorillo: Done

AdamGrzybkowski (approved):

PR #232, opened 2024-07-18
QuickEditor: SelectableAvatar Composable

### LINKED ISSUE #220: QuickEditor: Avatar selection composable in `quickeditor` module
image

hamorillo:
Closes #220 

### Description

This PR adds an internal composable to be used in the Native Avatar Picker. As we are still working on the final design, this can change a bit, but this PR also allows us to set up Screenshot testing for this new module.

image
image

### Testing Steps

- You can run the `@Preview`s in a device in other to see how they look:
image
image

- Code Review and CI


gravatar-quickeditor/build.gradle.kts
```diff
@@ -15,7 +18,8 @@ android {
 
     defaultConfig {
         minSdk = 21
-
+        // targetSdkVersion has no effect for libraries. This is only used for the test APK
+        targetSdk = 34
```
hamorillo, line 22: ❕Without the `targetSdk`, screenshot tests don't render the images.

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/GravatarScreenshotTest.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import androidx.compose.material3.Surface
+import androidx.compose.runtime.Composable
+import com.gravatar.ui.GravatarTheme
+import com.gravatar.uitestutils.RoborazziTest
+
+internal fun RoborazziTest.gravatarScreenshotTest(composable: @Composable () -> Unit) {
+    screenshotTest {
+        GravatarTheme {
+            Surface {
```
hamorillo, line 11: ‼️ I've added the surface here because I had several problems with CI screenshots. I noticed that tests were indeed failing by just a few pixels (even though they worked locally):

```
<system-out>
Roborazzi: com.gravatar.quickeditor.ui.components.SelectableAvatarTest.selectableAvatarNotSelected.png The differ result :ComparisonResult(pixelDifferences=48, pixelCount=170569, width=413, height=413) changed:true Roborazzi: com.gravatar.quickeditor.ui.components.SelectableAvatarTest.selectableAvatarSelected.png The differ result :ComparisonResult(pixelDifferences=71, pixelCount=170569, width=413, height=413) changed:true
</system-out>
```

Rendering the `@Composable` under test over a `Surface` solves the issue. I would prefer not to do it, but I haven't found a better solution, so I think it's acceptable.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
```diff
@@ -0,0 +1,67 @@
...
+            model = imageUrl,
+            contentDescription = stringResource(id = R.string.selectable_avatar_content_description),
+            modifier = modifier
+                .clip(RoundedCornerShape(cornerRadius))
+                .then(
+                    if (isSelected) {
+                        Modifier.border(7.dp, MaterialTheme.colorScheme.primary, RoundedCornerShape(cornerRadius))
+                    } else {
+                        Modifier.border(2.dp, MaterialTheme.colorScheme.surfaceDim, RoundedCornerShape(cornerRadius))
+                    },
```
hamorillo, line 37: I picked this color based on a design palette. We will probably need to iterate on this when we have the final designs. 

Also, could it be a good moment to define a set of colors in `GravataTheme`? wdyt?

AdamGrzybkowski: >Also, could it be a good moment to define a set of colors in GravataTheme?


Yeah, I think so. I've noticed we are using the default ones, right?

hamorillo: Yes, we currently use `material3` default ones for light and dark themes. We can open a different PR with that.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
```diff
@@ -0,0 +1,57 @@
...
+import androidx.compose.ui.unit.dp
+import coil.compose.AsyncImage
+import com.gravatar.quickeditor.R
+import com.gravatar.ui.GravatarTheme
+
+@Composable
+private fun SelectableAvatar(
+    imageUrl: String,
+    isSelected: Boolean,
+    onAvatarClicked: () -> Unit,
```
AdamGrzybkowski, line 22: I believe Avatars will have some sort of `id ` that we will have to pass here. We can add it later though.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
```diff
@@ -0,0 +1,57 @@
...
+    modifier: Modifier = Modifier,
+) {
+    val cornerRadius = 7.dp
```
AdamGrzybkowski, line 25: I think this one will be increased based on the current design iteration but let's wait for the final one.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+@Preview
+@Composable
+private fun SelectableAvatarNotSelectedPreview() {
+    SelectableAvatar(
+        "https://gravatar.com/avatar/fd2188b818f15e629f7b62896b5c6075?s=250",
+        false,
+        { Log.i("SelectableAvatar", "Clicked") },
```
AdamGrzybkowski: Do you want to keep the log in the Preview?

gravatar-quickeditor/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="selectable_avatar_content_description">Avatar</string>
```
AdamGrzybkowski:
```suggestion
    <string name="selectable_avatar_content_description">Avatar Image</string>
```

Maybe we could also add an index, though not sure this would be more useful.

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/GravatarScreenshotTest.kt
AdamGrzybkowski: That's great you were able to fix that! 

Do you know why this happens?

hamorillo: I wonder if it could be related to how the transparencies are rendered, but to be completely honest, I don't know. 😖

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
hamorillo: Yeah, we may need to modify that param later and replace it with the `Avatar` itself, the `id` or the `ImageUrl`.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
hamorillo: I've used it when I run the `@Preview` on a device, but I can remove it. It doesn't add much value.

gravatar-quickeditor/src/main/res/values/strings.xml
hamorillo: We can use the current `altText` when we have it. If we don't have it, I would use this literal. 

Yeah, I don't think the index helps a lot.

AdamGrzybkowski (approved):

PR #239, opened 2024-07-22
Initial code for the QuickEditor that should start with the OAuth 

AdamGrzybkowski:
### Description
~I am opening this as a Draft to discuss the API and the structure in general. The code is not final yet.~

This is the module structure for now:
Screenshot 2024-07-24 at 12 47 47

- `oauth` package will hold all the OAuth related code (UI and network)
-  `editor` will contain what's related to the Editor including the possible error and result updates

**Integration**

Jetpack Compose
```
GravatarQuickEditorBottomSheet(
            gravatarQuickEditorParams = GravatarQuickEditorParams {
                appName = "yourAppName"
                oAuthParams = OAuthParams {
                    clientId = "clientId
                    clientSecret = "clientSecret"
                    redirectUri = "redirectUri"
                }
                scope = GravatarQuickEditorScope.AVATARS
            },
            onDismiss = {            },
   )
```

Activity/Fragment 
```
GravatarQuickEditor.show(
                activity = this,
                gravatarQuickEditorParams = GravatarQuickEditorParams {
                    appName = "yourAppName"
                    oAuthParams = OAuthParams {
                        clientId = "clientId"
                        clientSecret = "clientSecret"
                        redirectUri = "redirectUri"
                    }
                    scope = GravatarQuickEditorScope.AVATARS
                },
                onDismiss = {                },
            )
)
```

For the `OAuthParams` class I've used the same pattern as we use for the classes generated by OpenAPI - meaning the builder for Java and DSL for Kotlin. 

~About the `scope = GravatarQuickEditorScope.AVATARS` - I've followed the web here - they offer different scopes to edit by the user. Not sure if we are planning to have more than just Avatars so we could drop it for now.~

`I will continue with cleaning up and extracting the `oauth` package. We need ViewModel and separate services but let me know what you think so far. ` - this is handled in a separate [PR](https://github.com/Automattic/Gravatar-SDK-Android/pull/240)

### Testing Steps

Add these to your `local.properties` 
```
demo-app.wordpress.oauth.clientId=yourId
demo-app.wordpress.oauth.clientSecret=yourSecret
demo-app.wordpress.oauth.redirectUri=yourURI
```

and play with the demo app. 

1. Tap the `UpdateAvatar`
2. Continue with the required steps to finish the OAuthFlow
3. `Insert the real avatar picker page` text should appear - this is a placeholder for the avatar picker UI
4. Tap the above text
5. Toast message should appear with a confirmation
6. Tap the `UpdateAvatar` 
7. Close the browser with the `x` button
8. Tap the `Done` button
9. Toast with `Dissmised` info is shown
10. Tap the `UpdateAvatar` 
11. Tap `Deny`
12. Toast with `Error` info is shown


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -0,0 +1,60 @@
...
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import com.gravatar.quickeditor.ui.oauth.OAuthPage
+import com.gravatar.quickeditor.ui.oauth.OAuthParams
+
+@Composable
+internal fun GravatarQuickEditorPage(
```
AdamGrzybkowski, line 31: Internal for now, though we can open it later. That will open the possibility of using the quick editor flow in a new activity for example.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @param gravatarQuickEditorParams The Gravatar Quick Editor parameters.
+ * @param onDismiss The callback for when the bottom sheet is dismissed.
+ * @param modalBottomSheetState The state of the bottom sheet.
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+public fun GravatarQuickEditorBottomSheet(
+    gravatarQuickEditorParams: GravatarQuickEditorParams,
+    onDismiss: () -> Unit,
```
AdamGrzybkowski: What other callback we might want to provide here? 
We will likely need something like `onProfileUpdated`. I wonder if we want to communicate the OAuth errors here as well or do we want to keep it within the SDK and handle ourselves fully?

gravatar-quickeditor/build.gradle.kts
```diff
@@ -87,15 +96,26 @@ android {
...
-    testImplementation(libs.junit)
-    testImplementation(project(":uitestutils"))
+    implementation(libs.coil.compose)
+    implementation(libs.androidx.browser)
+    implementation(libs.retrofit)
+    implementation(libs.retrofit.gson.converter)
+
+    implementation(libs.android.material)
+    implementation(libs.androidx.appcompat)
+    implementation(libs.androidx.constraintlayout)
```
hamorillo: I think we don't need:

```
implementation(libs.android.material)
implementation(libs.androidx.constraintlayout)
```

Also, we are only using `implementation(libs.androidx.appcompat)` for the `GravatarQuickEditor.show(fragment)`. That method could be convenient in some legacy apps, but it will add that dependency to all projects. If my app is only Compose, I'll add a library that I don't need for anything. I think we should try to limit our dependencies when possible, so I am not sure if, in this case, it is worth it. Another option could be to provide a different module containing those utils for the legacy view system, but we need to evaluate if creating a "compatibility" module makes sense.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
hamorillo: > We will likely need something like onProfileUpdated

Yeah. Do we want to be less generic for the moment? For example,  `onAvatarUpdated`, and if we later add the possibility of modifying the profile, we'll add a different callback. 

> I wonder if we want to communicate the OAuth errors here as well or do we want to keep it within the SDK and handle ourselves fully?

Good question. I'm trying to use the third-party developer hat, and I guess it could be useful to know,  at least, that maybe I decide to show something in the UI? However, I wouldn't be too specific with the error details.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Quick Editor's configuration parameters
+ *
+ * @property appName The App Name that is launching the Gravatar Quick Editor
+ * @property oAuthParams The OAuth parameters for your WP.com application
+ * @property scope The scope of the Quick Editor
+ */
+public class GravatarQuickEditorParams private constructor(
+    public val appName: String,
+    public val oAuthParams: OAuthParams,
```
hamorillo: 🤔 I think we shouldn't include the `OAuthParams` inside `GravatarQuickEditorParams`. In my mind, being totally strict, the `QuickEditor` doesn't need the OAuthParams. It needs a token to connect with the backend. For that reason, I would consider keeping it outside this class. 

That way, we can have two `GravatarQuickEditorPage` composables. One when we already have the token, for example, if we finally decide to save it locally or the app has already the token:

```
@Composable
internal fun GravatarQuickEditorPage(
    gravatarQuickEditorParams: GravatarQuickEditorParams,
    oauthToken: String
    onAuthError: () -> Unit = {},
) {
```

And another option if want to go through the OAuth flow:

```
@Composable
internal fun GravatarQuickEditorPage(
    gravatarQuickEditorParams: GravatarQuickEditorParams,
    oAuthParams: OAuthParams,
    onAuthError: () -> Unit = {},
) {
```

hamorillo (commented):
Great job @AdamGrzybkowski  🎉!

I like the overall structure. It's easy to understand at first sight.

> About the scope = GravatarQuickEditorScope.AVATARS - I've followed the web here - they offer different scopes to edit by the user. Not sure if we are planning to have more than just Avatars so we could drop it for now.

We can ask around, and if we think we'll add more scopes, I'm not against adding it right now. If we add it now, we won't have changes in the public API later, and it will be easier to handle. However, if we have doubts about whether we'll add more scopes, I will drop it, and we'll handle it later.

gravatar-quickeditor/build.gradle.kts
AdamGrzybkowski: Oh you might be right, AS must have added this one when I added an activity while testing some stuff. 

> That method could be convenient in some legacy apps, but it will add that dependency to all projects.

I wouldn't call those legacy apps. It's probably the majority by a big margin. Jetpack Compse is still in a very early stage so it's almost certain users will need a way to use it from non-compose code.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
AdamGrzybkowski: Yeah, I can extract it. The main idea was to have fewer params (with callbacks this will grow quite a bit) and to have one object for the quick editor configuration.

>One when we already have the token, for example, if we finally decide to save it locally or the app has already the token:

Whether we store it or not the token might still expire, and in such a case, we won't be able to handle this. As long as we are handling the token ourselves (which is what we will do I believe) we need that OAuth config.

I guess one specific case is the Jetpack app, this one will have the token already, right?

gravatar-quickeditor/build.gradle.kts
AdamGrzybkowski: What I'm thinking is that - these extensions are more likely to be used than the Jetpack Compose components, so I think we shouldn't require devs to specify extra dependency to get those. 

Anyway, I think I will extract those into a separate PR anyway. So we can discuss it later.

hamorillo: > I wouldn't call those legacy apps. 

I agree. I didn't choose the best words. 

> we shouldn't require devs to specify extra dependency to get those

They would need to add only one dependency, right?

> I think I will extract those into a separate PR anyway. So we can discuss it later.

👍

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
AdamGrzybkowski: >Yeah. Do we want to be less generic for the moment? For example, onAvatarUpdated, and if we later add the possibility of modifying the profile, we'll add a different callback.


If we decide to go for the scopes I was thinking about something like `onProfileUpdated(ProfilePayload)` -> The `ProfilePayload` would then correspond to the scope selected.

Having a simple `onProfileUpdated` and returning a full profile would be simpler for third-party devs to handle I guess, but we won't have that. We might not even have access to the full profile. 

>Good question. I'm trying to use the third-party developer hat, and I guess it could be useful to know, at least, that maybe I decide to show something in the UI? However, I wouldn't be too specific with the error details.

Yeah, we probably need at least one error to propagate when the OAuth fails - as this could be a config issue.

AdamGrzybkowski:
>However, if we have doubts about whether we'll add more scopes, I will drop it, and we'll handle it later.


I will drop it for now

AdamGrzybkowski:
@hamorillo 
I've updated the PR with your feedback. 
As mentioned in another comment I have dropped the `GravatarQuickEditor.kt` file with the Activity/Fragment.

The main changes to the `GravatarQuickEditorBottomSheet` are listed in the commit message:

    - Dropping the Scope param for the editor. We will stick with one for now.
    - Extracting the OAuthParam from the GravatarQuickEditorParams
    - Deleting the GravatarQuickEditorParams. It had only one property at this point
    - Adding onAvatarUpdateResult: (AvatarUpdateResult) -> Unit callback

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/AvatarUpdateResult.kt
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Objects
+
+/**
+ * All possible results for the Avatar update flow
+ */
+public sealed class AvatarUpdateResult {
+    /**
+     * Quick Editor dismissed without any errors or updates to the Gravatar
+     */
+    public data object Dismissed : AvatarUpdateResult() {
```
AdamGrzybkowski: `data object` doesn't generate `copy()` or `componentN()`  functions so it's safe to use. This can be verified in the `gravatar-quickeditor.api` file.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
AdamGrzybkowski: >I guess one specific case is the Jetpack app, this one will have the token already, right?

I think for this, since we won't have to launch the OAuth at all we would create a new bottom sheet that will open the Avatar editor directly without all the OAuth -> Editor navigation logic. So we would pretty much skip the `GravatarQuickEditorPage` completely. Thoughts?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -16,29 +18,34 @@ import com.gravatar.quickeditor.ui.oauth.OAuthParams
 
 @Composable
 internal fun GravatarQuickEditorPage(
-    gravatarQuickEditorParams: GravatarQuickEditorParams,
-    onAuthError: () -> Unit = {},
+    appName: String,
+    oAuthParams: OAuthParams,
+    onAvatarSelected: (Uri) -> Unit,
```
hamorillo: 🤔 I'm thinking about whether `GravatarQuickEditorPage` should have a callback with `AvatarUpdateResult` and extract the `Dismissed` from that class. 

The `Dismissed` state only applies to the `BottomSheet`, and it's not directly related to the avatar update, so maybe a different callback only in the `GravatarQuickEditorBottomSheet` could make sense?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
hamorillo: > So we would pretty much skip the GravatarQuickEditorPage completely. 

Yeah, we will probably open the already authorized content of the `GravatarQuickEditorPage` that will be extracted in a different composable.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
AdamGrzybkowski: >The Dismissed state only applies to the BottomSheet

Yeah, that's why I kept two separate callbacks. But, the dismissed state doesn't only apply to `BottomSheet`, there's no reason we couldn't have a `cancel` button within the Page itself. 


>so maybe a different callback only in the GravatarQuickEditorBottomSheet could make sense?

What do you mean by that?

hamorillo: > But, the dismissed state doesn't only apply to BottomSheet, there's no reason we couldn't have a cancel button within the Page itself.

I agree; at some point, we can have a cancel button inside the `GravatarQuickEditorPage,` but that button will cancel the whole QuickEditor, right? If we add a profile edition to the `GravatarQuickEditorPage,` tapping that button will dismiss everything, not just the avatar stuff. That is why I think, the `Dismissed` results belong to `GravatarQuickEditorPage` / `GravatarQuickEditorBottomSheet` or both, but not to the `AvatarUpdateResult`.

What I was trying to say with the different callbacks is the following:

```
public fun GravatarQuickEditorBottomSheet(
    appName: String,
    oAuthParams: OAuthParams,
    onAvatarUpdateResult: (AvatarUpdateResult) -> Unit,
    onDismiss: () -> Unit,
    modalBottomSheetState: SheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true),
) {
```

In fact, it's possible that we need to call several times `onAvatarUpdateResult`, because I tap over different avatars in the QuickEditor and then I manually dismiss the bottomSheet.

AdamGrzybkowski: >In fact, it's possible that we need to call several times onAvatarUpdateResult, because I tap over different avatars in the QuickEditor and then I manually dismiss the bottomSheet.


Oh right! I forgot about that. Thanks!

>That is why I think, the Dismissed results belong to GravatarQuickEditorPage / GravatarQuickEditorBottomSheet or both, but not to the AvatarUpdateResult.

So the idea was that it would be changed to a more generic result later when we have more options than avatar. Something like `QuickEditorResult`. But the above point about multiple invocations makes sense. I will re-work that.

AdamGrzybkowski: Ok, I've modified it to

```
public fun GravatarQuickEditorBottomSheet(
    ...
    onAvatarSelected: (AvatarUpdateResult) -> Unit,
    onDismiss: (error: GravatarQuickEditorError?) -> Unit = {},
)
```

As per the docs

>- **onAvatarSelected** The callback for the avatar update result, check [AvatarUpdateResult]. Can be invoked multiple times while the Quick Editor is open.
>- **onDismiss** The callback for the dismiss action. [GravatarQuickEditorError] will be non-null if the dismiss was caused by an error.

The same callbacks are now present in the `GravatarQuickEditorPage` for consistency.

Let me know what you think

hamorillo: Thanks for the changes! 

I'm not sure about the `error` as a parameter on the `onDismiss` callback. I have some doubts about how I will handle the following case:

1. I tap in an avatar in the `GravatarQuickEditorPage`
2. The avatar is updated, so we'll receive the `onAvatarSelected` callback
3. For some reason, I lose my internet connection (any other error can replace this)
4. The `GravatarQuickEditorPage` is still open, so I select a different avatar
- Are we going to dismiss the `GravatarQuickEditorPage`? (I would say we shouldn't)
- If we don't dismiss the `GravatarQuickEditorPage`, how do we inform the app? (In case we want to notify the error. I think we should)

Maybe we should have a callback to inform the app of both the success and failure of the avatar update process and another callback to inform about the dismissal of the page/bottom sheet. Wdyt?

AdamGrzybkowski: Great example! 

I think in your example the error should be handled completely by us and we should properly inform the user about a network problem when selecting/uploading an avatar (the end user, not the app). I don't see a lot of value in sending those errors to the app, because there's nothing they can do about it. 

There's another group of errors though. Those that we can't handle ourselves e.g. wrong OAuth params. The app should be notified about those errors and handle properly. 

So to sum up, we have two groups (how I see it:
- errors happening during the flow related to our Gravatar service
- errors terminating the flow completely because there's not much we can do about it

Maybe the naming is wrong and the `onDismiss` param should be called `dismissReason` not an error.

hamorillo: > Maybe the naming is wrong and the onDismiss param should be called dismissReason not an error.

Great idea. I think `dismissReason` is more understandable for those unsolvable flows.

> I don't see a lot of value in sending those errors to the app, because there's nothing they can do about it.

That's a fair point. It is true that in the majority of the cases, they probably don't want to do anything. However, I wonder if there could be cases where the app wants to update its UI. 

For example, in the PocketCast app, we may want to show a loading state over the avatar while we are updating or uploading the new avatar. At the moment, we are not notifying the app about it, but we would like to do it at some point. Is it possible that the app decides to finish the flow after a network error?

I may be overthinking this, but I guess it is better now rather than when we have to deal with API changes 😅

AdamGrzybkowski: >I may be overthinking this, but I guess it is better now rather than when we have to deal with API changes 😅

Definitely! 

> For example, in the PocketCast app, we may want to show a loading state over the avatar while we are updating or uploading the new avatar. At the moment, we are not notifying the app about it, but we would like to do it at some point. Is it possible that the app decides to finish the flow after a network error?

hmm I still think we would want to keep the user within the QuickEditor while uploading. Uploading != Selecting - as per the web flow. So the loading UI will be handled in the QE itself, there's no need for third-party apps to show it as well.

hamorillo: I was thinking of a screen like:
Alt Text

I may be interested in using the loading state in the avatar behind the bottomsheet (App context) while the avatar is being updated. But yeah, maybe you are right, and that's too much. 

Do you want to update the param name? I'll do another quick round before approval. :)

AdamGrzybkowski: >I may be interested in using the loading state in the avatar behind the bottomsheet (App context) while the avatar is being updated. But yeah, maybe you are right, and that's too much.


Maybe! :D But again Uploading is not selecting so there's no point in showing the loading there (in the app) when it would switch back to the same avatar.

AdamGrzybkowski: >Do you want to update the param name? I'll do another quick round before approval. :)

yes!

AdamGrzybkowski: PR updated

demo-app/src/main/res/layout/activity_quick_editor_test.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
```
hamorillo: ⛏ We don't need this layout in this PR.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
```diff
@@ -0,0 +1,98 @@
...
+                        }
+                    },
+                )
+                GravatarQuickEditorPage(
+                    appName = appName,
+                    oAuthParams = oAuthParams,
+                    onDismiss = onDismiss,
+                    onAvatarSelected = {
+                        onAvatarSelected(it)
+                    },
```
hamorillo: ⛏
```suggestion
                    onAvatarSelected = onAvatarSelected,
```

hamorillo (approved):
Left two minor comments but apart from that looks good to me! Thanks for all the changes. 🚀

PR #240, opened 2024-07-24
Presentation logic for the OAuthPage

AdamGrzybkowski:
### Description
Another Draft PR to get some initial feedback. 

We haven't yet needed to do more complicated presentation logic so we don't have any base standards for that. 

This PR adds the ViewModel for the `OAuthPage`. 

I've decided to do manual DI to not add any more dependencies for now. 
One problem is that the `ViewModel` is scoped to the Activity so in our case we would always get the same `ViewModel`. To overcome that I've added the `navigation-compose` because each destination acts as a `ViewModelStoreOwner`. 

This allowed me to move some logic out of the composable like the `launchCustomTab()` call that was triggered in `DisposableEffect`.

One thing I had to add because of the `navigation-compose` is the `GravatarQuickEditorSplashPage`. It acts as the start destination in the nav graph and will handle the routing later when we store the token. There needs to be a `startDestination` defined so we need something and I don't want to start with the `oauth` route and close it quickly if we detect that the token is present. 

For the `ViewModel` itself, there's nothing extraordinary. One thing different than what we can be used to is that the `Actions`  emitted from `ViewModel` are separate and not part of the `UiState`. 

### Testing Steps

Add these to your local.properties

```
demo-app.wordpress.oauth.clientId=yourId
demo-app.wordpress.oauth.clientSecret=yourSecret
demo-app.wordpress.oauth.redirectUri=yourURI
```
and play with the demo app.

1. Tap the `UpdateAvatar`
2. Continue with the required steps to finish the OAuthFlow
3. `Insert the real avatar picker page` text should appear - this is a placeholder for the avatar picker UI
4. Tap the above text
5. Toast message should appear with a confirmation
6. Tap the `UpdateAvatar` 
7. Close the browser with the `x` button
8. Tap the `Done` button
9. Toast with `Dissmised` info is shown
10. Tap the `UpdateAvatar` 
11. Tap `Deny`
12. Toast with `Error` info is shown

### DIFF
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
@@ -34,10 +34,12 @@ import androidx.compose.material3.TextField
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.neverEqualPolicy
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -139,7 +141,7 @@ private fun GravatarTabs(
     onGravatarUrlChanged: (String) -> Unit,
     showSnackBar: (String?, Throwable?) -> Unit,
 ) {
-    var tabIndex by remember { mutableStateOf(0) }
+    var tabIndex by rememberSaveable { mutableIntStateOf(0) }
 
     val tabs = listOf(
         stringResource(R.string.tab_label_avatar),
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -28,6 +28,7 @@ constraintlayout = "2.1.4"
 activity = "1.9.0"
 androidXLifecycle = "2.8.3"
 navigationCompose = "2.8.0-beta05"
+turbine = "1.1.0"
 
 [libraries]
 desugarJdk = { group = "com.android.tools", name = "desugar_jdk_libs", version.ref = "desugarJdk" }
@@ -64,6 +65,7 @@ roborazzi-compose = { group = "io.github.takahirom.roborazzi", name = "roborazzi
 roborazzi-junit-rule = { group = "io.github.takahirom.roborazzi", name = "roborazzi-junit-rule", version.ref = "roborazzi" }
 androidx-constraintlayout = { group = "androidx.constraintlayout", name = "constraintlayout", version.ref = "constraintlayout" }
 androidx-activity = { group = "androidx.activity", name = "activity", version.ref = "activity" }
+turbine = { group = "app.cash.turbine", name = "turbine", version.ref = "turbine" }
 
 
 [plugins]
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -26,6 +26,8 @@ composeBom = "2024.06.00"
 browser = "1.8.0"
 constraintlayout = "2.1.4"
 activity = "1.9.0"
+androidXLifecycle = "2.8.3"
+navigationCompose = "2.8.0-beta05"
 
 [libraries]
 desugarJdk = { group = "com.android.tools", name = "desugar_jdk_libs", version.ref = "desugarJdk" }
@@ -35,6 +37,8 @@ android-material = { group = "com.google.android.material", name = "material", v
 androidx-core-ktx = { group = "androidx.core", name = "core-ktx", version.ref = "ktx" }
 androidx-appcompat = { group = "androidx.appcompat", name = "appcompat", version.ref = "appcompat" }
 androidx-activity-compose = { group = "androidx.activity", name = "activity-compose", version.ref = "activityCompose" }
+androidx-navigation-compose = { group = "androidx.navigation", name = "navigation-compose", version.ref = "navigationCompose" }
+androidx-viewmodel-compose = { group = "androidx.lifecycle", name = "lifecycle-viewmodel-compose", version.ref = "androidXLifecycle" }
 androidx-compose-ui-tooling = { group = "androidx.compose.ui", name = "ui-tooling", version.ref = "composeUiTest" }
 androidx-compose-junit = { group = "androidx.compose.ui", name = "ui-test-junit4", version.ref = "composeUiTest" }
 androidx-compose-manifest = { group = "androidx.compose.ui", name = "ui-test-manifest", version.ref = "composeUiTest" }
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/service/WordPressOAuthApi.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/service/WordPressOAuthApi.kt
@@ -10,10 +10,10 @@ internal interface WordPressOAuthApi {
     @POST("/oauth2/token")
     @FormUrlEncoded
     suspend fun getToken(
-        @Field("client_id") clientId: String?,
-        @Field("redirect_uri") redirectUri: String?,
-        @Field("client_secret") clientSecret: String?,
-        @Field("code") user: String?,
-        @Field("grant_type") grantType: String?,
+        @Field("client_id") clientId: String,
+        @Field("redirect_uri") redirectUri: String,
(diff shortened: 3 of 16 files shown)


demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -139,7 +141,7 @@ private fun GravatarTabs(
     onGravatarUrlChanged: (String) -> Unit,
     showSnackBar: (String?, Throwable?) -> Unit,
 ) {
-    var tabIndex by remember { mutableStateOf(0) }
+    var tabIndex by rememberSaveable { mutableIntStateOf(0) }
```
AdamGrzybkowski, line 144: The tab index would be lost in the demo app so it wasn't possible to test screen orientation changes with the Quick Editor opened. That fixes it.

gradle/libs.versions.toml
```diff
@@ -26,6 +26,8 @@ composeBom = "2024.06.00"
 browser = "1.8.0"
 constraintlayout = "2.1.4"
 activity = "1.9.0"
+androidXLifecycle = "2.8.3"
+navigationCompose = "2.8.0-beta05"
```
AdamGrzybkowski, line 30: There's a [bug in the library](https://issuetracker.google.com/issues/297258205) in older versions where it's impossible to disable the transition animations. You can test it by changing the library to 2.7.7 and opening the QE. There will be a weird scale animation shown and a slightly different one when opening the QE from an Acitvity 🤷 

I can either use an old 2.7.0 or the latest beta.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
```diff
@@ -0,0 +1,69 @@
...
+    val uiState: StateFlow<OAuthUiState> = _uiState.asStateFlow()
+
+    private val _actions = Channel<OAuthAction>(Channel.BUFFERED)
+    val actions = _actions.receiveAsFlow()
+
+    init {
+        viewModelScope.launch {
+            _actions.send(OAuthAction.StartOAuth)
+        }
+    }
```
AdamGrzybkowski, line 30: This is possible thanks to `navigation-compose`. Otherwise, we would always get the same `ViewModel` instance, so doing anything in the `init` block wouldn't make sense (the same instance on each QE launch)

gradle/libs.versions.toml
hamorillo: The latest beta is fine for now. Perhaps when we release it, we'll have the stable version.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -1,43 +1,69 @@
...
     onAuthError: () -> Unit = {},
 ) {
-    val isAuthorized = rememberSaveable { mutableStateOf(false) }
+    val navController = rememberNavController()
 
-    Surface {
-        if (!isAuthorized.value) {
+    NavHost(
+        navController,
+        startDestination = "entry",
```
hamorillo: > There needs to be a startDestination defined so we need something and I don't want to start with the oauth route and close it quickly if we detect that the token is present.

Why not keep the `isAuthorized` here and use it to define the `starDestination`? Something like:

```suggestion
    val isAuthorized by rememberSaveable { mutableStateOf(false) }
    NavHost(
        navController,
        startDestination = if (isAuthorized) "quickeditor" else "oauth",
```

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
```diff
@@ -0,0 +1,69 @@
...
+import kotlinx.coroutines.launch
+
+internal class OAuthViewModel(
+    private val wordPressOAuthService: WordPressOAuthService,
+) : ViewModel() {
+    private val _uiState = MutableStateFlow(OAuthUiState())
+    val uiState: StateFlow<OAuthUiState> = _uiState.asStateFlow()
+
+    private val _actions = Channel<OAuthAction>(Channel.BUFFERED)
+    val actions = _actions.receiveAsFlow()
```
hamorillo, line 24: > For the ViewModel itself, there's nothing extraordinary. One thing different than what we can be used to is that the Actions emitted from ViewModel are separate and not part of the UiState.

Why do you decide to have those `actions` outside the `UiState`? I would like to understand more about the reasons for that. I was reading a bit about it, and I'm not sure it's a good practice.

I know it's not the exact same case, but in [this post](https://medium.com/androiddevelopers/viewmodel-one-off-event-antipatterns-16a1da869b95) from Google, they describe some antipatterns that can apply here. 

I'm not saying this is wrong, but I would like to understand more why it's better to have it here.

hamorillo (commented):
Awesome work @AdamGrzybkowski !

Using `navigation-compose` sounds like a good solution!

gradle/libs.versions.toml
AdamGrzybkowski: 🤞

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
AdamGrzybkowski: This will be an IO operation later so not sure we want to block the thread here 🤔

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
AdamGrzybkowski: Here's a counterpoint for the article you mentioned - https://proandroiddev.com/viewmodel-events-as-state-are-an-antipattern-35ff4fbc6fb6.

When you do as Google suggests you end up with events/actions in the state that need to be [cleared once handled](https://developer.android.com/topic/architecture/ui-layer/events#consuming-trigger-updates). This feels like unnecessary overhead and complication. Not to mention that the State will grow quite a bit with more complicated navigation.

So my preference is to keep them separate.



BTW. I've decided to use `Action` name, to keep the `Event` name for UI events.

hamorillo: Thanks for the article. Great discussion there! That makes everything more clear to me. 

> BTW. I've decided to use Action name, to keep the Event name for UI events.

👍

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModelTest.kt
```diff
@@ -0,0 +1,97 @@
...
+class OAuthViewModelTest {
+    private val wordPressOAuthService = mockk<WordPressOAuthService>()
+
+    private lateinit var dispatcher: TestDispatcher
+    private lateinit var viewModel: OAuthViewModel
+
+    @Before
+    fun setup() {
+        dispatcher = StandardTestDispatcher()
+        Dispatchers.setMain(dispatcher)
```
hamorillo: 💡Should we create a rule like the suggested [here](https://developer.android.com/kotlin/coroutines/test#setting-main-dispatcher)? Or at least invoke [resetMain](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/reset-main.html) in a `@After` method?

AdamGrzybkowski: Good call, I will add a rule for it!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/service/WordPressOAuthApi.kt
```diff
@@ -0,0 +1,19 @@
...
+
+internal interface WordPressOAuthApi {
+    @POST("/oauth2/token")
+    @FormUrlEncoded
+    suspend fun getToken(
+        @Field("client_id") clientId: String?,
+        @Field("redirect_uri") redirectUri: String?,
+        @Field("client_secret") clientSecret: String?,
+        @Field("code") user: String?,
+        @Field("grant_type") grantType: String?,
```
hamorillo: ⛏ 
```suggestion
        @Field("client_id") clientId: String,
        @Field("redirect_uri") redirectUri: String,
        @Field("client_secret") clientSecret: String,
        @Field("code") user: String,
        @Field("grant_type") grantType: String,
```

hamorillo (approved):
Awesome work! 🚀

PR #241, opened 2024-07-24
Add extensions to start the GQE from Activity/Fragment

AdamGrzybkowski:
### Description

Extra functions that help start the Quick Editor from either Fragment or Activity.

This was briefly discussed [here](https://github.com/Automattic/Gravatar-SDK-Android/pull/239#discussion_r1687831585). 
@hamorillo correctly pointed out that this adds an extra dependency that not everyone might use.

My thoughts on that are the `appcompat` library is likely already included in any Android project and we should make sure our SDK works by default with the old system as it's likely still used in the majority of the apps.

### Testing Steps

To make this easier you can comment the `<intent-filter>` for the `.MainActivity` in the demoApp manifest. 

1. Tap the `Test with Activity without Compose`
2. Tap `Update Avatar`
3. Continue with the required steps to finish the OAuthFlow
4. `Insert the real avatar picker page` text should appear - this is a placeholder for the avatar picker UI
5. Tap the above text
6. Toast message should appear with a confirmation
7. Tap the `UpdateAvatar` 
8. Close the browser with the `x` button
9. Tap the `Done` button
10. Toast with `Dissmised` info is shown
11. Tap the `UpdateAvatar` 
12. Tap `Deny`
13. Toast with `Error` info is shown

### DIFF
--- a/demo-app/src/main/AndroidManifest.xml
+++ b/demo-app/src/main/AndroidManifest.xml
@@ -36,6 +36,23 @@
                     android:scheme="wp-oauth-test" />
             </intent-filter>
         </activity>
+        <activity
+            android:name=".ui.activity.QuickEditorTestActivity"
+            android:launchMode="singleTask"
+            android:label="Old good Activity"
+            android:theme="@style/Theme.Gravatar"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data
+                    android:host="authorization-callback"
+                    android:scheme="wp-oauth-test" />
+            </intent-filter>
+        </activity>
     </application>
 
 </manifest>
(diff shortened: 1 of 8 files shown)


hamorillo (approved):
I still have some doubts about if this should be in a different module. Anyway, it works great and it's useful for a huge number of apps. I guess we can always split it if issues arise so :shipit:

hamorillo:
**Note:** Toast messages differ slightly from the test steps, but everything looks good.

AdamGrzybkowski:
@mlumeau Thoughts on this one?

mlumeau:
While I would be opposed to the `appcompat` dependency being required in the `core` or `ui` modules, I think the nature of the `quickeditor` module allows it. In production, this dependency will be trimmed by ProGuard if we don't use these functions, right?

AdamGrzybkowski:
>In production, this dependency will be trimmed by ProGuard if we don't use these functions, right?

[Correct](https://developer.android.com/build/shrink-code#:~:text=from%20your%20app%20and%20its%20library%20dependencies), assuming code shrinking is enabled.

PR #244, opened 2024-07-29
QuickEditor: Update OpenApi specification and regenerate code

### LINKED ISSUE #243: QuickEditor: Integrate new endpoints

hamorillo:
Closes #243 

### Description

Updating OpenApi specification with the proposed definition. **We likely need to update this with the final specifications.** For the moment, this allows us to generate the models and integrate the new endpoint to continue with the implementation.

In this PR we are not changing anything, just regenerating models and API files so everything should work as before in tests and DemoApp.

### Testing Steps

- Code review and CI


config/detekt/detekt.yml
```diff
@@ -25,6 +25,9 @@ coroutines:
 naming:
   FunctionNaming:
     ignoreAnnotated: ['Composable']
+  FunctionParameterNaming:
```
hamorillo, line 28: We need this because of the:

```
@Part `data`: MultipartBody.Part,
```

The generator handles it as a [reserved word](https://openapi-generator.tech/docs/generators/kotlin/#reserved-words) even though it will work as a param name.

AdamGrzybkowski (approved):

PR #245, opened 2024-07-29
QuickEditor: Create services for Avatars and Identities

### LINKED ISSUE #243: QuickEditor: Integrate new endpoints

hamorillo:
Closes #243 

### Description

Adding new methods in `AvatarService` to upload images and retrieve information about already uploaded avatars. 
Creating `IdentityService` to retrieve identity info (we need this to know which avatar is currently selected) and to switch between avatars for a given `Identity`.

### Testing Steps

As we don't have an implementation using those services/endpoints, **CI and code review should be enough**. We'll need to double-check in following PRs when we implemented the QuickEditor selector.

### DIFF
--- a/gravatar/src/main/java/com/gravatar/services/AvatarService.kt
+++ b/gravatar/src/main/java/com/gravatar/services/AvatarService.kt
@@ -2,6 +2,7 @@ package com.gravatar.services
 
 import com.gravatar.di.container.GravatarSdkContainer.Companion.instance
 import com.gravatar.logger.Logger
+import com.gravatar.restapi.models.Avatar
 import com.gravatar.types.Email
 import kotlinx.coroutines.withContext
 import okhttp3.MultipartBody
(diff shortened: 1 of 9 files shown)


gravatar/src/main/java/com/gravatar/services/IdentityService.kt
```diff
@@ -0,0 +1,93 @@
...
+import com.gravatar.restapi.models.Identity
+import com.gravatar.restapi.models.SelectAvatar
+import kotlinx.coroutines.withContext
+import okhttp3.OkHttpClient
+import com.gravatar.di.container.GravatarSdkContainer.Companion.instance as GravatarSdkDI
+
+/**
+ * Service for managing Gravatar identities.
+ */
+public class IdentityService(private val okHttpClient: OkHttpClient? = null, oauthToken: String) {
```
hamorillo: I have doubts about using the name Identity. We don't use that term widely so maybe we should look for another name.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -48,4 +51,76 @@ public class AvatarService(okHttpClient: OkHttpClient? = null) {
...
+    }
+
+    /**
+     * Retrieves a list of available avatars for the authenticated user.
+     *
+     * @param oauthToken The OAuth token to use for authentication
+     * @return The list of avatars
+     */
+    public suspend fun retrieve(oauthToken: String): List<Avatar> = withContext(GravatarSdkDI.dispatcherIO) {
+        val service = instance.getGravatarV3Service(okHttpClient, oauthToken)
```
AdamGrzybkowski, line 100: ❓ Creating these services on each function call seems unnecessary. Can't we use [Dynamic Headers](https://futurestud.io/tutorials/retrofit-add-custom-request-header) in Retrofit?

gravatar/src/main/java/com/gravatar/services/IdentityService.kt
AdamGrzybkowski: As long as we keep the naming consistent with the backend I'm ok with this one.

AdamGrzybkowski: ❓ Why do we pass the auth token in the constructor? It's also a param in each function.

hamorillo: I forgot to remove it. Good catch.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
hamorillo: That was my first idea. The problem was that API classes are generated by the OpenApi generator, and I haven't found a way for the generator to add those as a Dynamic Header. 

The only way I've found so far is to declare that header in the OpenApi specification explicitly. However, I don't like that solution, as we've already declared the security scheme the endpoint should use. It looks like a concrete problem with Kotlin/our generator, which should be solved without changes in the specifications.

I dedicated some time to this yesterday and finally decided to move on with this approach. As it's something internal in the library, we can try to find a way without retro compatibility issues.

AdamGrzybkowski (approved):

PR #246, opened 2024-07-30
OAuth token storage

### LINKED ISSUE #235: QuickEditor: Handle token storage

AdamGrzybkowski:
Closes #235

### Description

In order to keep the user logged in we have to store the token that authenticates with the Gravatar API. 

To do that I've used DataStore preferences. The token is not encrypted for now. I will take a look at what's possible after this PR.

I've added another dependency - [Jetpack App Startup](https://developer.android.com/topic/libraries/app-startup). This one is used to initialize our library with the application context without forcing the user to init the library manually. 

I've also added the `GravatarQuickEditor.logout()` so third-party devs can clean the session for the given email. 

https://github.com/user-attachments/assets/2860b904-ff9d-4d8d-a175-57c81fbaa7cb

### Testing Steps

1. Start the demo app
2. Tap the "Update avatar" on the 3 tab
3. Follow the OAuth process (keep in mind there are two activities handling the redirect URL, make sure you pick the right one)
4. Confirm you see "Insert the real avatar picker here"
5. Tap "Done"
6. Tap the "Update avatar"
7. Confirm you see "Insert the real avatar picker here"
8. Tap "Done"
9. Tap "Logout user"
10. Tap the "Update avatar"
11. Confirm auth flow is started


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
```diff
@@ -30,19 +32,19 @@ internal class OAuthViewModel(
...
-                redirectUri = redirectUri,
+                clientId = oAuthParams.clientId,
+                clientSecret = oAuthParams.clientSecret,
+                redirectUri = oAuthParams.redirectUri,
             )
 
             when (result) {
                 is Result.Success -> {
-                    // todo: Save access token
+                    tokenStorage.storeToken(email, result.value)
```
hamorillo: We can use the `hash` instead of the `email` as key in the `tokenStorage`. I guess I someone obtain the `token` we have a big problem but a least they won't have the relation between `mail<->token`.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * All non-auth params required to launch the Gravatar Quick Editor
+ *
+ * @property appName The appName of the app that is launching the Quick Editor
+ * @property email The email of the user
+ */
+public class GravatarQuickEditorParams private constructor(
+    public val appName: String,
+    public val email: String,
```
hamorillo: We have a Gravatar `Email` class that we can use here. I have mixed feelings about whether we should or not. For consistency, I guess the answer is yes for consistency with other modules, but it is not a strong opinion.

hamorillo (commented):
Great job! I left some minor comments, and I'll try to test the `demoApp` tomorrow.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/initializer/QuickEditorInitializer.kt
```diff
@@ -0,0 +1,15 @@
+package com.gravatar.quickeditor.initializer
+
+import android.content.Context
+import androidx.startup.Initializer
+import com.gravatar.quickeditor.QuickEditorContainer
+
+internal class QuickEditorInitializer : Initializer<QuickEditorContainer> {
```
hamorillo: ⛏ I think it looks better with:

```suggestion
internal class QuickEditorContainerInitializer : Initializer<QuickEditorContainer> {
```

hamorillo (commented):
DemoApp works like a charm.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
AdamGrzybkowski: hmm we will need that email value (to display it) and this object doesn't expose it (only the hash) by design, right?

AdamGrzybkowski: oh it does, via the `toString()` function. I guess that works 👍

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
AdamGrzybkowski: Good idea! Done

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/initializer/QuickEditorInitializer.kt
AdamGrzybkowski: Agree, fixed!

hamorillo (approved):
We need to update the `.api` file but apart from that looks and works great! 🚀

PR #250, opened 2024-08-01
Use encrypted Datastore for token storage

### LINKED ISSUE #248: Token storage encryption
Ideally, we should secure the token by encrypting it before storing it. 

We can use the [Encrypted DataStore](https://github.com/osipxd/encrypted-datastore) library for that, though it requires bumping the minSdk to 23. 

API 23 covers 98.5% of Android usage based on [this](https://apilevels.com/) so I think it's safe to update.

AdamGrzybkowski:
Closes #248

### Description

This PR adds token storage encryption via this [Encrypted Datastore](https://github.com/osipxd/encrypted-datastore) library.

This library uses `security-crypto` under the hood, requiring us to bump the `minSdk` to 23. I think that's completely acceptable at this point. API 23 covers 98.5% of Android usage based on [this](https://apilevels.com/). Also, API 21 is 10 years old... 
I did update the `minSdk` only for the `:quickeditor` module. Would it be better to keep it the same on all the modules for consistency?

There are two things I'm not super happy about: 
- `security-crypto` [was silently deprecated by Google](https://github.com/osipxd/encrypted-datastore/issues/31). It's unclear what's next at this point as there's no communication. 
- Having encrypted storage means third-party devs have to add exclusion rules so that the files storing the token are not included in the backup. You can see those rules in the demo app config. Additionally, I've added a fallback mechanism to clear the files if the fatal exception is thrown in case the rules are not present. 

So far I'm opening this as a draft - I would like to get your thoughts on this one.

### Testing Steps
To test the backup scenario I've used [these steps](https://arophix.com/2019/09/12/a-step-by-step-tutorial-to-test-android-backup-and-restore/) - but before running the restore command I switched the devices to restore it in a different one (using the same one doesn't reproduce the crash).

First, let's make sure this still works as it did before the encryption. Follow these steps:
1. Start the demo app
2. Tap the "Update avatar" on the 3 tab
3. Follow the OAuth process (keep in mind there are two activities handling the redirect URL, make sure you pick the right one)
4. Confirm you see "Insert the real avatar picker here"
5. Tap "Done"
6. Tap the "Update avatar"
7. Confirm you see "Insert the real avatar picker here"
8. Tap "Done"
9. Tap "Logout user"
10. Tap the "Update avatar"
11. Confirm auth flow is started

Now, let's test the backup scenario. As mentioned above I used these [steps](https://arophix.com/2019/09/12/a-step-by-step-tutorial-to-test-android-backup-and-restore/). You will need two devices (or emulators) A and B. 

1. Configure both emulators using steps 1-3 from the above tutorial
2. Install the DemoApp on both devices. Clean the DemoApp storage on device B (to make sure it has no files) and close the device B
3. Open the app on device A, and make sure to follow the OAuth flow to have the token saved. You can verify that you have it stored by opening the `Device Explorer` and going to `data/data/com.gravatar.demoapp/files/datastore`
4. Run `adb shell bmgr list transports`
5. Run `adb shell bmgr backupnow com.gravatar.demoapp`
6. Switch to device B
7. Run `adb shell dumpsys backup | grep Current` to get the backup token
8. Run `adb shell bmgr restore {token_from_step_7)  com.gravatar.demoapp`
9. Verify that datastore and share prefs files were not restored on device B in `Device Explorer`


demo-app/src/main/res/xml/backup_rules.xml
```diff
@@ -1,13 +1,9 @@
...
-    <!--
-   <include domain="sharedpref" path="."/>
-   <exclude domain="sharedpref" path="device.xml"/>
--->
+    <exclude
+        domain="sharedpref"
+        path="__androidx_security_crypto_encrypted_file_pref__.xml" />
+    <exclude
+        domain="file"
+        path="datastore/quick-editor-preferences.preferences_pb" />
```
AdamGrzybkowski, line 8: This is what we would have to add to the documentation to make sure third-party devs include those rules in their config. 

We can provide this in the library, but there's no guarantee they will be used. If the third-party app has its own rules, ours will be discarded. 

With the extra code to clean up the backup files, we could technically avoid this. But I would rather now have those file at all to not do the cleanup at all. The `__androidx_security_crypto_encrypted_file_pref__` file is not used only by us.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/datastore/SecureDataStoreUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.security.crypto.EncryptedFile
+import androidx.security.crypto.MasterKeys
+import java.io.File
+
+/**
+ * This creates the encrypted file and fallbacks to a new file when previous one was restored from the
+ * backup on another device.
+ */
+@Suppress("TooGenericExceptionCaught", "SwallowedException")
+internal fun Context.createEncryptedFileWithFallbackReset(name: String): EncryptedFile {
```
AdamGrzybkowski, line 15: Source - https://issuetracker.google.com/issues/230651240

demo-app/src/main/res/xml/backup_rules.xml
mlumeau: I think we should provide these rules as a default configuration within the SDK if we can, so that it works out of the box for most use cases.

AdamGrzybkowski: >I think we should provide these rules as a default configuration within the SDK if we can, so that it works out of the box for most use cases.

Android Studio adds empty rules by default to a new project. We have to assume that the configuration we provided won't work as a default case.

maxme:
Why can't we use [`EncryptedSharedPreferences`](https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences) directly?

AdamGrzybkowski:
>Why can't we use [EncryptedSharedPreferences](https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences) directly?


Mainly because it's now recommended to use `DataStore`. You can see the comparison [here](https://android-developers.googleblog.com/2020/09/prefer-storing-data-with-jetpack.html).

maxme (commented):
Thinking out loud, and I don't know if this was discussed elsewhere. 

If we have other modules that needs the auth token, we'll likely need to duplicate the feature or move this under another module. 

Currently `quickeditor` is a module that's on the bottom of the dependency tree, so I wonder we should move `SecureDataStoreUtils.kt` under `gravatar`. I'd also move `TokenStorage` there as this would be the only way to get/store the auth token from any component that needs it.

AdamGrzybkowski:
>If we have other modules that needs the auth token, we'll likely need to duplicate the feature or move this under another module.

The token storage is just an internal implementation detail of the QuickEditor. If we move it to the `:gravatar` module we would have to make it public and open it for anyone which is not ideal and not something I wanted. 

Also, because this is internal and has such a visibility modifier we can always move it and put it in another module. We won't be able to do it the other way around.

AdamGrzybkowski:
>If we have other modules that needs the auth token

Any modules you have in mind?

maxme (approved):
> Any modules you have in mind?

Nope, but that could come from anything supported in the API but not supported in the quick editor. And you're right we can open it up or move it later if it's necessary.

:shipit:

PR #255, opened 2024-08-06
QuickEditor: Profile Card in the Quick Editor

### LINKED ISSUE #254: QuickEditor: Show profile card at the top of the picker

AdamGrzybkowski:
Closes #254 

### Description

This PR adds the Profile Card component at the top of the Quick Editor. 

The `ProfileSummary` component from the `:gravatar-ui` module was used and put inside a Card. 
For the dark theme we can't have the shadow so the background is uses a variation of the `surface` color to be different from the background. 

| Light | Dark |
|---|---|
| Screenshot 2024-08-06 at 12 47 42 | Screenshot 2024-08-06 at 12 47 50 |

Note: Shadows do not work with screenshot testing so the recorded images look slightly different in a light theme. 

### Testing Steps 
1. Run the DemoApp 
2. Go to Avatar Update tab
3. Tap `Update Avatar` button
4. Follow the steps to log in 
5. Confirm you can see the profile component at the top
6. Run `adb shell cmd uimode night yes` to enable dark mode
7. Confirm the UI looks fine


mlumeau (approved):
Looks good to me! Code looks clean and tests are 👌.
One ⛏️  I got while testing is that we don't check that the email that is passed to the `AvatarPicker` belongs to the user that is logged in, therefore we can display a different user in the profile card. ~~This is not something I think we should fix right now, though.~~ EDIT: It looks like #259 fixes this by failing when selecting the avatar associated to the email

AdamGrzybkowski:
>One ⛏️ I got while testing is that we don't check that the email that is passed to the AvatarPicker belongs to the user that is logged in, therefore we can display a different user in the profile card. 

Hmm I will test it but I think this shouldn't happen. We take the token for the given email, so if the email is new we should start with the OAuth flow 🤔

PR #260, opened 2024-08-08
QuickEditor: Avatar selection

AdamGrzybkowski:
Closes #224 

### Description

This PR adds the option to select one of the uploaded avatars. 

As per the design, when the avatar is clicked:
- an overlay is shown with the circular progress 
- avatar is instantly selected (reverts back if request fails)

We can't use `Scaffold` on the bottom sheet so I did add `SnackbarHost` manually. 

There's one known issue now https://github.com/Automattic/Gravatar-SDK-Android/issues/258
The avatar in the `ProfileCard` is not updated instantly - this is partially because there's no trigger for it yet in the ViewModel. Still, even if we refresh the profile the avatar won't update because of the cache.
This must be solved by loading the selected image manually or using the cache buster. 

I did update the DemoApp to update the avatar as well.

https://github.com/user-attachments/assets/335c59cb-e602-4c46-8a7b-5692e42000dc

### Testing Steps
**Important Note**: The backend is currently returning the wrong format for the `updated date` field. As a workaround, you can apply this patch to comment that field from the Avatar object.
[Gravatar-Android-13-04-48.patch](https://github.com/user-attachments/files/16455284/Gravatar-Android-13-04-48.patch)

1. Make sure you have at least two avatars uploaded on Gravatar
2. Run the DemoApp 
3. Go to Avatar Update tab
4. Tap `Update Avatar` button
5. Follow the steps to log in 
8. Select a different avatar
9. Confirm you can see the progress indicator
10. Confirm the new avatar was highlighted 
11. Confirm the snack was shown

PR #261, opened 2024-08-09
QuickEditor: Update the avatar in the ProfileCard after selection

### LINKED ISSUE #258: QuickEditor: Always show up-to-date avatar on the Profile Card
We have the same problem we experience in Pocket Casts where the profile image is not updated instantly because of the cache. 

One solution is to use the cache buster to ensure we always show the updated avatar in the profile card.

This is what we want to avoid (notice the selected image vs the one on the profile card):
Screenshot 2024-08-07 at 12 09 15

AdamGrzybkowski:
Closes #258

### Description

As mentioned in the previous [PR](https://github.com/Automattic/Gravatar-SDK-Android/pull/260) the avatar in the ProfileCard is not updated instantly. That is because by default it loads the avatar based on the email hash and this doesn't change when a new avatar is selected. If the URL is static, it means cache is used so the Avatar won't update for some time. 

To overcome this, I've added a new `Avatar` component that uses the unique avatar URL. We get that URL for each avatar so when we select one we can update the Profile with such a URL.

https://github.com/user-attachments/assets/44412c06-4c8a-421b-80a7-11c801a284d5

### Testing Steps

2. Run the DemoApp 
3. Go to Avatar Update tab
4. Tap `Update Avatar` button
8. Select a different avatar
9. Confirm it was updated in the ProfileCard at the top of the bottom sheet

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
@@ -267,5 +267,5 @@ private fun AvatarSectionPreview() {
     }
 }
 
-private val Avatar.fullUrl: String
+internal val Avatar.fullUrl: String
     get() = "https://www.gravatar.com$imageUrl?size=200"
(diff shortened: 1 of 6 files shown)


mlumeau (approved):
Works as expected, code changes LGTM. ~~Is there any way to add unit tests this feature?~~ Nevermind, just saw the change in `AvatarPickerViewModelTest.kt`.

PR #265, opened 2024-08-13
QuickEditor: Upload image from the gallery

### LINKED ISSUE #226: QuickEditor: Option to pick a photo from media library to upload a new avatar
Add a button to pick an image from the device.

AdamGrzybkowski:
Closes #226

### Description

This PR adds an option to pick an image from the phone gallery and upload it to the server after cropping. 

For the image picking I went with the latest recommended solution from Google - [Photo Picker](https://developer.android.com/training/data-storage/shared/photopicker).
For the cropping, I used the UCrop library as we did already for the WordPress picker. 

I've mimicked the web by disabling the `Upload image` button while the avatar is being uploaded to the web.

https://github.com/user-attachments/assets/9b5df0cd-2504-4929-a79c-e46da2e081c7

### Testing Steps

2. Run the DemoApp 
3. Go to Avatar Update tab
4. Tap `Update Avatar` button
8. Tap the `Upload Image` button
9. Tap the `Choose a Photo`
10. Pick a photo from the gallery
11. Crop if you want and confirm
12. Confirm you see the photo in the list with the progress indicator
13. Confirm the photo was uploaded and can be selected

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
@@ -11,10 +11,6 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.wrapContentSize
-import androidx.compose.foundation.lazy.LazyRow
-import androidx.compose.foundation.lazy.items
-import androidx.compose.foundation.shape.RoundedCornerShape
-import androidx.compose.material3.Button
 import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Snackbar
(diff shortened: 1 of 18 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -90,6 +89,10 @@ internal fun AvatarPicker(
                                 duration = SnackbarDuration.Long,
                             )
                         }
+
+                        is AvatarPickerAction.LaunchImageCropper -> {
+                            uCropLauncher.launchAvatarCrop(action.imageUri, action.tempFile, context)
```
mlumeau: Should we introduce some abstraction layer here in order to reduce coupling with a third party library?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -154,75 +163,23 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onAvatarSelected: (Avata
...
-        if (popupVisible) {
-            MediaPickerPopup(
-                alignment = Alignment.BottomCenter,
-                onDismissRequest = { popupVisible = false },
-                offset = IntOffset(0, -popupYOffset - 30),
-                onChoosePhotoClick = { popupVisible = false },
-                onTakePhotoClick = { popupVisible = false },
-            )
-        }
+private fun ActivityResultLauncher<Intent>.launchAvatarCrop(image: Uri, tempFile: File, context: Context) {
```
mlumeau: This could be in its own file along with the `UCROP_COMPRESSION_QUALITY` constant

mlumeau (changes requested):
Just a few nitpicks to try and keep the `AvatarPicker.kt` file as tidy as possible. I like that you refactored `AvatarsSection` into its own file though.

mlumeau (changes requested):
Also I noted while testing that the uploaded image shows at the start (left) of the list. If we're scrolled all the way to the right before uploading a new picture, we won't see it being uploaded. Could we reset the scroll position when we're back from the media picker?

AdamGrzybkowski:
>Also I noted while testing that the uploaded image shows at the start (left) of the list. If we're scrolled all the way to the right before uploading a new picture, we won't see it being uploaded. Could we reset the scroll position when we're back from the media picker?


hmm that should work, I will check that out!

mlumeau:
> > Also I noted while testing that the uploaded image shows at the start (left) of the list. If we're scrolled all the way to the right before uploading a new picture, we won't see it being uploaded. Could we reset the scroll position when we're back from the media picker?
> 
> hmm that should work, I will check that out!

Weird, I tested again and this time it worked! Nevermind then.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
AdamGrzybkowski: Done! Let me know if this is what you meant!

mlumeau (approved):
Looks good now!

PR #266, opened 2024-08-14
QuickEditor: Take photo action in the media picker

### LINKED ISSUE #225: QuickEditor: Option to use camera to upload a new avatar
Add the button to open the camera to capture the new avatar.

AdamGrzybkowski:
Closes #225

### Description

This PR implements the `Take photo` option from the media picker popup.

An important bit here is the `FileProvider` config - this needs to be custom so that it will work when the app implementing our library has its own `FileProvider` setup.  You can read about it [here](https://commonsware.com/blog/2017/06/27/fileprovider-libraries.html). 
I've also added a config to our DemoApp to make sure this works (that's how I discovered this needs to be handled).

https://github.com/user-attachments/assets/7e9e399d-957e-46a6-a64a-cf6d0e70e0d9

### Testing Steps

2. Run the DemoApp 
3. Go to Avatar Update tab
4. Tap `Update Avatar` button
8. Tap the `Upload Image` button
9. Tap the `Take photo`
10. Take photo with camera
11. Crop if you want and confirm
12. Confirm you see the photo in the list with the progress indicator
13. Confirm the photo was uploaded and can be selected


gradle/libs.versions.toml
```diff
@@ -16,7 +16,7 @@ junit = "4.13.2"
 mockk = "1.13.9"
 appcompat = "1.7.0"
 material = "1.12.0"
-ucrop = "2.2.9"
+ucrop = "2.2.9-native"
```
AdamGrzybkowski: I had to switch to the native version because the compression wouldn't work for me and the uploaded avatar would be the same size as the original image despite the compression.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -163,7 +164,8 @@ internal fun AvatarPicker(
     }
 }
 
-private const val UCROP_COMPRESSION_QUALITY = 100
+private const val UCROP_COMPRESSION_QUALITY = 90
+private const val UCROP_MAX_IMAGE_SIZE = 1080
```
AdamGrzybkowski: Set this as the max size for now - seems enough for an Avatar though happy to hear your opinions.

gradle/libs.versions.toml
mlumeau: From uCrop's documentation: 
> implementation 'com.github.yalantis:ucrop:2.2.8-native' - get power of the native code to preserve image quality (+ about 1.5 MB to an apk size)

1.5 MB is a lot. Is there any workaround?

AdamGrzybkowski: I haven't found one in this library. We can handle the compression ourselves on top of it though, but that will be tricky as we have to be careful not to duplicate the compression when the one from the library works (I imagine this might be dependent on the system version for example). 

Using the `-native` variant seemed like a better pick for now to not spend too much time on it.

mlumeau: @maxme I'm curious to know what you think of this

maxme: Adding 1.5Mb is not ideal, but what worries me most is adding native code to our library dependencies. For deployments on the app store, it's going to be extra steps, and might be limiting for our users. They might be surprised they have native code and might look for the culprit dependency.

maxme: > I had to switch to the native version because the compression wouldn't work for me and the uploaded avatar would be the same size as the original image despite the compression.

Do you get some logs or is this a silent error (they skip the compression steps maybe)?

AdamGrzybkowski: I haven't seen any errors (I haven't looked for them though at that point) just that the final image had the exact same size as the photo taken with the camera. I will take another look at it.

AdamGrzybkowski: I'm testing this on different devices/emulators and the image size is always the same. The compression seems to work in a way that you can clearly see the drop in the image quality and the width/height sizes. Yet the file size remains unchanged.

AdamGrzybkowski: I removed the `-native` version for now.

maxme:
Note: we're currently waiting for a reply/merge to fix a bug in uCrop - https://github.com/Yalantis/uCrop/pull/926

AdamGrzybkowski:
>Note: we're currently waiting for a reply/merge to fix a bug in uCrop - https://github.com/Yalantis/uCrop/pull/926

I don't think it makes sense to block this PR because of this. We can fix this later by either updating the UCrop version or switching to our fork.

maxme (approved):

PR #271, opened 2024-08-19
QuickEditor: Error snackbar styling

### LINKED ISSUE #268: QuickEditor: Error snackbar styling

AdamGrzybkowski:
Closes #268

### Description
This PR implements custom snakbard styling for two types: INFO and ERROR.

This is accomplished by creating a `QESnackbarVisuals` and passing it to the custom `QESnackbarHost`.

Additionally, two error snack bars are added: when upload fails and when selecting an avatar fails. 

**Upload failure**
https://github.com/user-attachments/assets/35b4d4c4-41fd-4d8d-8b7f-5ea7d2522c14

**Selection failure**
Screenshot 2024-08-19 at 12 51 39

### Testing Steps

2. Run the DemoApp 
3. Go to Avatar Update tab
4. Tap `Update Avatar` button
6. Upload an avatar (make sure to fake error response with a tool like [HTTP toolkit](https://httptoolkit.com/) or disable the network)
7. Confirm an error snackbar was shown
8. Tap retry (make sure this time the request is not mocked)
9. Confirm the image was uploaded
10. Tap the new avatar to select it (fake the response here so that it fails)
11. Confirm an error snackbar was shown

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/extensions/QESnackbar.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/extensions/QESnackbar.kt
@@ -1,0 +1,81 @@
+package com.gravatar.quickeditor.ui.extensions
+
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Snackbar
+import androidx.compose.material3.SnackbarDuration
+import androidx.compose.material3.SnackbarHost
+import androidx.compose.material3.SnackbarHostState
+import androidx.compose.material3.SnackbarResult
+import androidx.compose.material3.SnackbarVisuals
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+
+internal data class QESnackbarVisuals(
+    val snackbarType: SnackbarType,
+    override val actionLabel: String?,
+    override val duration: SnackbarDuration,
+    override val message: String,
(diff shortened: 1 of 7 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/extensions/QESnackbar.kt
```diff
@@ -0,0 +1,81 @@
+package com.gravatar.quickeditor.ui.extensions
```
maxme, line 1: Nitpicking here and I don't really mind it, so feel free to ship it like that ;)

I think we haven't prefixed component names in other places and IMO we shouldn't do it (so I'd drop `QE` here).

maxme (approved):
Code looks good, snackbar error works as expected, retry button also works fine, LGTM, :shipit: 

light|dark
-|-
Screenshot_20240826_093813|Screenshot_20240826_095116

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/extensions/QESnackbar.kt
AdamGrzybkowski: I started prefixing those that have clashing names. I'm not the biggest fan when you get multiple options with the same name in the autocomplete popup. 

This one is just for the file name so I can drop it - but I did it for some Composables in other PRs.

mlumeau: I don't think it's bad to prefix it especially since they're `internal` so it's just for our convenience.

PR #272, opened 2024-08-20
QuickEditor: Small UI updates

AdamGrzybkowski:
### Description 

This PR adds small changes to the UI that were introduced in the final designs.

1. `DragHandle` - I had to make a custom one because the default one has hardcoded vertical padding significantly bigger than what we want.
2. Gravatar Icon in the top bar of the bottom sheet. Additionally, I've extracted the whole bottom bar to a separate Composable.
3. Add `.scrollable` modifier to the AvatarPicker to make the content scrollable. 
4. Remove the `fillMaxWidth` modifier from the ModalBottomSheet to not expand to full width on bigger screens.

### Testing Steps
1 and 2 can be tested by just opening the QuickEditor. 
To test 3. you have to rotate the device/emulator to horizontal mode.
To test 4. you have to run the demo app on a tablet or a foldable.


maxme (approved):
LGTM :shipit:

PR #274, opened 2024-08-20
[Tooling] Automation for downloading translations

iangmaia:
### Description

This PR adds basic Fastlane automation for downloading translation strings from GlotPress.

Additionally, it also sets up Danger in the project, mainly to do Ruby-related checks for now, but of course the checks can be expanded as needed.

### Testing Steps

1. Checkout this branch
1. Update one of the translation strings on GlotPress ([ui module](https://translate.wordpress.com/projects/gravatar/gravatar-android-sdk/gravatar-ui/), [quick editor module](https://translate.wordpress.com/projects/gravatar/gravatar-android-sdk/gravatar-quick-editor/))
1. Run `bundle install` + `bundle exec fastlane download_translations`
1. If there are updates, the tool should download the updated strings of the corresponding locale, update and commit them in the project


AdamGrzybkowski:
Hi @iangmaia 

When I run `bundle exec download_translations` this is what I get:
```
bundler: command not found: download_translations
Install missing gem executables with `bundle install`
```

I did run `bundle install`. Would you happen to have any ideas on what's missing?

AliSoftware:
> Hi @iangmaia
> 
> When I run `bundle exec download_translations` this is what I get:
> 
> ```
> bundler: command not found: download_translations
> Install missing gem executables with `bundle install`
> ```
> 
> I did run `bundle install`. Would you happen to have any ideas on what's missing?

@AdamGrzybkowski It's a typo in the testing instructions, this should be `bundle exec fastlane download_translations`. I've edited the PR description to fix it.

AdamGrzybkowski:
Thanks @AliSoftware! I've just found out this myself and was about to edit the comment.

AdamGrzybkowski (commented):
Look good, thanks @iangmaia!

At first, I thought the command `bundle exec fastlane download_translations` didn't work for me, because I was expecting untracked changes to the strings file, but it turned out the updates are automatically committed. 

I have some questions about the process:
1. How do we upload new strings for translations? Is this a manual process? 
2. When we want to update an English copy, should we create a new string resource rather than updating the current one?
3. What happens when a string is removed? Will it be automatically removed from other languages?

AliSoftware:
Those are great questions!

> 1. How do we upload new strings for translations? Is this a manual process?

New strings are typically imported into GlotPress by a wpcom cron job polling the repo for the `strings.xml` file every day or so — the diff to add gravatar-sdk-android to this job can be found in D158935-code.

That being said, that raises a question about your team's plan for the release process of the SDK:
 - Do you plan to have a code-freeze and beta-testing phase for releases of this SDK, similar to what we do for apps?
    - If so we should make the code-freeze also freeze the strings file by making a copy of it in some other location during code-freeze then have the wpcom cron job only import that copy, so that changes in the original `strings.xml` file from `main` during the development phase would not affect strings in GlotPress being translated for the currently-code-frozen code and branch)
    - That way new strings would only appear in GlotPress after each code-freeze, and the translators would have the duration of the beta-testing phase to translate those new strings, before you finally run `download_translations` just before shipping the release at the end of the beta-testing phase.
 - Or do you plan to just create a new tag directly from `main` whenever you decide to publish a new release—be it a beta releases with versions like `1.0-beta-2` or a final ones like `1.0`)—without having a code-freeze and week-long beta-testing phase process and `release/*` branch beforehand?
   - If so, that means the original `strings.xml` file that you would update as part of landing new features to `main` should be the one imported continuously by the wpcom cron job (like is currently implemented in D158935-code)
   - That way new strings would be imported to GlotPress daily, with the translators translating them as they come.
   - And when you decide to do a new tag to do a new release, and would run `download_translations` beforehand, you might be missing translations of strings that were just introduced in `main` too recently for them to have been translated by translators in GlotPress in time

It's up to the team really what they plan for the release process tbh.
 - For mobile app projects at a8c we usually use the code-freeze + beta-testing process
 - While for homegrown library projects used by those apps (e.g. WordPress-Authenticator, etc), we don't really use a code-freeze and just tag them when we need a new version; but also the way those libraries are translated is a bit special-case, and besides we don't have dedicated teams working on those libraries (but rather devs working on client apps do the modification they need on them as they implement new features in their app needing lib changes)
 - But Gravatar SDK is in a unique position which is a bit in-between those two, as it's a product on its own (not a library we only use internally for our products, but a proper SDK) with a dedicated team working on it. So I can see both release processes making sense, depending on how/if you want to have a code-freeze + beta-testing phase of the SDK before each release or not, etc.

> 2. When we want to update an English copy, should we create a new string resource rather than updating the current one?

This will mostly depend on the above. But in general I'd still highly recommend to create a new string resource rather than updating the current one in those cases, especially for cases where the copy update is not just a typo fix in the English copy (missing punctuation, typo in a word…) but a bigger change (especially one that might involve a change in the placeholders used in the string, or the meaning conveyed by the string).

This is because, otherwise if you're still using the same key when updating the English copy, there's a risk that the updated copy would not be translated in time, and we'd end up still using the translation of the old copy for some locales. 
 - This could lead to crashes for cases like if the old copy (and thus its old translation) used `%1$s` but the new copy now uses `%1%d` for example: if we end up still having translations for the old copy because they haven't been updated with the new copy in time, the app will work in English but crash in those other locales.
 - Even if the updated copy didn't change the placeholders, if its _meaning_ is changed, that would cause discrepencies in the UI (e.g. old English copy being "This Gravatar feature is free!" and you later update the copy to "This Gravatar feature is available to paid users" without changing the key, old translations still mentioning the feature being free would be misleading and inconsistent to users of those locales.

> 3. What happens when a string is removed? Will it be automatically removed from other languages?

If I'm not mistaken, when the `strings.xml` file is imported in GlotPress by the wpcom cron job, any key that was previously imported in GlotPress but that is no longer present in the latest `strings.xml` file will be marked as deprecated/deleted in GlotPress and be excluded from translations and exports (i.e. it's similar to GlotPress deleting the entry in its db, except they only "move it to Trash" without really deleting it… mostly so that Translation Memory and history is preserved I think)

iangmaia:
Thanks for the detailed explainer @AliSoftware -- I was writing an answer myself but yours is a lot more detailed 😄 

My suggested setup was based on always sync'ing what's on `trunk`, but good points raised on what process the team will follow for releases 👍 

> > 3. What happens when a string is removed? Will it be automatically removed from other languages?
> 
> If I'm not mistaken, when the `strings.xml` file is imported in GlotPress by the wpcom cron job, any key that was previously imported in GlotPress but that is no longer present in the latest `strings.xml` file will be marked as deprecated/deleted in GlotPress and be excluded from translations and exports 

This is exactly what I saw when doing some tests with GlotPress when I was setting up the project 👍

Dangerfile
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+github.dismiss_out_of_range_messages
+
+# `files: []` forces rubocop to scan all files, not just the ones modified in the PR
+# `skip_bundle_exec` prevents RuboCop from running using `bundle exec`, which we don't want on the linter agent
+rubocop.lint(files: [], force_exclusion: true, inline_comment: true, fail_on_inline_comment: true, include_cop_names: true, skip_bundle_exec: true)
```
dangermattic, line 7: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/Gravatar-SDK-Android/blob/2805492494021bbdb70339073544abc0e00048f2/Dangerfile#L7"></span>Layout/LineLength: Line is too long. [147/120]</td>
    </tr>
  </tbody>
</table>

AdamGrzybkowski:
Thanks for all these answers! 

>That being said, that raises a question about your team's plan for the release process of the SDK:

So far we've been tagging new versions without doing code freezes and beta testing. I'm not sure this would bring that much benefit to our development when we can simply wait for translations before creating a new tag. 

Speaking of that - is there any guarantee the translation will be ready in let's say a week? Is there any guarantee the translations will be done at all?

AdamGrzybkowski:
>My suggested setup was based on always sync'ing what's on trunk, but good points raised on what process the team will follow for releases 👍


This sounds good to me, at least at this point! Thanks!

AdamGrzybkowski (approved):
LGTM 👍

fastlane/Fastfile
```diff
@@ -31,17 +31,32 @@ SUPPORTED_LOCALES = [
...
+    next if skip_commit
+
+    strings_paths = [RES_UI_DIR_PATH, RES_QUICK_EDITOR_DIR_PATH]
+    git_add(path: strings_paths)
+    git_commit(
+      path: strings_paths,
+      message: 'Update translations',
+      allow_nothing_to_commit: true
     )
   end
```
AliSoftware: Alternate implementation idea (just being fancy here—though this could make the code easier to update if the projects ends up adding more modules—and thus more `strings.xml` files to translate—in the future?):

1. Replace the 4 constants at the top of the Fastfile (L7–L10) with
```ruby
GLOTPRESS_PROJECT_BASE_URL = 'https://translate.wordpress.com/projects/gravatar/gravatar-android-sdk'
RESOURCES_TO_TRANSLATE = {
  File.join('gravatar-ui', 'src', 'main', 'res') => "#{GLOTPRESS_PROJECT_BASE_URL}/gravatar-ui/",
  File.join('gravatar-quickeditor', 'src', 'main', 'res') => "#{GLOTPRESS_PROJECT_BASE_URL}/gravatar-quick-editor/"
}
```
2. Then use that to implement the rest of that lane in a loop:
```ruby
  RESOURCES_TO_TRANSLATE.each do |res_dir, gp_url|
    android_download_translations(
      res_dir: res_dir,
      glotpress_url: gp_url,
      locales: SUPPORTED_LOCALES,
      skip_commit: true
    )
  end

  next if skip_commit

  strings_paths = RESOURCES_TO_TRANSLATE.keys
  git_add(path: strings_paths)
  git_commit(
    path: strings_paths,
    message: 'Update translations',
    allow_nothing_to_commit: true
  )
```

Again, this is just me being fancy with Ruby; optional and not a blocker of course 😉

iangmaia: Good idea, I'll take it -- perhaps Gravatar will remain as two modules but I think it makes this code more reusable / generic and doesn't complicate it a lot more. Updated on b26a37e1704f10b05a85fbf7bf334813a0e09e15

iangmaia:
@AdamGrzybkowski 

> but it turned out the updates are automatically committed.

I've pushed a9e4aaba6446670056d87a4bbb9441f181ce0ac4 to add an option to the lane to not commit, so you can run it with:
```console
bundle exec fastlane download_translations skip_commit:true
```
While at it, I also changed it to group the changes in a single commit, as previously it would create a commit for each changed module.

> Speaking of that - is there any guarantee the translation will be ready in let's say a week? Is there any guarantee the translations will be done at all?

I don't think there are guarantees. Perhaps worth having a chat with the folks at Team Global (#i18n channel).

AdamGrzybkowski:
Thanks @iangmaia

PR #279, opened 2024-08-23
Initial implementation of the QuickEditor with the provided token

### LINKED ISSUE #251: QuickEditor: Option to launch the picker with the provided token
There should be a way to launch the picker with the provided token, assuming the third-party app has one already. This should bypass the OAuth flow and go directly to the picker.

AdamGrzybkowski:
Closes #251

### Description

**This is a DRAFT PR for now to discuss this solution**

An initial implementation of the `GravatarQuickEditor` that can be launched with the given auth token. This changes the flow and skips the OAuth completely since the third-party app should already be authorized to perform actions.

Because this is an "outside" token that is handled by the third-party app developers I don't want to store and only keep it in the "in-memory" token storage. For that I created an interface `TokenStorage` and now we will have two: DataStore and In-Memory implementations. A proper implementation will be injected to `AvatarsRepository` depending on how the QE was launched.

The above solution was mainly driven by the fact that `AvatarPicker` has no information about the token and I wanted to keep it that way. 

What do you think about that? Do you have other ideas how to handle this? 

### Testing Steps

You can add the token to the `local.properties` file like that:
`demo-app.bearer.token={token}`

This will be read in the demo app. Tap the checkbox next to the token if you want to launch the QE with it. 
Screenshot 2024-08-23 at 12 52 06

Testing the Log in/Close CTA:

1. With the provided token, it's simple - just use invalid token.
2. For the OAuth flow to ensure there's no regression use Charles of another app to mock network response.


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -65,3 +67,55 @@ internal fun GravatarQuickEditorPage(
...
+        exitTransition = { ExitTransition.None },
+    ) {
+        composable(QuickEditorPage.SPLASH.name) {
+            SplashPage(
+                email = gravatarQuickEditorParams.email,
+                token = authToken,
+            ) {
+                navController.navigate(QuickEditorPage.EDITOR.name)
+            }
+        }
```
AdamGrzybkowski, line 112: I was hoping to skip the `SplashPage` completely but there's a timing issue. 

`DisposableEffect` is run after the composition is successful, so if we only have the `AvatarPicker` its ViewModel will be initialized with the wrong dependencies, first, and after that the effect code will be run.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/AuthenticationMethod.kt
```diff
@@ -0,0 +1,36 @@
+package com.gravatar.quickeditor.ui.editor
+
+import com.gravatar.quickeditor.ui.oauth.OAuthParams
+import java.util.Objects
+
+/**
+ * Represents the authentication method used for the Gravatar Quick Editor.
+ */
+public sealed class AuthenticationMethod {
```
AdamGrzybkowski, line 9: We have two options:
1. Overload `GravatarQuickEditorBottomSheet` with different params: `OAuthParams` or `String` (for the token), or...

2. Have one `GravatarQuickEditorBottomSheet` that takes a sealed class as a parameter. I think this one is cleaner and better represents the options. It requires less copy pasting as well.

maxme: To be honest, both solutions look good to me. Let's go with version 2 since it's already there.

AdamGrzybkowski: ~I think I will also place that in the `GravatarQuickEditorParams` to keep the whole QE config in one place. WDYT?~ nvm, that's not a good idea since we need this to be separate

maxme (approved):
I got a really weird issue that I was able to capture in video.

- I made sure the bearer token checkbox was unchecked
- I clicked the logout button
- I clicked the "update avatar"
- I was expecting the login screen to show up, but I was still kind of logged in
- Notice that profile shows up (not the login button), but the avatar list doesn't load (session expired)
- Then when I check the token checkbox and retry, everything load as expected.

https://github.com/user-attachments/assets/a2ccf78e-2704-4036-a238-428a3d6e715a

I then built a debug version and launched the debugger to check why I had a valid token after logging out but I'm not able to reproduce this anymore, the logout button works as expected.


https://github.com/user-attachments/assets/1df738fe-f908-4884-b44b-233bca94cf86


I triple checked the code and couldn't find a way that some authentication data would remain when the checkbox is unchecked and after the logout button is pressed. 

I guess there was a network hiccup on my first test.

:shipit:

AdamGrzybkowski:
Thanks @maxme for the review.

I will add a few things, modify others, and request another review later.

AdamGrzybkowski:
@maxme I've added one commit that changes the CTA copy on the error section when the token is invalid. 

It shows "close" instead of "Log in" because we can handle the OAuth for this scenario. 

| Close | Log in |
|---|---|
| Screenshot 2024-09-04 at 17 21 13 | Screenshot 2024-09-05 at 09 43 01 |

maxme (approved):
I tried again to reproduce the weird "token exists but invalid" state, but I couldn't.

LGTM :shipit:

PR #282, opened 2024-08-29
QuickEditor: Avatar upload error UI

### LINKED ISSUE #267: QuickEditor: Avatar upload error UI
Implement an error overlay for the avatar with an option to retry.

AdamGrzybkowski:
Closes #267 

### Description

This PR adds an option to retry failed uploads by keeping those failed tries in the state and displaying them with an error indicator. The dialog with `remove upload` and `retry` options is shown when tapped.

I removed the `LocalAvatar` composable and put the "error" overlay in the `SelectableAvatar`. 
I added `AvatarLoadingState` sealed class to make the state modeling easier instead of a bunch of Bool values. 

### Testing Steps

2. Run the DemoApp 
3. Go to Avatar Update tab
4. Tap `Update Avatar` button
11. Enable request mocking for post requests or change the aspect ratio in the UCropCropperLauncher
12. Pick or take a photo and upload it
13. Confirm you see an error overlay
14. Tap it
15. Retry
16. Confirm you see an error overlay
17. Tap it
18. Remove upload
19. Confirm it's gone
20. Disable request mocking


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
```diff
@@ -4,82 +4,157 @@ import androidx.compose.foundation.background
...
+                    Modifier.border(4.dp, MaterialTheme.colorScheme.primary, RoundedCornerShape(cornerRadius))
+                } else {
+                    Modifier.border(
+                        1.dp,
+                        MaterialTheme.colorScheme.surfaceDim,
+                        RoundedCornerShape(cornerRadius),
+                    )
+                },
+            )
+    ) {
```
AdamGrzybkowski: I extracted this border to the outer composable so that it's not covered by the overlay.

AdamGrzybkowski: This one caused all the updates to the screenshot tests. To be honest not sure why, but there's like a one-pixel difference maybe - very hard to spot.

AdamGrzybkowski:
I will update this once we have a final OpenAPI spec.

AdamGrzybkowski:
This is ready for review now. 

I do have one small issue in the horizontal layout - the `scrollToItem` doesn't seem to work 100% when starting an upload. Investigating...

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/FailedAvatarUploadAlertDialog.kt
```diff
@@ -0,0 +1,43 @@
...
+                },
+                dismissButton = {
+                    TextButton(
+                        onClick = { onRemoveUploadClicked(avatarUploadFailure.uri) },
+                    ) { Text(stringResource(R.string.avatar_upload_failure_dialog_remove_upload)) }
+                },
+            )
+        }
+    }
+}
```
hamorillo, line 44: ⛏
```suggestion
}

@Preview
@Composable
private fun FailedAvatarUploadAlertDialogPreview() {
    FailedAvatarUploadAlertDialog(
        avatarUploadFailure = AvatarUploadFailure(Uri.EMPTY, "Error message"),
        onRemoveUploadClicked = {},
        onRetryClicked = {},
        onDismiss = {},
    )
}
```

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
```diff
@@ -21,66 +25,107 @@ import androidx.compose.ui.unit.dp
...
+    }
+}
+
+@Composable
+private fun FailureOverlay(modifier: Modifier = Modifier) {
+    Overlay(modifier) {
+        Icon(
+            imageVector = Icons.Rounded.Warning,
+            contentDescription = stringResource(R.string.failed_to_load_avatar_content_description),
+            tint = Color.White,
```
hamorillo, line 90: ⛏ If we use hardcoded colors and someone overrides the `GravatarTheme` to use a different color palette, these colors won't change. In this case, the `White` is hardcoded in the overlay, which has a hardcoded color, so at least it should always look good.

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModelTest.kt
```diff
@@ -526,6 +576,113 @@ class AvatarPickerViewModelTest {
...
+            assertEquals(AvatarUploadFailure(uri, errorMessage), awaitItem().failedUploadDialog)
+        }
+    }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Test
+    fun `given failed avatar upload dialog shown when FailedAvatarDismissed then UiState updated`() = runTest {
+        val uri = mockk<Uri>()
+        val identityAvatarsCopy = emailAvatars.copy(avatars = avatars, selectedAvatarId = "1")
+        every { fileUtils.deleteFile(any()) } returns Unit
```
hamorillo: ⛏
```suggestion

```

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModelTest.kt
```diff
@@ -526,6 +576,113 @@ class AvatarPickerViewModelTest {
         }
     }
 
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Test
+    fun `given failed avatar upload when FailedAvatarTapped then UiState updated`() = runTest {
+        val uri = mockk<Uri>()
+        val identityAvatarsCopy = emailAvatars.copy(avatars = avatars, selectedAvatarId = "1")
+        every { fileUtils.deleteFile(any()) } returns Unit
```
hamorillo: ⛏
```suggestion
```

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModelTest.kt
```diff
@@ -526,6 +576,113 @@ class AvatarPickerViewModelTest {
...
+            assertEquals(emptySet<AvatarUploadFailure>(), awaitItem.failedUploads)
+        }
+    }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Test
+    fun `given failed avatar upload dialog shown when FailedAvatarDialogDismissed then UiState updated`() = runTest {
+        val uri = mockk<Uri>()
+        val identityAvatarsCopy = emailAvatars.copy(avatars = avatars, selectedAvatarId = "1")
+        every { fileUtils.deleteFile(any()) } returns Unit
```
hamorillo: ⛏
```suggestion
```

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModelTest.kt
```diff
@@ -526,6 +576,113 @@ class AvatarPickerViewModelTest {
...
+            assertEquals(setOf(AvatarUploadFailure(uri, errorMessage)), awaitItem.failedUploads)
+        }
+    }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Test
+    fun `given failed avatar upload dialog shown when ImageCropped then UiState updated`() = runTest {
+        val uri = mockk<Uri>()
+        val identityAvatarsCopy = emailAvatars.copy(avatars = avatars, selectedAvatarId = "1")
+        every { fileUtils.deleteFile(any()) } returns Unit
```
hamorillo: ⛏
```suggestion

```

hamorillo (approved):
Works as described. 

A minor thing is that, as we are using the warning icon provided by Material Design, I am not sure if we have agreed with the iOS/design folks on a common icon.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
AdamGrzybkowski: I've used hardcoded colors here on purpose as we display the icon on top of an image - so we don't have a variable (light/dark) background here. 

What color from the theme should I use instead?

hamorillo: It's not about light/dark themes. We let developers fully [customize colors](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/docs/get-started/get-started.md#customization) inside the Gravatar SDK via GravataTheme.

For example, with just a few lines, I can modify how the demo-app shows the QE bottomsheet:

```kotlin
CompositionLocalProvider(LocalGravatarTheme provides object : GravatarTheme {
        @Composable
        get() = MaterialTheme.colorScheme.copy(surface = Color.Red, primary = Color.Green)
}) {

```

image

If we hardcode the colors here, this will be the only place where we won't use the theme-provided colors.

> What color from the theme should I use instead?

That's a good question. `MaterialTheme.colorScheme.error` sounds reasonable, but maybe it doesn't look as good as the white you've picked.

image

AdamGrzybkowski: >It's not about light/dark themes. We let developers fully [customize colors](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/docs/get-started/get-started.md#customization) inside the Gravatar SDK via GravataTheme.


You're right. Tbh, I think we shouldn't allow that in the QE. We should for `:ui` components, but not in the Editor. 

QuickEditor is not a component devs will embed into their screens directly - it's an in-app Gravatar app pretty much. So you launch it within the app but it should be clear you are in the "Gravatar" world in that bottom sheet.

hamorillo: > I think we shouldn't allow that in the QE. We should for :ui components, but not in the Editor.
> 
> So you launch it within the app but it should be clear you are in the "Gravatar" world

I completely see the point of that. However, we had similar discussions with the UI components, and in the end, if someone decides to customize the Editor, they can do it by forking the repo and applying a different theme. That might be a good discussion for the whole team.

PR #284, opened 2024-08-29
QuickEditor: Use contentPadding for avatars list

AdamGrzybkowski:
### Description

A small change to make the avatars go to the end of the section when scrolling horizontally. This can be accomplished with a simple `contentPadding` param in Compose. 

| Before | After | 
|---|---|
| Screenshot 2024-08-29 at 13 21 22 | Screenshot 2024-08-29 at 13 14 21 |

### Testing Steps

Important Note: The backend is currently returning the wrong format for the updated date field. As a workaround, you can apply this patch to comment that field from the Avatar object.

1. Open the QE and make sure you have enough avatars to scroll


maxme (approved):
LGTM :shipit:

PR #286, opened 2024-09-02
QuickEditor: MediaPickerPopup enter animation

### LINKED ISSUE #273: QuickEditor: MediaPicker popup animation
As suggested [here](https://github.com/Automattic/Gravatar-SDK-Android/pull/264#pullrequestreview-2245408442) by @mlumeau

AdamGrzybkowski:
Closes #273

### Description

This PR adds a `scaleIn` animation to the `MediaPickerPopup` using the `AnimatedVisibility` composable. 

Additionally, I've made some tweak to the popup itself:
1. Offset calculation is now completely based on the button, without hardcoded values.
2. Popup is focusable. This means when the user taps outside of it, the taps won't be propagates to the views below. For example, the user won't be able to select an image when the popup is visible without dismissing it first.
3. The Popup elevation, and tonal elevation were adjusted. 

Adding the `exit` animation seems to be more tricky, the popup is gone before you can see the animation but I hacked it a bit to see how it looks and I think no exit animation is better here. 

https://github.com/user-attachments/assets/b6a69ffb-f7e5-4b40-b950-b03220e02d05

@mlumeau Does this look good too you? 

### Testing Steps

2. Run the DemoApp 
3. Go to Avatar Update tab
5. Tap `Update Avatar` button
6. Follow the steps to log in (if necessary)
7. Tap the "Upload Image` button


mlumeau:
> Adding the exit animation seems to be more tricky, the popup is gone before you can see the animation but I hacked it a bit to see how it looks and I think no exit animation is better here.

Agreed, we don't need an exit animation.

maxme (approved):
LGTM, :shipit:

PR #289, opened 2024-09-03
DemoApp: Make it easier to test the QE

AdamGrzybkowski:
### Description

We have two activities set up to handle redirecting URLs from the OAuth flow. This PR tries to clean this up a bit by:
1. Configuring the `intent-filter` based on `local.properties` file. The `redirect_uri` can be different for any dev yet we had it harcoded in the AndroidManifest.xml.
2. Commenting the `intent-filter` of the XML activity.
3. Showing the option to launch the XML activity only in DEBUG builds. This is handy for us to test but I don't think we need that if we push the release version to internal testers. 

### Testing Steps

1. Run the demo app and make sure the OAuth works.
2. You can run the release build to test that the button to launch the XML Activity is gone. To do that you need to configure the signing config of the release build. This works as a quick solution `signingConfig = signingConfigs.getByName("debug")`

### DIFF
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -60,6 +60,10 @@ android {
                 "DEMO_WORDPRESS_REDIRECT_URI",
                 "\"${properties["demo-app.wordpress.oauth.redirectUri"]?.toString() ?: ""}\"",
             )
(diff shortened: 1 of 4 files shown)


maxme (approved):
Good idea, less steps to signin 👍 , :shipit:

PR #292, opened 2024-09-03
Adam/oauth user email

AdamGrzybkowski:
### Description

This PR updates the OAuth flow: 
1. `user_email` param is added to the URL to pre-fill the login form.
2. The `OAuthPage` UI is updated to match the design of the QE.

Screenshot 2024-09-03 at 13 41 49

The copy is from me and it's based on the invalid session error. I'm still not sure whether we should say to the user to "log in" or "authorize $app_name" kind of message. What do you think?

### Testing Steps

1. Start the demo app
2. Logout the user
3. Logout from the WP in the browser 
4. Start the QE
5. Confirm you see an email pre-filled
6. Tap "x" in the top bar
7. Confirm you see the update UI
8. Tap the "Log in" 
9. Finish the OAuth flow
10. Confirm it worked


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/WordPressOauth.kt
```diff
@@ -1,11 +1,20 @@
...
+            .scheme("https")
+            .authority("public-api.wordpress.com")
+            .appendPath("oauth2")
+            .appendPath("authorize")
+            .appendQueryParameter("client_id", clientId)
+            .appendQueryParameter("redirect_uri", redirectUri)
+            .appendQueryParameter("response_type", "code")
+            .appendQueryParameter("user_email", email.toString())
+            .build()
+            .toString()
```
AdamGrzybkowski: I think this is way cleaner and handles the URL encoding on to of that.

maxme: Code looks better that way, but we need to keep the `auth` scope parameter. If we don't keep it will [default to "full access to single site"](https://developer.wordpress.com/docs/oauth2/#2-optional-parameters-) which fails for users who don't have any site.

Example for the same user with no site:


`scope=auth`| no `scope` param
-|-
Screenshot_20240903_142021|Screenshot_20240903_142047

maxme (changes requested):
We should add the `scope` parameter back. Besides that, it's good to ship.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/WordPressOauth.kt
AdamGrzybkowski: hmm interesting, this works just fine for me. I'm not saying we shouldn't though, will ask Lighthouse folks. 

BTW. Looks like you have an app created in the WP console. You should see Gravatar UI not WP one. That might be the difference.

maxme: Right, I was using an old wp.com clientId/clientSecret pair. This works fine now with a gravatar.com clientId/clientSecret and the default scope. We don't need to add any scope if there is no gravatar specifics.

AdamGrzybkowski: I added the auth scope for now. We can update it later once we have the Gravatar scopes.

maxme: 👍 Created https://github.com/Automattic/Gravatar-SDK-Android/issues/294 to keep track of it

maxme (approved):
:shipit:

PR #293, opened 2024-09-03
QuickEditor: Avatar picker height 

AdamGrzybkowski:
### Description

It looks like I must have wrongly resolved some merge conflicts and the code that sets the height for the loading composable was removed. 

This PR fixes that in the first commit and additionally improves the refresh by skipping Profile data fetch if the data was loaded properly. 

### Testing Steps

1. Do not apply the patch, the parsing error helps with testing
2. Open the QE
3. Log in
4. See the error
5. Tap "Try again" 
6. Confirm the height of the bottom sheet stayed the same when showing the loading indicator
7. Confirm the profile was not fetched again. The skeleton view shouldn't be shown.


maxme (approved):
I confirm the bottom sheet height stays the same when retrying to login.

LGTM :shipit:

PR #299, opened 2024-09-04
QuickEditor: Fix weird glitch when opening the QE

### LINKED ISSUE #291: QuickEditor: Fix the jumping of the bottom sheet when transitioning from OAuth to the picker
This can be observed when going through the full flow OAuth -> Picker. The bottom sheet seems to change height more than it should.

AdamGrzybkowski:
Closes #291

### Description

Soo it's a tricky one. I can' reliably replicate this but I've found two places that don't look great: 
1. Opening the QE
2. Coming back to the QE from the custom tab

The issue exists only if the soft keyboard was visible before launching the QuickEditor. Here's how nr 1 looks like on Pixel 6 (I couldn't reproduce this on an emulator.): 

https://github.com/user-attachments/assets/349ee52d-031b-4fc7-a292-f921e8a337d1

Unfortunately, we can't fix this from the QE itself. The moment QE is launched is too late. This needs to be handled in the demo app or other third-party apps. Here's how it looks:

https://github.com/user-attachments/assets/909abce8-b53b-483c-a8d2-32f6b1290f48

When working on this I started playing with edge to edge to make sure this is not related and because this will be the default in Android 15 I think we should make sure we handle that. I enabled the edge-to-edge in the demo app and updated the QE insets a bit. 

### Testing Steps

1. Start the demo app
2. Tap on the text field to open the keyboard
3. Tap "Update avatar" 
4. Notice the bottom sheet slides from the bottom without any weird jumps


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -126,6 +127,7 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onEvent: (AvatarPickerEv
     Surface(
         Modifier
             .fillMaxWidth()
+            .animateContentSize()
```
AdamGrzybkowski, line 131: This smooths the UI and removes the UI jumping when loading the content.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -171,7 +173,7 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onEvent: (AvatarPickerEv
                             .fillMaxWidth(),
                     )
             }
-            Spacer(modifier = Modifier.height(24.dp))
+            Spacer(modifier = Modifier.height(10.dp))
```
AdamGrzybkowski, line 180: With the bottom insets, there's a bit too much padding now, so using a smaller value here.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
```diff
@@ -47,20 +47,21 @@ public fun GravatarQuickEditorBottomSheet(
...
 
     GravatarTheme {
         ModalBottomSheet(
-            modifier = Modifier
-                .heightIn(min = DEFAULT_PAGE_HEIGHT, max = screenHeightDp.dp * 0.9f),
             onDismissRequest = { onDismiss(GravatarQuickEditorDismissReason.Finished) },
             sheetState = modalBottomSheetState,
             dragHandle = { QEDragHandle() },
+            containerColor = MaterialTheme.colorScheme.surface,
+            tonalElevation = 1.dp,
```
AdamGrzybkowski, line 90: The `ModalBottomSheet` has some changes in the latest 1.3.0-rc01 version that make our bottom sheet look weird. These changes fix that. I decided not to update the library and keep those changes to make sure the QE will look properly if third-party apps have newer dependencies.

maxme (approved):
> I couldn't reproduce this on an emulator

Same, before the patch, I tried to disable the hardware keyboard on the emulator but it's correctly dismissed before the bottom sheet shows up

I could reproduce the issue on a Pixel 7 though and I confirm the patch is fixing it, so :shipit: !

PR #301, opened 2024-09-09
QuickEditor: Implement vertical content layout for the Avatar picker

### LINKED ISSUE #270: QuickEditor: Option to enable vertical scroll for avatar list
We should switch to a vertically scrolling list when the number of avatars exceeds 4. This should add a new type of upload button to the list with a "+" icon.

AdamGrzybkowski:
Closes #270 

### Description

This PR implements the vertically scrolling list of Avatars in the picker. This is configurable via `GravatarQuickEditorParams`.

Two three required some modifications to make this work:
- The vertically scrolling section had to be done separately to include the title and the description in the Lazy grid with max span item count (to take full width). This makes them both part of the scrolling items and thus they go away on scroll down leaving more space for the avatars.
- `skipPartiallyExpanded` state of the bottom sheet. This is set to false for vertically scrolling and additionally for `compact` height devices (mainly phones in landscape) to match the iOS behavior.
- `MediaPickerPopup` had to be modified to allow for global positioning on the screen. Up until now, it could only be positioned within the parent composable. This is not enough with the Upload button being part of the scrollable content, because it can place on the edge of the parent layout leaving no space for the popup. 

https://github.com/user-attachments/assets/a822f371-4e92-41d1-8617-d7390c382458

### Testing Steps

2. Run the DemoApp 
3. Go to Avatar Update tab
4. Tap the `Vertical` option
5. Tap `Update Avatar` button
6. Follow the steps to log in (if necessary)
7. Play with the picker  - make sure the `+` button works properly and the popup shows at the right place.
8. Mock the `/avatars` request to return empty list of avatars (or just delete your avatars)
9. Open the QE again and confirm the empty state matches the `horizontal` layout. 
10. Type `upload image` button and confirm the popup is shown properly. 
11. Do some regression tests for the `horizontal` scroll as well 🙏


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/ContentLayout.kt
```diff
@@ -0,0 +1,16 @@
+package com.gravatar.quickeditor.ui.editor
+
+/**
+ * The contentLayout of the Avatar picker in the Quick Editor.
+ */
+public sealed class ContentLayout {
```
AdamGrzybkowski: Any thoughts about the name?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
```diff
@@ -8,10 +8,12 @@ import java.util.Objects
  *
  * @property appName The appName of the app that is launching the Quick Editor
  * @property email The email of the user
+ * @property contentLayout The contentLayout of the Quick Editor.
  */
 public class GravatarQuickEditorParams private constructor(
     public val appName: String,
     public val email: Email,
+    public val contentLayout: ContentLayout,
```
AdamGrzybkowski: Technically this `contentLayout` applies to the avatar picker, but in the future, we might be launching other scopes for editions and this param might not make much sense. Do you think we should put it somewhere else?

maxme (changes requested):
Great work, it’s surprisingly straightforward to use with the two different scrolling areas.

I noticed a consistent crash in vertical mode when you rotate the screen:

```
 java.lang.IllegalArgumentException: The initial value must not be set to PartiallyExpanded if skipPartiallyExpanded is set to true.
 	at androidx.compose.material3.SheetState.<init>(SheetDefaults.kt:111)
 	at androidx.compose.material3.SheetState.<init>(SheetDefaults.kt:106)
 	at androidx.compose.material3.SheetState.<init>(SheetDefaults.kt:100)
 	at androidx.compose.material3.SheetState$Companion$Saver$2.invoke(SheetDefaults.kt:292)
 	at androidx.compose.material3.SheetState$Companion$Saver$2.invoke(SheetDefaults.kt:289)
 	at androidx.compose.runtime.saveable.SaverKt$Saver$1.restore(Saver.kt:68)
 	at androidx.compose.runtime.saveable.RememberSaveableKt.rememberSaveable(RememberSaveable.kt:87)
 	at androidx.compose.material3.SheetDefaultsKt.rememberSheetState(SheetDefaults.kt:484)
 	at androidx.compose.material3.ModalBottomSheet_androidKt.rememberModalBottomSheetState(ModalBottomSheet.android.kt:364)
 	at com.gravatar.quickeditor.ui.editor.bottomsheet.GravatarQuickEditorBottomSheetKt.GravatarQuickEditorBottomSheet(GravatarQuickEditorBottomSheet.kt:56)                                                                                                    	
```

AdamGrzybkowski:
Hmm interesting - kinda looks like a bug 🤔 The `initial value` is remembered from the state in the portrait mode, and when you rotate to landscape while having the bottom sheet partially expanded it remembers that state. But in the landscape, we want to skip it...

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
```diff
@@ -44,7 +46,33 @@ public fun GravatarQuickEditorBottomSheet(
...
+        onDismiss = onDismiss,
+        modalBottomSheetState = if (windowHeightSizeClass == WindowHeightSizeClass.COMPACT) {
+            rememberModalBottomSheetState(
+                skipPartiallyExpanded = true,
+            )
+        } else {
+            rememberModalBottomSheetState(
+                skipPartiallyExpanded = gravatarQuickEditorParams.contentLayout == ContentLayout.Horizontal,
+            )
+        },
```
AdamGrzybkowski: @maxme 
This is what I found to be working. 

Another temporary solution would be to skip the `windowHeightSizeClass == WindowHeightSizeClass.COMPACT`. It's not a dealbreaker. 

As mentioned, it feels like this is a bug. We have no control over the `initial state` so it seems quite wild to crash like that. I've reported it [here](https://issuetracker.google.com/issues/366067966).

demo-app/src/main/java/com/gravatar/demoapp/ui/AvatarUpdateTab.kt
```diff
@@ -66,6 +68,9 @@ fun AvatarUpdateTab(modifier: Modifier = Modifier) {
     val scope = rememberCoroutineScope()
     var avatarUrl: String? by remember { mutableStateOf(null) }
     val keyboardController = LocalSoftwareKeyboardController.current
+    var pickerContentLayout: ContentLayout by rememberSaveable(stateSaver = ContentLayoutSaver) {
+        mutableStateOf(ContentLayout.Horizontal)
+    }
```
AdamGrzybkowski: This wouldn't be persisted on screen orientation so that would also cause a crash. This is a bit specific for our demo app I think, so I kept the Saver private.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/ContentLayout.kt
maxme: Perhaps `LayoutOrientation` or `LayoutDirection`, it's less generic

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
maxme: Yeah, it already seems a bit out of place. If we add it as a parameter to `GravatarQuickEditorBottomSheet` and all, it will be all over the place. Were you thinking of adding some sort of `DisplaySettings`?

AdamGrzybkowski: One option that I was thinking about is something like:

```
quickEditorParams = GravatarQuickEditorParams {
      appName = ""
      email = ""
      scope = GravatarQuickEditorScope.Avatar(LayoutDirection.Horizontal)
}
```

The `scope` is what iOS and the web use in order to show different screens to edit. We don't have other screens so I skipped that param.

About the `DisplaySettings` - would you put them in the `GravatarQuickEditorParams`? I specifically created this one class to encapsulate all the configuration options, to not have many separate params in `GravatarQuickEditorBottomSheet`

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/ContentLayout.kt
AdamGrzybkowski: Sounds better, thanks!

AdamGrzybkowski: `AvatarPickerContentLayout` is what iOS uses [here](https://github.com/Automattic/Gravatar-SDK-iOS/blob/7c9cb44d706159423c090083764c2f0bbfd607fa/Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift#L23).

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
maxme: Yes, the `scope` will be better if we have more options. I think about future potential breaking changes for third parties and this kind of change, switching from a layoutDirection to scope, will be trivial: the build will break, they will notice the layoutDirection is wrapped, will replace it and be done.

I'm not sure sure we should change this now since we don't have plans for other options.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
maxme: Thanks for reporting it, I subscribed to the ticket.

AdamGrzybkowski:
@maxme Have you had a chance to review this one? I see some reactions from you so maybe you forgot to leave the review.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
AdamGrzybkowski: >I think about future potential breaking changes for third parties and this kind of change, switching from a layoutDirection to scope, will be trivial: the build will break, they will notice the layoutDirection is wrapped, will replace it and be done.


We shouldn't make breaking changes like that though. Unless bumping up the major version. So it's better to have it defined now in a way that will allow for some modifications. 

I'm leaning toward the suggested solution with the scope, but that's not in line with iOS.

maxme (approved):
Crash on orientation change is fixed 🎉 

LGTM :shipit:

hamorillo:
I've smoke-tested the demo app and skimmed the code. This is a big one, so I've struggled a bit reviewing all the code in depth. Also, there are some conflicts that may change the code slightly. 

As @maxme has already reviewed and approved it, let me know if you want me to do another round through the code :)

I couldn't reproduce any crash, so that looks fixed 🥳

❓ We let the third-party app choose the layout. We don't finally want to change it based on the number of avatars.

AdamGrzybkowski:
Thanks @hamorillo!

If you want you can leave your opinions about the public API - https://github.com/Automattic/Gravatar-SDK-Android/pull/301#discussion_r1750029779

>We let the third-party app choose the layout. We don't finally want to change it based on the number of avatars.

Not for now - it's a bit tricky with the initial state of the bottom sheet and making sure the CTA button is always visible. One of the reasons I used a sealed class for `ContentLayout` is that we can expand it with another option like `Treshold(count)`.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/ContentLayout.kt
hamorillo: `AvatarPickerContentLayout` sounds better to me.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
hamorillo: It's difficult to guess how the quick-editor will evolve. Maybe we decide to allow users to navigate between different "scopes". In that case, the SDK should allow multiple configurations (one for each possible scope). I don't know if that could make sense 🤷🏽‍♂️. 

Anyway, I would keep it as it is and if at some point we add more scopes, we'll be able to `@Deprecate` the old method until a new major release.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/ContentLayout.kt
AdamGrzybkowski: Renamed to match with iOS 👍

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorParams.kt
AdamGrzybkowski: Yeah, I agree we can modify it later without breaking changes.

AdamGrzybkowski:
There were quite a lot of conflicts - I did test it myself but if one of you could run the testing steps I would appreciate it 🙏

hamorillo:
> I did test it myself but if one of you could run the testing steps I would appreciate it 🙏

I've run the testing steps again and it works as expected.

PR #303, opened 2024-09-09
Downgrade kotlin to 1.9.25

### LINKED ISSUE #183: `gravatar`- Investigate issue when changing values in `local.properties`
I've noticed that after merging #173, when you update a property in `local.properties,` the change is not reflected when launching the app.

I've verified that disabling `kotlin.incremental` solves the issue. But I would like to investigate more before opening a PR with that solution that will affect compilation performance.

```
kotlin.incremental=false
```

I guess K2 does something different that is not working as expected.

### LINKED ISSUE #297: Revert Kotlin 2.0.0
This will enforce that version for all library consumers. That might be a big blocker for other apps to use our SDK. Let's go back to the 1.+ version. 

The PR to revert - https://github.com/Automattic/Gravatar-SDK-Android/pull/173

AdamGrzybkowski:
Closes #297 
Closes #183 

### Description

As per the issue description: 

```
This will enforce that version for all library consumers. That might be a big blocker for other apps to use our SDK. Let's go back to the 1.+ version.

The PR to revert - https://github.com/Automattic/Gravatar-SDK-Android/pull/173
```

### Testing Steps

Smoke test the demo app.

### DIFF
--- a/.idea/kotlinc.xml
+++ b/.idea/kotlinc.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="KotlinJpsPluginSettings">
-    <option name="version" value="2.0.0" />
+    <option name="version" value="1.9.25" />
   </component>
 </project>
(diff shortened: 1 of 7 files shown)


maxme (approved):
LGTM :shipit:

PR #308, opened 2024-09-10
Remove all deprecated classes/function from :gravatar and :gravatar-ui modules

### LINKED ISSUE #307: Remove all the deprecated classes and functions
Some were deprecated [here](https://github.com/Automattic/Gravatar-SDK-Android/pull/207) and [here](https://github.com/Automattic/Gravatar-SDK-Android/pull/206).

AdamGrzybkowski:
Closes #307

### Description

We're planning on releasing the new 2.0.0 version with the Quick Editor so it's a good time to clean up all the deprecated classes.

### Testing Steps

CI and smoke test the demo app.


maxme (approved):
Should we also drop the WP specific upload function? https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/gravatar/src/main/java/com/gravatar/services/AvatarService.kt#L31-L53 ?

If we want to update bump Gravatar version to WP and Woo apps, we'll replace the current implementation with the quick editor, right?

if we want to keep WP upload function, then :shipit:

AdamGrzybkowski:
That's a good question. This is used in [GravatarImagePickerWrapper](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/gravatar-ui/src/main/java/com/gravatar/ui/GravatarImagePickerWrapper.kt) but I do not know where this is used. 

Is this an equivalent of the QuickEditor? I think we can remove it in a separate PR anyway.

PR #309, opened 2024-09-10
Demo app with secrets

wzieba:
### Description

This PR adds support for `mobile-secrets` in order to provide OAuth secrets for the demo-app prototype build.

### Testing Steps

Install the app from the comment below and verify that Quick Editor works as expected.

### DIFF
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -10,13 +10,12 @@ plugins {
     alias(libs.plugins.detekt)
 }
 
-fun localProperties(): Properties {
-    val localProperties = Properties()
-    val localPropertiesFile = rootProject.file("secrets.properties")
-    if (localPropertiesFile.exists()) {
-        localProperties.load(FileInputStream(localPropertiesFile))
-    }
-    return localProperties
+fun secretProperties(): Properties {
+    val properties = Properties()
+    rootProject.file("secrets.properties")
+        .takeIf { it.exists() }
+        ?.let { properties.load(FileInputStream(it)) }
+    return properties
 }
 
 android {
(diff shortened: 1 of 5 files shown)


demo-app/build.gradle.kts
```diff
@@ -10,13 +10,13 @@ plugins {
...
-fun localProperties(): Properties {
-    val localProperties = Properties()
-    val localPropertiesFile = rootProject.file("local.properties")
-    if (localPropertiesFile.exists()) {
-        localProperties.load(FileInputStream(localPropertiesFile))
-    }
-    return localProperties
+fun secretProperties(): Properties {
+    val properties = Properties()
+    rootProject.file("secrets.properties")
```
wzieba, line 15: After merge, the team will have to move their current secrets from `local.properties` to `secrets.properties` or run `bundle exec fastlane run configure_apply` to use the shared secrets.

AdamGrzybkowski (approved):
Tested the build from the PR and it worked, thanks! 

One question: What steps would we have to take to add new properties (if needed)?

demo-app/build.gradle.kts
AdamGrzybkowski: When I try to run this command I get:
>[!] Decryption key could not be found

wzieba:
> One question: What steps would we have to take to add new properties (if needed)?

Good question! With review comment I see you don't have `.mobile-secrets` setup locally. Please configure this (check out "mobile secrets" page on the field guide) and then just edit `android/Gravatar-SDK-Android/secrets.properties` file and push the change.

After 2m, in this repository you have to run `bundle exec fastlane configure_update` and push the updated `.enc` file. It's all explained there with more details in the guide 👍

PR #310, opened 2024-09-11
Remove Java desugaring requirement

### LINKED ISSUE #295: Consider dropping Java Desugaring
This forces SDK consumers to enable it in their apps. Let's consider alternatives to the java.time.

AdamGrzybkowski:
Closes #295 

### Description

We shouldn't force third-party devs to enable desugaring and we shouldn't force them to use particular `java.time`  as they might want to use something else. 

This PR removes the `java.time` classes from the generated API models and the desugaring.

### DIFF
--- a/gravatar/build.gradle.kts
+++ b/gravatar/build.gradle.kts
@@ -113,13 +113,13 @@ openApiGenerate {
     )
     importMappings.set(
         mapOf(
-            "DateTime" to "java.time.Instant",
+            "DateTime" to "String",
         ),
     )
 
     typeMappings.set(
         mapOf(
-            "DateTime" to "java.time.Instant",
+            "DateTime" to "String",
         ),
     )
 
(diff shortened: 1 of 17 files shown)


gravatar/build.gradle.kts
```diff
@@ -117,13 +113,13 @@ openApiGenerate {
...
         mapOf(
-            "DateTime" to "java.time.Instant",
+            "DateTime" to "String",
         ),
     )
 
     typeMappings.set(
         mapOf(
-            "DateTime" to "java.time.Instant",
+            "DateTime" to "String",
```
maxme, line 122: TIL we could do that here 👍

maxme (approved):
Work as expected with current API (tested with `trunk` and `adam/294_scopes` merged in) - LGTM, :shipit:

PR #312, opened 2024-09-12
QuickEditor: Gravatar OAuth scopes

### LINKED ISSUE #294: Update the OAuth2 scope parameter after backend implements it

AdamGrzybkowski:
Closes #294 

### Description

Backend started to check the scopes we we need to request the proper ones. This PR adds all three:
```
- auth
- gravatar-profile:read
- gravatar-profile:manage
```

Additionally, I have removed the `Avatar.fullUrl` extension because backend returns full URL now. 

### Testing Steps

1. Apply the patch as usual.
2. Logout
3. Open the QE and follow OAuth
4. Confirm you can see the avatars. 
5. Upload an image.

Note. Selecting an avatar will crash. We need to generate new models based on the OpenAPI spec.


maxme (approved):
LGTM, :shipit:

PR #313, opened 2024-09-13
Do not use data class for AvatarQueryOptions

### LINKED ISSUE #302: Do not use `data` class for AvatarQueryOptions
We're planning to release version 2.0.0 so it's a good time to get rid of the problematic `data` class, as we did with all the REST API models.

AdamGrzybkowski:
Closes #302 

### Description

Using a `data class` in a public-facing API is problematic because we can't add more properties without introducing breaking changes. 

I'm changing the `AvatarQueryOptions` following the same patterns we used for API models. 

### Testing Steps
Run the demo app and test Avatar fetching.


hamorillo (approved):
LGTM! 🚀

PR #315, opened 2024-09-16
Run lint and upload results to Github Code Scanning

### LINKED ISSUE #118: Run the Android linter in CI
We should run the Android Linter to catch some warning / error that aren't caught by ktlint or detekt. We should also promote some warnings to errors (like `UnusedResources` to keep the library clean).

wzieba:
Closes #118 

### Description

This PR integrates Android Lint with GitHub Code Scanning support.

### Testing Steps

Green light from CI is just fine.


.buildkite/commands/lint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -u
+
+echo "--- 🧹 Linting"
+# Run lint on the app module first to ensure that the SARIF file is always generated.
+./gradlew :gravatar-quickeditor:lintRelease
```
AdamGrzybkowski: Technically `:quickeditor` is not an app module. Will the lint check also run against the `:ui` module?

hamorillo: > Will the lint check also run against the :ui module?

I've made a quick test, and I would say we need to run lint for each module.

wzieba: Because `quickeditor` uses `ui` module, the lint will also check the `ui` module (see: `checkDependencies = true`)

I configured it this way, because I want us to have a single lint report that we can send to GitHub Code Scanning registry. If we uploaded per-module reports, they would override each other.

AdamGrzybkowski: I guess that should work, we just need to keep in mind that this has to be updated if we have new modules.

How would that setup look if we had a new module that depends on the `:ui` but not on `:quickeditor`? 

Could we maybe run the lint against the demo app, but ignore errors/warnings from the demo app module?

hamorillo: > Because quickeditor uses ui module, the lint will also check the ui module (see: checkDependencies = true)

You are right. I've rechecked it, and it works as you described. Thanks!

wzieba: > Could we maybe run the lint against the demo app, but ignore errors/warnings from the demo app module?

Neat idea @AdamGrzybkowski , thanks! I think it's doable, with a little tweak in `lint.xml`: [ece18ea](https://github.com/Automattic/Gravatar-SDK-Android/pull/315/commits/ece18eaede2aaab46c95721b10130c41cd9c7e99)

When working on this, I've realised that I had misconfigured `lint.xml` previously. The updated configuration is now correct, but it reports errors. ~What should we do with them?~ I proposed fixes here

gravatar-quickeditor/src/main/res/values/available_languages.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--Warning: Auto-generated file, do not edit.-->
 <resources xmlns:tools="http://schemas.android.com/tools">
-  <string-array name="available_languages" translatable="false" tools:ignore="InconsistentArrays">
+  <string-array name="available_languages" translatable="false" tools:ignore="InconsistentArrays, UnusedResources">
```
wzieba, line 4: Is this ok @iangmaia ? I don't know how GlotPress integration works: is this file automatically generated, or is it fine to add configuration to it like this?

iangmaia: It was originally auto-generated, but I think this will be fine (as in: I don't think this will be overwritten, only the locale list should be).

.buildkite/commands/lint.sh
AdamGrzybkowski: I missed some uCrop related code in the `:ui` module that was making the Lint to fail (already useful :D). I pushed the changes.


EDIT: There was also some leftover `java.time` code 🙃

AdamGrzybkowski (approved):
Thanks @wzieba!

wzieba:
Thanks for the fixes @AdamGrzybkowski - it's great that lint caught e.g. 1b9a5f1!

PR #323, opened 2024-09-20
Replace Gson with Moshi

### LINKED ISSUE #177: Migrate from Gson to Kotlinx Serialization
Kotlinx Serialization is a more modern serialization library with Kotlin as a first-class citizen.

AdamGrzybkowski:
Closes #177

### Description

The Gson library we use for Json parsing is old and doesn't work well with Kotlin. I think we can safely call this a legacy code in our fairly new code base, so I think it's best to replace it with something more modern.

OpenAPI generator supports Gson, Jackson, Moshi, and Kotlin Serialization. 

Jackson is Java-first as Gson, so I didn't consider it at all.

Kotlin Serialization was my first initial pick but after trying to implement it I came to the conclusion that it may not be the best pick for an SDK:
1. It requires enabling the plugin in the app-level `build.gradle`. This would require some extra work for third-party devs to do something that's our internal implementation.
2. You can't set serializers for custom types like `URI` globally. [There are different ways to handle it](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#specifying-serializers-for-a-file) but each requires updates to the way models are generated (modifying .mustache templates) or exposing type aliases that again should be our implementation detail. 

Moshi it is then!  There are two options Reflection or Codegen. The latter requires ksp/kapt enabled and I'm worried this could cause potential conflict with third-party apps so I decided to use reflection for our SDK. 

### Testing Steps

Smoke test the Demo app including the QE.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -57,7 +57,6 @@ compose-ui-tooling-preview = { module = "androidx.compose.ui:ui-tooling-preview"
 compose-material3 = { module = "androidx.compose.material3:material3" }
 compose-material-icons = { module = "androidx.compose.material:material-icons-extended" }
 compose-adaptive = { group = "androidx.compose.material3.adaptive", name = "adaptive", version.ref = "materialAdaptive" }
-moshi = { group = "com.squareup.moshi", name = "moshi", version.ref = "moshi" }
 moshi-kotlin = { group = "com.squareup.moshi", name = "moshi-kotlin", version.ref = "moshi" }
 retrofit = { group = "com.squareup.retrofit2", name = "retrofit", version.ref = "retrofit" }
 retrofit-moshi-converter = { group = "com.squareup.retrofit2", name = "converter-moshi", version.ref = "retrofit" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -31,6 +31,7 @@ navigationCompose = "2.8.1"
 turbine = "1.1.0"
 dataStore = "1.1.1-beta03"
 startup = "1.1.1"
+moshi = "1.15.1"
 
 [libraries]
 kotlinx-coroutines = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-android", version.ref = "kotlinCoroutines" }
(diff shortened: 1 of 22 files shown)


AdamGrzybkowski:
@hamorillo @maxme I've updated this branch with the latest trunk.

gradle/libs.versions.toml
```diff
@@ -56,8 +57,10 @@ compose-ui-tooling-preview = { module = "androidx.compose.ui:ui-tooling-preview"
 compose-material3 = { module = "androidx.compose.material3:material3" }
 compose-material-icons = { module = "androidx.compose.material:material-icons-extended" }
 compose-adaptive = { group = "androidx.compose.material3.adaptive", name = "adaptive", version.ref = "materialAdaptive" }
+moshi = { group = "com.squareup.moshi", name = "moshi", version.ref = "moshi" }
```
hamorillo: ⛏
```suggestion
```

gravatar-quickeditor/build.gradle.kts
```diff
@@ -101,6 +101,8 @@ dependencies {
     implementation(project(":gravatar"))
     implementation(project(":gravatar-ui"))
 
+    implementation(libs.moshi)
+    implementation(libs.moshi.kotlin)
```
hamorillo: ⛏
```suggestion
```

gravatar/build.gradle.kts
```diff
@@ -69,8 +69,10 @@ android {
 
 dependencies {
     api(libs.okhttp)
+    implementation(libs.moshi)
```
hamorillo:
```suggestion
```

hamorillo (commented):
I don't think it should be a big deal for us but just want to mention:

> Note that the reflection adapter transitively depends on the kotlin-reflect library which is a 2.5 MiB .jar file.

Smoke-tested the app! ✅

AdamGrzybkowski:
> Note that the reflection adapter transitively depends on the kotlin-reflect library which is a 2.5 MiB .jar file.

I'm afraid there's not much we can do now. If we wanted to use moshi codegen we would have to change all our API models to not have `private` constructors. Then there's the problem with ksp/kapt compatibility when we use that in the SDK. I'm not 100% sure, but I imagine this would be a problem.

hamorillo:
Yeah, it's totally acceptable in our case.

hamorillo (approved):
🚀

PR #324, opened 2024-09-20
Parse the Error object from the error response body

### LINKED ISSUE #12: Replace `enum class ErrorType` by a sealed class to add properties to errors

AdamGrzybkowski:
Closes #12 

### Description

With the recently updated OpenAPI spec, we can now parse the error body to the `Error` object type when the API request fails. 

To pass that error down we need to update the `ErrorType` to a sealed class instead of an enum.

This is required for this [PR](https://github.com/Automattic/Gravatar-SDK-Android/pull/282).

### Testing Steps

Smoke test the Demo app. Test avatar upload - you can either mock the response or modify this [line](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/copperlauncher/UCropCropperLauncher.kt#L27) to change the aspect ratio.

### DIFF
--- a/gravatar/src/main/java/com/gravatar/services/HttpException.kt
+++ b/gravatar/src/main/java/com/gravatar/services/HttpException.kt
@@ -9,6 +9,7 @@ import retrofit2.Response
  * [message] The HTTP status message.
  */
 public class HttpException internal constructor(response: Response<*>) : RuntimeException() {
+    internal val rawErrorBody: String? = response.errorBody()?.string()
     public val code: Int = response.code()
-    public override val message: String = "HTTP ${response.code()} ${response.errorBody()?.string()}"
+    public override val message: String = "HTTP ${response.code()} $rawErrorBody"
 }
--- a/gravatar/src/main/java/com/gravatar/services/ServiceUtils.kt
+++ b/gravatar/src/main/java/com/gravatar/services/ServiceUtils.kt
@@ -1,5 +1,6 @@
 package com.gravatar.services
 
+import com.gravatar.di.container.GravatarSdkContainer
 import kotlinx.coroutines.CancellationException
 import com.gravatar.di.container.GravatarSdkContainer.Companion.instance as GravatarSdkDI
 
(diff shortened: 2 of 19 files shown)


gravatar/src/main/java/com/gravatar/services/HttpException.kt
```diff
@@ -9,6 +9,7 @@ import retrofit2.Response
  * [message] The HTTP status message.
  */
 public class HttpException internal constructor(response: Response<*>) : RuntimeException() {
+    internal val rawErrorBody: String? = response.errorBody()?.string()
```
AdamGrzybkowski, line 12: You can only read the error body once so we need it here if we want to expose it and keep the `message` property.

gravatar/src/main/java/com/gravatar/services/ServiceUtils.kt
```diff
@@ -10,12 +11,12 @@ internal inline fun <T> runCatchingRequest(block: () -> T?): Result<T, ErrorType
...
             Result.Success(result)
         } else {
-            Result.Failure(ErrorType.NOT_FOUND)
+            Result.Failure(ErrorType.NotFound)
         }
     } catch (cancellationException: CancellationException) {
         throw cancellationException
     } catch (ex: Exception) {
-        Result.Failure(ex.errorType())
+        Result.Failure(ex.errorType(GravatarSdkContainer.instance.gson))
```
hamorillo, line 19: ❓ Why pass `gson` instance through all methods instead of getting it in `errorTypeFromHttpCode` where it's used?

hamorillo (approved):
I left a minor ❓

LGTM! 🚀

gravatar/src/main/java/com/gravatar/services/ServiceUtils.kt
AdamGrzybkowski: Generally, I didn't want to access a static Gson instance from an extension function that we have tests for. It wouldn't be a problem now, but if we needed to configure it differently for the tests we would have to rely on the `GravatarSdkContainerRule`

PR #327, opened 2024-09-24
QuickEditor: Verify authorized account email matches with app's email

### LINKED ISSUE #306: QuickEditor: Verify the email matched the account used to authorize the app
The current OAuth flow won't force to user to use the correct email for the authorization step. We need to make sure that the user authorized the app with the account register with the same email that was used to open the QE. This should be possible via the new `/me` endpoint.

hamorillo:
Closes #306 

### Description

This PR enhances the OAuth flow to ensure the user has authorized the app with the correct email address. It implements a validation step using the new `/me/associated-email` endpoint to verify that the email used for authorization matches the one associated with the user's QE account. This change prevents potential mismatches when the user is already logged in with a different email.

**`Note:`** ~~Currently, we are handling that error like a generic OAuth flow. However, we may want to create a spefic case for it. We need to discuss this with the designers.~~
_Edited_: As discussed we should show an specific error for this case along with the "retry" (`Log in`) button

```
Title: Ooops

Message: It looks like you used the wrong email to log in. Please try again using email@domain.com this time. Thanks!

Button: "Log in"
```

### Testing Steps

**Note:** These testing steps assume you are not already logged in to the browser with any Gravatar account.
** Note 2: ** You'll need two Gravatar accounts. Let's say accounts `A` and `B`.

1. Navigate to the `Avatar Update` tab in the `demo-app`
2. With your account `A` click on the `Update Avatar` button
3. Follow the OAuth process with the account `A`
4. You should see the **Avatars Picker**, tap `Done` button
5. Fill the `Gravatar E-Mail` with the account `B`
6. Click the `Update Avatar` button
7. You should see the OAuth approval screen, but with `A` account
8. Tap on the `Approve` button
9. QE bottom sheet should show the specific error for this case
10. Tap over the `Log in` button
11. You should see the OAuth approval screen, but with `A` account
12. Tap on the `switch account` link
13. Follow the OAuth process with the account `B`
14. Tap on the `Approve` button
15. You should see the **Avatars Picker**


gravatar/src/main/java/com/gravatar/restapi/models/AssociatedEmail200Response.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.google.gson.annotations.SerializedName
+import java.util.Objects
+
+/**
+ *
+ *
+ * @param associated Whether the email is associated with a Gravatar account.
+ */
+
+public class AssociatedEmail200Response private constructor(
```
hamorillo, line 19: `AssociatedEmail200Response` The OpenApi generator auto-generated this class name because we haven't defined an object for the endpoint response. I've opened a PR on the backend to refactor this.

AdamGrzybkowski (commented):
The code looks good but as I mentioned elsewhere we should handle the wrong token here by showing a proper copy and an option to re-authorize. 

The way I see it whether we should show an error state with CTA or close the bottom sheet is :
- Can this be fixed by us or with a user action? Stay in the QE
- Can't be fixed? Close QE with an error. (an example here would be wrong OAuthParams)

gravatar/src/main/java/com/gravatar/restapi/models/AssociatedEmail200Response.kt
AdamGrzybkowski: Yeah, makes sense - I saw the same name for a class in iOS.

hamorillo:
I've updated PR as per discussion.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthUiState.kt
```diff
@@ -1,9 +1,17 @@
...
+    val status: OAuthStatus = OAuthStatus.LoginRequired,
 )
 
+internal sealed class OAuthStatus {
+    internal data object Authorizing : OAuthStatus()
+
+    internal data object LoginRequired : OAuthStatus()
+
+    internal data object WrongEmailAuthorized : OAuthStatus()
+}
```
AdamGrzybkowski, line 13: That's nice, I like it.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
```diff
@@ -28,14 +30,21 @@ internal class OAuthViewModel(
...
 
     init {
+        startOAuth()
+    }
+
+    fun startOAuth() {
         viewModelScope.launch {
             _actions.send(OAuthAction.StartOAuth)
+            _uiState.update { currentState ->
+                currentState.copy(status = OAuthStatus.LoginRequired)
```
AdamGrzybkowski, line 40: You can set that as default in this line
```
private val _uiState = MutableStateFlow(OAuthUiState(status = OAuthStatus.LoginRequired))
```

AdamGrzybkowski: Ohh it's already a default state so no need for that. I think you're setting it after the user taps the login if an email was wrong, right? I think I would keep the `WrongEmailAuthorized` status in case the user decides to close the browser to preserve that error message.

gravatar-quickeditor/src/main/res/values/strings.xml
```diff
@@ -28,6 +28,8 @@
     <string name="avatar_picker_network_error_message">Uh oh! Looks like you\'re not connected to the internet.</string>
     <string name="login_required">Login required</string>
     <string name="login_required_message">To modify your Gravatar profile you need to log&#160;in first</string>
+    <string name="oauth_wrong_email_authenticated_error_title">Ooops</string>
```
AdamGrzybkowski: We already have that one, I wonder if we should reuse it?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
hamorillo: > I think you're setting it after the user taps the login if an email was wrong, right?

Exactly.

> I think I would keep the WrongEmailAuthorized status in case the user decides to close the browser to preserve that error message.

I implemented that way because I thought about it like:

1.- The email was wrong
2.- The user sees the error and clicks the login button again
3.- The error is no longer valid; the user has seen it and opened the flow again. 

I'll ask iOS folk what they are doing.

gravatar-quickeditor/src/main/res/values/strings.xml
hamorillo: Yeah. I miss that one. I saw the others with more text, but not that one. 

Updated.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
AdamGrzybkowski: sure! Either way is okay I guess. I will approve then!

AdamGrzybkowski (approved):

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
hamorillo: 👍 I'll merge it, and if we decide on a different thing, I'll open another PR.

PR #328, opened 2024-09-24
QuickEditor: Migrate to Implicit OAuth flow

### LINKED ISSUE #288: QuickEditor: Use "token" response type for OAuth

AdamGrzybkowski:
Closes #288

### Description

This PR changes the `response` to `token` from `code`. This enabled the implicit OAuth flow where we receive the token directly in the Fragment part of the redirect URL.

I've updated the secrets with the new whitelisted redirect URL so that the generated demo app works fine.

### Testing Steps

Test the OAuth flow in the QuickEditor - make sure you use `wpcom-local-dev` scheme.
Test the same for the demo app built for this PR.


hamorillo (approved):
Tested CI demo-app ✅
Works great! 🚀

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthViewModel.kt
```diff
@@ -82,6 +64,9 @@ internal class OAuthViewModel(
                 }
 
                 is Result.Failure -> {
+                    _uiState.update { currentState ->
+                        currentState.copy(status = OAuthStatus.LoginRequired)
+                    }
                     _actions.send(OAuthAction.AuthorizationFailure)
```
AdamGrzybkowski: I wonder what would be best to do here. Technically, we have the `token` so we could assume it's correct and move to the editor, if it's invalid we do have the error handling there. 

Sending `OAuthAction.AuthorizationFailure` seems a bit harsh as it's closing the QE.

@hamorillo thoughts?

hamorillo: > we could assume it's correct and move to the editor, if it's invalid we do have the error handling there.

If we assume that the `token` is correct, we should also save it into the `tokenStorage`. If, for some reason, the token wasn't valid, I would say we don't have a way to recognize and delete it to request a new one. (I mean, we have a logout method, but at the moment, it is only provided for external use). Given that, I wouldn't assume the token is correct. 

> Sending OAuthAction.AuthorizationFailure seems a bit harsh as it's closing the QE.

I agree. Maybe a generic error with a retry action would do the trick. I don't think this endpoint should give us more problems than a timeout in case of poor connectivity or something similar.

AdamGrzybkowski: >If, for some reason, the token wasn't valid, I would say we don't have a way to recognize and delete it to request a new one.

We do, we will get 401 unauthorized response from the server and we will show an option to log in again. 

We are trying to cover an edge case scenario where the user will authorize with a wrong account. We were assuming the token is correct all the time until now, I think it's better to move the user to the editor. At the end, even if the token is invalid the user will have to perform the same number of taps to reauthorize again. 

I have pushed the code, given we can detect invalid token, but if you still prefer to not store it I will revert it.

hamorillo: > we will get 401 unauthorized response from the server and we will show an option to log in again.

That's not what happens, at least at the moment. 

I've mocked some responses with the new code. What I can see is that if the `associated-email` request fails, we save the wrong token (wrong meaning that it doesn't match the email but is a valid token). When requesting the avatars in the QE, we don't receive a 401 (which, as you mentioned, will restart the login flow), but we receive a 404, so we are trapped in the error.

I guess, the backend returns a 404 because we are authorized with a valid token. However, it can't find the requested email hash in the account associated with that token.

I can record a video with the mocked responses if you need it.

AdamGrzybkowski: Ohh got it - you're right. I will revert that and show an option to retry the email association check.

AdamGrzybkowski: Replaced the logic - we do show an error now with an option to "Try again"

Screenshot 2024-09-27 at 12 48 39

hamorillo (approved):
Thanks for the changes! Works great 🚀

PR #331, opened 2024-09-25
Update compose dependencies

AdamGrzybkowski:
### Description

This PR updates the compose bom to 2024.09.02 - it includes an update of the material3 library to the latest stable 1.3.0 version with some fixes to the bottom sheet. If we are to force a given version I would rather have the one with fewer bugs.

Additionally, I've updated the compose navigation to a stable version. 

### Testing Steps

Smoke test QE


AdamGrzybkowski:
It looks like there are some small changes to rendering the `rounded` components so I had to update the baseline.

AdamGrzybkowski:
🙃

uitestutils/src/main/java/com/gravatar/uitestutils/RoborazziTest.kt
```diff
@@ -43,6 +43,7 @@ abstract class RoborazziTest {
             roborazziOptions = RoborazziOptions(
                 compareOptions = RoborazziOptions.CompareOptions(
                     imageComparator = SimpleImageComparator(maxDistance = 0.007F, hShift = 1),
+                    changeThreshold = 0.01f,
```
AdamGrzybkowski, line 46: Not happy about that, but it's getting quite annoying to have tests fail on CI because of system differences. I made some changes some time ago that fixed the majority of issues, but now there's one test that fails `AvatarPickerListLoaded` - the dark version passes...

I think this will work for now and when time allows we should modify the CI flow to generate the baseline as per [this](https://github.com/Automattic/Gravatar-SDK-Android/pull/328).

gradle/libs.versions.toml
```diff
@@ -21,13 +21,13 @@ activityCompose = "1.9.0"
...
 kotlinCoroutines = "1.7.3"
-composeBom = "2024.06.00"
+composeBom = "2024.09.02"
 materialAdaptive = "1.0.0"
 browser = "1.8.0"
 constraintlayout = "2.1.4"
 activity = "1.9.0"
 androidXLifecycle = "2.8.3"
-navigationCompose = "2.8.0-beta05"
+navigationCompose = "2.8.1"
```
hamorillo, line 30: We have two more dependencies somehow related with compose that we can update:

```
activityCompose = "1.9.2"
composeUiTest = "1.7.2"
```

uitestutils/src/main/java/com/gravatar/uitestutils/RoborazziTest.kt
hamorillo: Yeah, I know the feeling. It's quite annoying. ->

```
The differ result :ComparisonResult(pixelDifferences=3, pixelCount=1404000, width=1080, height=1300) changed:true 
```

Can we create a ticket to deal with this when possible?

> I think this will work for now and when time allows we should modify the CI flow to generate the baseline as per this.

I wonder if you used the wrong link 🤔

hamorillo (approved):

uitestutils/src/main/java/com/gravatar/uitestutils/RoborazziTest.kt
AdamGrzybkowski: [yes, this is the correct one.](https://github.com/takahirom/roborazzi/?tab=readme-ov-file#q-why-do-my-screenshot-tests-fail-inconsistently-across-different-operating-systems-like-macos-ubuntu-and-windows)

AdamGrzybkowski: https://github.com/Automattic/Gravatar-SDK-Android/issues/333

PR #337, opened 2024-09-27
QuickEditor: Accept-Language header for Avatar requests

### LINKED ISSUE #311: Add language header to API calls
This will be used by the backend to return localized error messages.

eg. Accept-Language: en-US, fr-CA, es-ES

AdamGrzybkowski:
Closes #311

### Description

I've taken a different approach than those discussed in #311. Instead of providing a specific option to set the languages when configuring the `Gravatar` object, I've decided to reuse what we currently offer - passing an `OkHttpClient` with an already specified interceptor.

To achieve that I had to modify how we use the provided `OkHttpClient` - instead of just using it and ignoring Gravatar interceptors, I'm using it as a base and putting our interceptors on top of it. 

I feel like this solution solves issues we had with previous suggestions:
- We don't need any extra dependencies in the `:gravatar` module
- Users don't have to implement custom language providers etc. 
- We can set the language header per service whenever needed, same for third-party devs using QE, we won't override their configs as they will have different `OkHttpClients`. 

### Testing Steps

Run QE while monitoring the network. Check the headers for `/me/avatars` endpoint.

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/AcceptedLanguageInterceptor.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/AcceptedLanguageInterceptor.kt
@@ -1,0 +1,43 @@
+package com.gravatar.quickeditor.data
+
+import android.content.Context
+import androidx.core.app.LocaleManagerCompat
+import androidx.core.os.LocaleListCompat
+import okhttp3.Interceptor
+import okhttp3.Request
+import okhttp3.Response
+import java.util.Locale
+
+internal class AcceptedLanguageInterceptor(private val context: Context) : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        return chain.proceed(
+            chain.request().newBuilder()
+                .addAcceptedLanguageHeader(context.languagesList.asLocaleList.acceptedLanguageHeader).build(),
+        )
+    }
+
+    private fun Request.Builder.addAcceptedLanguageHeader(languages: String) = addHeader("Accept-Language", languages)
+
+    private val List<Locale>.acceptedLanguageHeader: String
+        get() {
+            var weight = 1.0F
+            return map { it.language }.reduce { accumulator, language ->
+                weight -= 0.1F
+                "$accumulator,$language;q=$weight"
+            }
+        }
+
+    private val LocaleListCompat.asLocaleList: List<Locale>
+        get() = buildList {
+            if (this@asLocaleList.size() > 0) {
+                for (index in 0 until this@asLocaleList.size()) {
+                    add(this@asLocaleList.get(index))
+                }
+            }
+        }.filterNotNull()
+
+    private val Context.languagesList: LocaleListCompat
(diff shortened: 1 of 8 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/AcceptedLanguageInterceptor.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    private fun Request.Builder.addAcceptedLanguageHeader(languages: String) = addHeader("Accept-Language", languages)
+
+    private val List<Locale>.acceptedLanguageHeader: String
+        get() {
+            var weight = 1.0F
+            return map { it.language }.reduce { accumulator, language ->
+                weight -= 0.1F
+                "$accumulator,$language;q=$weight"
+            }
```
hamorillo: This produces something like:

SCR-20240930-kizm

This is caused by the repeated 0.1F subtraction and how the floating-point works internally.

We can round those values or replace with something like:

```suggestion
            return map { it.language }.mapIndexed { index, language ->
                val weight = 1.0F - (index * 0.1F)
                "$language;q=$weight"
            }.joinToString(",")
```

Whose output is:
SCR-20240930-kkfm

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
```diff
@@ -31,12 +31,12 @@ internal class GravatarSdkContainer private constructor() {
...
     fun getGravatarV3Service(okHttpClient: OkHttpClient? = null, oauthToken: String? = null): GravatarApi {
         return getRetrofitApiV3Builder().apply {
-            client(okHttpClient ?: buildOkHttpClient(oauthToken))
+            client(okHttpClient.buildOkHttpClient(oauthToken))
         }.addConverterFactory(GsonConverterFactory.create(gson))
             .build().create(GravatarApi::class.java)
     }
 
-    private fun buildOkHttpClient(oauthToken: String?) = OkHttpClient()
+    private fun OkHttpClient?.buildOkHttpClient(oauthToken: String?) = (this ?: OkHttpClient())
```
hamorillo, line 43: We'll need to inform developers that we are overriding the `OkHttpClient` to avoid errors. I don't think several developers will provide a custom client, but just in case.

hamorillo (commented):
Nice approach 👏

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/AcceptedLanguageInterceptor.kt
AdamGrzybkowski: Great catch - I've fixed this by using `BigDecimal`. I added some tests as well.

gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt
AdamGrzybkowski: I added an explanation in the docs.

hamorillo (approved):
Works great! 🚀

PR #338, opened 2024-09-27
Remove WordPress references from the OAuth variable/secrets

### LINKED ISSUE #298: Remove references to WordPress in the gravatar library
Remove references to DEMO_WORDPRESS_X in the demo app.

There might be some other places where we used the wordpress bearer token that need to be clean up.

hamorillo:
Closes #298 

### Description

Removing WordPress references from the OAuth variables and secrets. 

**Note:** I've updated the secrets, so `demo-app` in CI should be correctly generated

### Testing Steps

- CI and code review
- Smoke test `demo-app` from CI (mainly the OAuth flow)


AdamGrzybkowski (approved):
Tested with the generated demo app, looks good 👍

PR #340, opened 2024-10-01
QuickEditor: ProfileCard using AvatarQueryOptions when requesting the Avatar

### LINKED ISSUE #281: QuickEditor: Improve quality of the initial avatar image in the profile card
This is the code that we use to show the Profile card:

```
ProfileSummary(
                state = it,
                modifier = Modifier
                    .fillMaxWidth()
                    .background(backgroundColor)
                    .padding(horizontal = 16.dp, vertical = 11.dp),
                avatar = {
                    Avatar(
                        state = profile.transform { profileValue -> profileValue.avatarUrl.toString() },
                        size = 72.dp,
                        modifier = Modifier.clip(CircleShape),
                    )
                },
)
```

The thing is - the `avatarUrl` property doesn't have the `size` param so a low quality image is loaded.

hamorillo:
Closes #281 

### Description

The Avatar component used by ProfileCard didn't allow AvatarQueryOptions to pass, so we completely relied on the `String` used to specify the URL.

`/userimage` endpoint only accepts `size=512` as a query parameter, while `/avatar` accepts both, `s=512` and `size=512`. To be more consistent, this PR migrates the query params to their long-form. 

Using the long form allow us to modify the `Avatar` composable to accept those AvatarQueryOptions when the URL is from `/userimage`.

### Testing Steps

1. Smoke test the `Avatar Tab` in the demo app. Everything should work as before, but use the "long" query parameters.
2. In the `Avatar Update Tab` with the Network Inspector opened, switch between different images. 
3. Verify the request contains the `size` param

image

### DIFF
--- a/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
@@ -11,6 +11,7 @@ import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import coil.compose.AsyncImage
 import com.gravatar.AvatarQueryOptions
+import com.gravatar.asQueryParameters
 import com.gravatar.extensions.avatarUrl
 import com.gravatar.extensions.defaultProfile
 import com.gravatar.restapi.models.Profile
@@ -19,6 +20,8 @@ import com.gravatar.ui.components.ComponentState
 import com.gravatar.ui.components.LoadingToLoadedProfileStatePreview
 import com.gravatar.ui.components.isNightModeEnabled
 import com.gravatar.ui.skeletonEffect
+import java.net.URI
+import java.net.URL
 
 /**
  * [Avatar] is a composable that displays a user's avatar.
@@ -88,16 +91,24 @@ public fun Avatar(
  * @param state Avatar URL wrapped in ComponentState
  * @param size The size of the avatar
  * @param modifier Composable modifier
+ * @param avatarQueryOptions Options to customize the avatar query
  */
 @JvmName("AvatarWithURLComponentState")
 @Composable
-public fun Avatar(state: ComponentState<String>, size: Dp, modifier: Modifier = Modifier) {
+public fun Avatar(
+    state: ComponentState<URI>,
+    size: Dp,
+    modifier: Modifier = Modifier,
+    avatarQueryOptions: AvatarQueryOptions? = null,
+) {
     when (state) {
         is ComponentState.Loading -> SkeletonAvatar(size = size, modifier = modifier)
 
         is ComponentState.Loaded -> {
             Avatar(
-                model = state.loadedValue,
+                model = state.loadedValue.toURL()?.let { url ->
+                    URL(url.protocol, url.host, url.path.plus(avatarQueryOptions.asQueryParameters()))
+                }.toString(),
                 size = size,
                 modifier = modifier,
             )
--- a/gravatar/src/main/java/com/gravatar/AvatarQueryOptions.kt
+++ b/gravatar/src/main/java/com/gravatar/AvatarQueryOptions.kt
@@ -107,6 +107,9 @@ public fun AvatarQueryOptions(initializer: AvatarQueryOptions.Builder.() -> Unit
     return AvatarQueryOptions.Builder().apply(initializer).build()
 }
 
+/**
+ * Function to convert the [AvatarQueryOptions] to a string with the query parameters.
+ */
 public fun AvatarQueryOptions?.asQueryParameters(): String {
     val queryList = mutableListOf<String>()
     this?.defaultAvatarOption?.let {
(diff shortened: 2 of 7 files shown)


gravatar/src/main/java/com/gravatar/AvatarQueryOptions.kt
```diff
@@ -105,3 +106,23 @@ public class AvatarQueryOptions private constructor(
 public fun AvatarQueryOptions(initializer: AvatarQueryOptions.Builder.() -> Unit): AvatarQueryOptions {
     return AvatarQueryOptions.Builder().apply(initializer).build()
 }
+
+/**
+ * Function to convert the [AvatarQueryOptions] to a string with the query parameters.
+ */
+public fun AvatarQueryOptions?.asQueryParameters(): String {
```
AdamGrzybkowski, line 113: If we leave it as an extension I would keep it internal. My understanding is that this was added internally in the `AvatarUrl` so the the whole config happens by using `AvatarQueryOptions` and not manually adding the params to the string, right?

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
```diff
@@ -88,16 +91,24 @@ public fun Avatar(
...
+) {
     when (state) {
         is ComponentState.Loading -> SkeletonAvatar(size = size, modifier = modifier)
 
         is ComponentState.Loaded -> {
             Avatar(
-                model = state.loadedValue,
+                model = state.loadedValue.toURL()?.let { url ->
+                    URL(url.protocol, url.host, url.path.plus(avatarQueryOptions.asQueryParameters()))
+                }.toString(),
```
AdamGrzybkowski, line 111: Couldn't we use the `AvatarUrl` like we do in the other `Avatar` composable that takes the full profile?

AdamGrzybkowski: In the `ProfileCard` the avatar URI after the selection is different `https://www.gravatar.com/userimage/...`. I wonder if this is okey to use the params that are defined for `https://gravatar.com/avatar/HASH`

Maybe instead of passing the `/userimage` we should pass the `/avatar` with the cache buster... 🤔

hamorillo: > Couldn't we use the AvatarUrl like we do in the other Avatar composable that takes the full profile?

`/userimage` doesn't contain the user email hash, and `AvatarUrl` works by using that hash so we couldn't use it in this case.

> Maybe instead of passing the /userimage we should pass the /avatar with the cache buster... 🤔

Yeah, it sounds like a valid approach. What do you think about adding a `cacheBuster` parameter to `AvatarQueryOptions` ? That way, we'll open it also to third-party developers.

AdamGrzybkowski: I believe the `AvatarQueryOptions` were designed to reflect [this](https://docs.gravatar.com/api/avatars/images/). At least the public stuff. We could have it there as internal though.

hamorillo: We can see the `cacheBuster` as part of the public queryParams. (It can not be there, but the fact is that we don't have another option to retrieve an updated avatar after uploading a new one). Basically, we need it in some cases, for example, in the QE, but any third-party would require the same if they have a custom Avatar component but use our SDK to upload and generate the AvatarUrls. (That can be us in PocketCast, for example).

> We could have it there as internal though.

I don't think it can be internal because it's used from the QE.

AdamGrzybkowski: >I don't think it can be internal because it's used from the QE.

Right.

I would avoid adding it as a public API because we might be able to change it if the server starts respecting the no-cache header. Once we add it to the public API we won't be able to easily remove it. 

What about following iOS:
- Add a `forceRefresh` param to `Avatar` composable
- Append the cache buster query separately

So in general, I would add an option to force a fresh image to our SDK so that third-party apps can use i, but keep the implementation internal. 

WDYT?

hamorillo: > I would avoid adding it as a public API because we might be able to change it if the server starts respecting the no-cache header.

I don't know if that would be anytime soon, but it is a fair point.

About the `forceRefresh` approach, I have some doubts, but let me think and play a bit more with the code.

hamorillo: ~~I am still considering an approach to `forceRefresh` that matches well with Jetpack Compose. I'll try to come up with something tomorrow, but if you have some ideas, feel free to share them.~~

Edited: Never mind, I'll open a PR with a proposal soon.

hamorillo: I've opened #350 to discuss the `forceRefresh` approach. If we decide to move forward with that, I'll close this one.

hamorillo:
Closing this PR in favor of #350

PR #342, opened 2024-10-01
Quick Editor docs

### LINKED ISSUE #253: QuickEditor: Write the documentation for the QE
Make sure to add the information about the backup rules - https://github.com/Automattic/Gravatar-SDK-Android/pull/250
Include the information about `Activity` setup - make sure to mention `singleTask` launchMode.

AdamGrzybkowski:
Closes #253 

### Description

This is the initial draft for the Quick Editor docs. I've rearranged it a bit by moving the example with the Profile service to the top and treating it as a separate section (one section per module).

### Testing Steps

-

### DIFF
--- a/docs/get-started/get-started.md
+++ b/docs/get-started/get-started.md
@@ -241,39 +241,14 @@ The `:gravatar-quickeditor` module provides a fully featured component that allo
 
 To do that the QuickEditor needs an authorization token to perform requests on behalf of the user. There are two ways for that:
 
-### 1. Obtain the token yourself and provide it to the Quick Editor
-
-Quick Editor can be launched with the provided token. To obtain it, you have to follow the [OAuth docs](https://docs.gravatar.com/?page_id=2555&preview=true) and implement the OAuth flow within your app.
-
-Once you have the token, here's how you can embed the QuickEditor in your Compose screen:
-
-```kotlin
-var showBottomSheet by rememberSaveable { mutableStateOf(false) }
-
-if (showBottomSheet) {
-    GravatarQuickEditorBottomSheet(
-        gravatarQuickEditorParams = GravatarQuickEditorParams {
-            email = Email("{USER_EMAIL}")
-            avatarPickerContentLayout = AvatarPickerContentLayout.Horizontal
-        },
-        authenticationMethod = AuthenticationMethod.Bearer("{TOKEN}"),
-        onAvatarSelected = { avatarUpdateResult -> ... },
-        onDismiss = { gravatarQuickEditorDismissReason ->
-            showBottomSheet = false
-            ...
-        },
-    )
-}
-```
-
-### 2. Leave the OAuth flow to Quick Editor
+### 1. Let the Quick Editor handle the OAuth flow
 
 Quick Editor can handle the heavy lifting of running the full OAuth flow, so you don't have to do that. We will still need a few things from you.
 First, you have to go to [OAuth docs](https://docs.gravatar.com/?page_id=2555&preview=true) and create your Application. Define the `Redirect URLs`.
 
 > Keep in mind that you need to use the `https` scheme. Internally, QuickEditor uses Implicit OAuth flow (`response_type=token`) and for security reasons, the server doesn't allow custom URL schemes.
 
-For the sake of this example let's assume the redirect URL is `https://gravatar.com/redirect-url`.
+For the sake of this example let's assume the redirect URL is `https://yourhost.com/redirect-url`.
 
 In your `AndroidManifest.xml` you need to add an `<intent-filter>` and the `android:launchMode="singleTask"` to the activity that will 
 launch the Quick Editor (or the last/main activity depending on your app architecture). This is important because the Quick Editor will be waiting for the `onNewIntent()` callback to handle OAuth redirection.
@@ -291,7 +266,7 @@ launch the Quick Editor (or the last/main activity depending on your app archite
 
         <data
             android:scheme="https"
-            android:host="gravatar.com"
+            android:host="yourhost.com"
             android:pathPrefix="/redirect-url"
         />
     </intent-filter>
@@ -314,7 +289,7 @@ if (showBottomSheet) {
         authenticationMethod = AuthenticationMethod.OAuth(
             OAuthParams {
                 clientId = "{YOUR_CLIENT_ID}"
-                redirectUri = "{YOUR_REDIRECT_URL}" // In our example this would be https://gravatar.com/redirect_url
+                redirectUri = "{YOUR_REDIRECT_URL}" // In our example this would be https://yourhost.com/redirect_url
             },
         ),
         onAvatarSelected = { avatarUpdateResult -> ... },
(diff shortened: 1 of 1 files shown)


docs/get-started/get-started.md
```diff
@@ -239,4 +242,172 @@ CompositionLocalProvider(LocalGravatarTheme provides object : GravatarTheme {
...
+            ...
+        },
+    )
+}
+```
+
+### 2. Leave the OAuth flow to Quick Editor
+
+Quick Editor can handle the heavy lifting of running the full OAuth flow, so you don't have to do that. We will still need a few things from you.
+First, you have to go to [OAuth docs](https://docs.gravatar.com/?page_id=2555&preview=true) and create your Application. Define the `Redirect URLs`.
```
AdamGrzybkowski: This is not published yet, we either have to remove it or wait for it.

docs/get-started/get-started.md
```diff
@@ -239,4 +242,172 @@ CompositionLocalProvider(LocalGravatarTheme provides object : GravatarTheme {
...
+        <data
+            android:scheme="https"
+            android:host="gravatar.com"
+            android:pathPrefix="/redirect-url"
+        />
+    </intent-filter>
+</activity>
+```
+
+> Make sure to follow official [Android App Links documentation](https://developer.android.com/training/app-links#android-app-links) to properly setup App Link.
```
AdamGrzybkowski, line 276: Not sure if this is enough - but copy-pasting these docs doesn't seem like the best idea either.

docs/get-started/get-started.md
```diff
@@ -239,4 +242,172 @@ CompositionLocalProvider(LocalGravatarTheme provides object : GravatarTheme {
...
+    <intent-filter android:autoVerify="true">
+        <action android:name="android.intent.action.VIEW" />
+
+        <category android:name="android.intent.category.DEFAULT" />
+        <category android:name="android.intent.category.BROWSABLE" />
+
+        <data
+            android:scheme="https"
+            android:host="gravatar.com"
+            android:pathPrefix="/redirect-url"
```
AdamGrzybkowski: Is it confusing if I use the real example here? Should I use sth like: {SCHEME} instead?

docs/get-started/get-started.md
```diff
@@ -239,4 +242,172 @@ CompositionLocalProvider(LocalGravatarTheme provides object : GravatarTheme {
...
+
+With the provided `clientId` and the `redirectUrl` Quick Editor can launch and handle the full OAuth flow. Once obtained the token will be stored in an encrypted Data Store.
+This token will be later used in subsequent Quick Editor launches to make the user experience more seamless by not having to go through the OAuth flow each time.
+
+When the user logs out form the app, make sure to run:
+
+```kotlin
+GravatarQuickEditor.logout(Email("{USER_EMAIL}"))
+```
+
```
AdamGrzybkowski, line 312: Should we put more emphasis on this part?

docs/get-started/get-started.md
```diff
@@ -239,4 +242,172 @@ CompositionLocalProvider(LocalGravatarTheme provides object : GravatarTheme {
...
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <exclude
+        domain="sharedpref"
+        path="__androidx_security_crypto_encrypted_file_pref__.xml" />
+    <exclude
+        domain="file"
+        path="datastore/quick-editor-preferences.preferences_pb" />
+</full-backup-content>
+```
```
AdamGrzybkowski, line 367: This is technically optional because the SDK will catch an exception and clean the old files from the backup. Should we hide it/move it?

docs/get-started/get-started.md
```diff
@@ -239,4 +242,172 @@ CompositionLocalProvider(LocalGravatarTheme provides object : GravatarTheme {
...
+
+## Quick Editor
+
+The `:gravatar-quickeditor` module provides a fully featured component that allows the user to modify their avatar without leaving your app. 
+
+To do that the QuickEditor needs an authorization token to perform requests on behalf of the user. There are two ways for that:
+
+### 1. Obtain the token yourself and provide it to the Quick Editor
+
+Quick Editor can be launched with the provided token. To obtain it, you have to follow the [OAuth docs](https://docs.gravatar.com/?page_id=2555&preview=true) and implement the OAuth flow within your app.
```
hamorillo: 💡 I wonder if we should position this method first instead of "Leaving the OAuth flow to Quick Editor." As the other method is more straightforward and takes care of saving the token, I would switch the order.

docs/get-started/get-started.md
hamorillo: As we required it to be `https` I would keep the scheme as it is but just modify the host:

```suggestion
        <data
            android:scheme="https"
            android:host="{YOUR-HOST}"
            android:pathPrefix="/redirect-url"
```

Or maybe we can use another example URL instead of the one containing "gravatar.com" because, yeah, it can be confusing. Something like: `https://yourhost.com/redirect-url`

docs/get-started/get-started.md
hamorillo: Agree. I hope it's enough 🤞🏼

docs/get-started/get-started.md
hamorillo: I think that's OK

docs/get-started/get-started.md
hamorillo: If we are ready to handle it internally, I think we can hide this because it makes the SDK more challenging to integrate. If we don't crash, the final result will be the same, right?

AdamGrzybkowski: Yes, although I think excluding those files from the backup is a better practice.

docs/get-started/get-started.md
AdamGrzybkowski: Sure - from the SDK integration point of view, the simpler way is if you provide the token, but only for the apps that have it already. I'm okay with reordering.

docs/get-started/get-started.md
AdamGrzybkowski: Done

docs/get-started/get-started.md
AdamGrzybkowski: I put that in the `<details>` tag.

docs/get-started/get-started.md
AdamGrzybkowski: Done

docs/get-started/get-started.md
```diff
@@ -239,4 +242,179 @@ CompositionLocalProvider(LocalGravatarTheme provides object : GravatarTheme {
...
+if (showBottomSheet) {
+    GravatarQuickEditorBottomSheet(
+        gravatarQuickEditorParams = GravatarQuickEditorParams {
+            email = Email("{USER_EMAIL}")
+            avatarPickerContentLayout = AvatarPickerContentLayout.Horizontal
+        },
+        authenticationMethod = AuthenticationMethod.OAuth(
+            OAuthParams {
+                clientId = "{YOUR_CLIENT_ID}"
+                redirectUri = "{YOUR_REDIRECT_URL}" // In our example this would be https://gravatar.com/redirect_url
```
hamorillo: ⛏
```suggestion
                redirectUri = "{YOUR_REDIRECT_URL}" // In our example this would be https://yourhost.com/redirect_url
```

hamorillo (approved):
LGTM! 

We still have [this](https://github.com/Automattic/Gravatar-SDK-Android/pull/342/files#r1782750135) pending. I'm not sure what to say... we can merge as it is and if when we plan to release as stable (not rc) remove it.

PR #343, opened 2024-10-01
QuickEditor: Specify the proper avatar size in the Avatar Sections

### LINKED ISSUE #341: QuickEditor: AvatarsSections don't load proper size of the avatars
At the moment, the avatar endpoint response is something like:

```json
[{
	"image_id": "82e14b248c959aef4f90f52e6dbc2fc5",
	"image_url": "https://0.gravatar.com/userimage/252014521/82e14b248c959aef4f90f52e6dbc2fc5?size=512",
	"rating": "G",
	"updated_date": "2024-09-30T10:58:15Z",
	"alt_text": "TEST's avatar"
}
...
```

As we receive a size in the object, the avatars loaded in the AvatarsSections look good. However, if at some point we remove that (which I'm not pretty sure why is there), the avatar won't be as good as it is now:

| Now | Without harcoded size |
|----------|-----------|
|image|image

hamorillo:
Closes #341

### Description

We should request the proper side for the images in the avatar list in the QE. This PR specifies the correct dimension using the `size` query parameter.

### Testing Steps

- Using a network inspector, open the QE and verify you can see the `/userimage` requests with the proper size. It should be different from 512 which is being returned by the backend in the URL.
image

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/LazyAvatarList.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/LazyAvatarList.kt
@@ -8,8 +8,11 @@ import androidx.compose.foundation.lazy.LazyRow
 import androidx.compose.foundation.lazy.items
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import com.gravatar.quickeditor.ui.avatarpicker.AvatarUi
+import java.net.URL
 
 @Composable
 internal fun LazyAvatarRow(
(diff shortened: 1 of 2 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/LazyAvatarList.kt
```diff
@@ -30,6 +33,7 @@ internal fun LazyAvatarRow(
             Avatar(
                 avatar = avatarModel,
                 onAvatarSelected = { onAvatarSelected(avatarModel) },
+                size = avatarSize,
                 modifier = Modifier.size(avatarSize),
```
AdamGrzybkowski, line 37: We can move the `Modifier.size(avatarSize)` to the `Avatar` component now that we pass the `size` separately.

AdamGrzybkowski (approved):

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/LazyAvatarList.kt
hamorillo: Done!

hamorillo: I've reverted the last commit because it broke the UI (not sure why). I'll merge like this, and I'll take a look later.

PR #344, opened 2024-10-02
Remove appname from GravatarQuickEditorParams

AdamGrzybkowski:
### Description

The `appName` is not used now, so let's remove it. Additionally, I've noticed some overridden functions were incorrectly implemented so fixed those as well. 

### Testing Steps

CI should be enough.


hamorillo (approved):
Apart from removing the applicationName val from AvatarUpdateTab as the static analisys says it LGTM. 🚀

PR #346, opened 2024-10-02
Rename Result to GravatarResult

### LINKED ISSUE #296: Drop Result type in favor of kotlin.Result
Gravata.Result type doesn't add anything extra and as per this [post](https://kotlinlang.org/docs/api-guidelines-simplicity.html#reuse-existing-concepts) this will reduce the mental complexity and shrink the size of our API.

AdamGrzybkowski:
Closes #296 

### Description

As per the discussion in #296 - the `Result` could be easily confused with `kotlin.Result` so we won't to rename it for less confusion and better autocompletion. 

### Testing Steps

CI should be fine. Check the docs.

### DIFF
--- a/gravatar/src/main/java/com/gravatar/services/AvatarService.kt
+++ b/gravatar/src/main/java/com/gravatar/services/AvatarService.kt
@@ -51,15 +51,17 @@ public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
 
     /**
      * Uploads an image to be used as Gravatar avatar.
-     * This method will catch any exception that occurs during the execution and return it as a [Result.Failure].
+     * This method will catch any exception that occurs during
+     * the execution and return it as a [GravatarResult.Failure].
      *
      * @param file The image file to upload
      * @param oauthToken The OAuth token to use for authentication
      * @return The result of the operation
      */
-    public suspend fun uploadCatching(file: File, oauthToken: String): Result<Avatar, ErrorType> = runCatchingRequest {
-        upload(file, oauthToken)
-    }
+    public suspend fun uploadCatching(file: File, oauthToken: String): GravatarResult<Avatar, ErrorType> =
+        runCatchingRequest {
+            upload(file, oauthToken)
+        }
 
     /**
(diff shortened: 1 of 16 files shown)


gravatar/src/main/java/com/gravatar/services/GravatarResult.kt
```diff
@@ -3,15 +3,15 @@ package com.gravatar.services
...
+public sealed class GravatarResult<T, E> {
     /**
      * Represents a successful fetch operation.
      *
      * @param T The type of the value fetched
      * @param E The error type
      * @property value The fetched value
      */
-    public data class Success<T, E>(public val value: T) : Result<T, E>()
+    public data class Success<T, E>(public val value: T) : GravatarResult<T, E>()
```
AdamGrzybkowski, line 14: Should we drop the `data` here?

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -51,15 +51,17 @@ public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
 
     /**
      * Uploads an image to be used as Gravatar avatar.
-     * This method will catch any exception that occurs during the execution and return it as a [Result.Failure].
+     * This method will catch any exception that occurs during
+     * the execution and return it as a [Result.Failure].
```
hamorillo:
```suggestion
     * the execution and return it as a [GravatarResult.Failure].
```

gravatar/src/main/java/com/gravatar/services/GravatarResult.kt
hamorillo: In this case, I think we can keep it, or do you see a risk in that?

hamorillo (approved):
🚀

gravatar/src/main/java/com/gravatar/services/GravatarResult.kt
AdamGrzybkowski: I don't think we would add more properties to it, so I do also think it's safe.

PR #348, opened 2024-10-02
QuickEditor: Use uCrop fork

### LINKED ISSUE #305: QuickEditor: Downsize uploaded image
There's a bug in UCrop that cause the cropped images to keep their original file size. 

The [PR](https://github.com/Yalantis/uCrop/pull/926) with a fix is open but it seems unlikely we can get it merged.

AdamGrzybkowski:
Closes #305

### Description

This PR replaces the og uCrop with Autmattic fork.

cc @ParaskP7 

### Testing Steps

1.Run the Demo App 
2. Upload an image from the camera using a real device
3. Verify that the image size is smaller than the original (either stop the app before deleting an image and check in Device Explorer or monitor the API request)

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -25,7 +25,7 @@ dependencyResolutionManagement {
     repositories {
         google()
         mavenCentral()
-        maven { url = uri("https://jitpack.io") }
+        maven { url = uri("https://a8c-libs.s3.amazonaws.com/android") }
     }
 }
 
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -25,7 +25,13 @@ dependencyResolutionManagement {
     repositories {
         google()
         mavenCentral()
-        maven { url = uri("https://a8c-libs.s3.amazonaws.com/android") }
+        maven {
+            url = uri("https://a8c-libs.s3.amazonaws.com/android")
+            content {
+                includeGroup("com.automattic")
+                includeGroup("com.automattic.ucrop")
+            }
+        }
     }
 }
 
--- a/README.md
+++ b/README.md
@@ -172,13 +172,6 @@ repositories {
     maven {
         url "https://a8c-libs.s3.amazonaws.com/android"
     }
-    // Jitpack is used to fetch the uCrop library. Required only for gravatar-quickeditor module.
-    maven {
-        url "https://jitpack.io"
-        content {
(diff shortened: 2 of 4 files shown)


settings.gradle.kts
```diff
@@ -25,7 +25,7 @@ dependencyResolutionManagement {
     repositories {
         google()
         mavenCentral()
-        maven { url = uri("https://jitpack.io") }
+        maven { url = uri("https://a8c-libs.s3.amazonaws.com/android") }
```
ParaskP7: FYI: This configuration below is what you actually need:

```groovy
        maven {
            url = uri("https://a8c-libs.s3.amazonaws.com/android")
            content {
                includeGroup("com.automattic")
                includeGroup("com.automattic.ucrop")
            }
        }
```

AdamGrzybkowski: hmm what does this add? It works fine without it 🤔

ParaskP7: Good question @AdamGrzybkowski ! 💯 

So, the purpose of `content/includeGroup` is:
1. **Artifact Filtering**: The `content` block defines which artifacts are allowed to be downloaded from this specific repository.
2. **Group Restrictions**: By using `includeGroup`, the configuration limits the repository to only provide artifacts from the specified group IDs.

So, in our case, the repository configuration I shared above will do the following:
1. **Include com.automattic**: Allow artifacts with the group ID "com.automattic" to be downloaded from this repository.
2. **Include com.automattic.ucrop**: Also permit artifacts with the group ID "com.automattic.ucrop" to be retrieved from this repository.

This setup is in general what we have been applying to all our repos, just to optimize dependency resolution by only checking for the specified groups, prevent unintended artifacts from being downloaded and explicitly define which artifact groups are expected from this particular repository.

As such, this configuration ensures that only artifacts from the `com.automattic` and `com.automattic.ucrop` groups will be considered when resolving dependencies from the `https://a8c-libs.s3.amazonaws.com/android` repository. Any other group IDs will be ignored for this specific repository, even if they exist there.

Does that all make sense to you? 🙏

AdamGrzybkowski: Thanks for the explanation, makes total sense!

ParaskP7: Awesome, thank YOU for the inquiry! 🥇

hamorillo (approved):
:shipit:

PR #350, opened 2024-10-03
QuickEditor: ProfileCard using forceRefresh option for Avatar

### LINKED ISSUE #281: QuickEditor: Improve quality of the initial avatar image in the profile card

hamorillo:
Closes #281 

### Description

This PR uses a different approach than #340 to solve #281.

As discussed [here](https://github.com/Automattic/Gravatar-SDK-Android/pull/340#discussion_r1784288363) we can implement a `forceRefresh` param in the `Avatar` composable.

With this approach, we don't need a new Avatar composable. In addition, we increase the versatility of our current component.

If `forceRefresh` is enabled, everytime the `Avatar` composable is recomposed we'll update the cache buster internally. When we switch off that flag, we'll continue using the last `cacheBuster` used.

On the QE side, as the profile does not change, we need to force the recomposition. In this case, we use the "key" with the selected image ID.

**Important:** To me, this approach looks better than #340 and also reduces the number of composables we need to maintain. However, the avatar is not always updated as expected because we request the new avatar too fast. We'll need to introduce a delay to postpone the recomposition a bit. You can reproduce it by switching between already uploaded avatars. Wdyt?

<details>
<summary> Without delay </summary>

https://github.com/user-attachments/assets/d299f9fb-2dee-41f1-9313-61e93d4ba529
</details>

<details>
<summary> With Delay </summary>

https://github.com/user-attachments/assets/0c948716-90e4-4c68-bd6b-2142164ed40e
</details>

### Testing Steps

1. Smoke test the `Avatar Tab` in the demo app. Everything should work as before.
2. In the `Avatar Update Tab` with the Network Inspector opened, switch between different images. 
3. Verify the request contains the `size` param and the `cacheBuster`

image

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
@@ -24,6 +24,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.key
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
@@ -146,10 +147,13 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onEvent: (AvatarPickerEv
                     .fillMaxWidth()
                     .padding(bottom = 10.dp),
             )
-            ProfileCard(
-                profile = uiState.profile,
-                modifier = Modifier.padding(horizontal = 16.dp),
-            )
+            key(uiState.emailAvatars?.selectedAvatarId) {
+                ProfileCard(
+                    profile = uiState.profile,
+                    modifier = Modifier.padding(horizontal = 16.dp),
+                )
+            }
+
             val sectionModifier = Modifier.padding(top = 24.dp, bottom = 10.dp)
             when {
                 uiState.isLoading -> Box(
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
@@ -12,7 +12,6 @@ import com.gravatar.quickeditor.data.repository.AvatarRepository
 import com.gravatar.quickeditor.ui.editor.AvatarPickerContentLayout
 import com.gravatar.quickeditor.ui.editor.GravatarQuickEditorParams
 import com.gravatar.restapi.models.Avatar
-import com.gravatar.restapi.models.Profile
 import com.gravatar.services.ErrorType
 import com.gravatar.services.GravatarResult
 import com.gravatar.services.ProfileService
@@ -103,7 +102,6 @@ internal class AvatarPickerViewModel(
                             currentState.copy(
                                 emailAvatars = currentState.emailAvatars?.copy(selectedAvatarId = avatarId),
                                 selectingAvatarId = null,
-                                profile = currentState.profile?.copy { copyAvatar(avatar) },
                             )
                         }
                         _actions.send(AvatarPickerAction.AvatarSelected(avatar))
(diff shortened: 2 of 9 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -146,10 +147,13 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onEvent: (AvatarPickerEv
                     .fillMaxWidth()
                     .padding(bottom = 10.dp),
             )
-            ProfileCard(
-                profile = uiState.profile,
-                modifier = Modifier.padding(horizontal = 16.dp),
-            )
+            key(uiState.emailAvatars?.selectedAvatarId) {
```
AdamGrzybkowski, line 150: This one is new for me. What exactly is its usage here? Is it to only recompose when `selectedAvatarId` is changed? 

Won't that prevent updates when the `uiState.profile` is modified?

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
```diff
@@ -27,15 +31,23 @@ import com.gravatar.ui.skeletonEffect
...
     avatarQueryOptions: AvatarQueryOptions? = null,
+    forceRefresh: Boolean = false,
 ) {
     val sizePx = with(LocalDensity.current) { size.roundToPx() }
+    var cacheBuster by rememberSaveable { mutableStateOf<String?>(null) }
+
+    if (forceRefresh) {
+        cacheBuster = System.currentTimeMillis().toString()
+    }
+
```
AdamGrzybkowski, line 50: When `forceRefresh` is true the `cacheBuster` will be updated on each recomposition. What's the advantage of using `rememberSaveable` for it?

AdamGrzybkowski (commented):
The current version is without the delay, right? Where would that delay be implemented?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
hamorillo: > Is it to only recompose when selectedAvatarId is changed?

Yes. We know when the avatar is updated, but the Composable, which shows the avatar, doesn't depend on that state (the profile will be the same). So, using the key will force its recomposition based on the `selectedAvatarId` changes.

> Won't that prevent updates when the uiState.profile is modified

As far as I know, `ProfileCard` should still be updated if `uiState.profile` is modified.

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
hamorillo: At somepoint `forceRefresh` can be switched to `false` and, for example, rotate the device. The idea is to still use the previous cacheBuster.

hamorillo (commented):
> The current version is without the delay, right? 

Yes.

> Where would that delay be implemented?

That's a good question. I don't like the delay approach, but I couldn't come up with a better solution. I was thinking of adding the deley in the viewmodel, something like:

```kotlin
is Result.Success -> {
        delay(X ms)
        _uiState.update { currentState ->
```

But as I'm not a fan of this, I wanted to check with you if you have a better idea.

AdamGrzybkowski:
>However, the avatar is not always updated as expected because we request the new avatar too fast. We'll need to introduce a delay to postpone the recomposition a bit. You can reproduce it by switching between already uploaded avatars. Wdyt?

@pinarol Have you had a similar problem with the cache buster param not being enough to show the latest selected image? How did you handle that on iOS?

pinarol:
> @pinarol Have you had a similar problem with the cache buster param not being enough to show the latest selected image? How did you handle that on iOS?

When the user selects a different avatar, we directly use the recently selected image's URL to show it, so we don't use cache busting here. But this is also news to me if cache busting wasn't enough. 🤔

hamorillo:
I've added the delay before notifying the avatar switch.

I've also created #354 to remove the URL from the callback.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -34,6 +34,10 @@ internal class AvatarPickerViewModel(
     private val avatarRepository: AvatarRepository,
     private val fileUtils: FileUtils,
 ) : ViewModel() {
+    private companion object {
```
hamorillo, line 37: With less than 800 ms, I still see frequent issues.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -34,6 +34,10 @@ internal class AvatarPickerViewModel(
...
+        const val AVATAR_SWITCH_DELAY = 800L
```
hamorillo, line 38: With less than 800 ms, I still see frequent issues.

pinarol: Thanks for testing 👍 
cc: @etoledom

AdamGrzybkowski (approved):

PR #353, opened 2024-10-04
Add sdk headers to all Gravatar REST API requests

### LINKED ISSUE #349: Add HTTP headers to network requests
X-Platform: Android
X-SDK-Version: {sdk_version}

AdamGrzybkowski:
Closes #349

### Description

To track the adoption of the SDK we want to be able to filter API request by those from the SDK. 

This PR adds the source `Android` and the version of the SDK.

I didn't want to hardcode the version and require us to update it manually so I've included this via `BuildConfig`. That means I had to enable this class to be generated and it's public so it's exposed. What do you think about this? Can we make it internal somehow? 

### Testing Steps
Run the demo app and inspect request headers. There should be two extra: `X-Platform: Android` and  `X-SDK-Version: 1.1.0`

### DIFF
--- a/gravatar/build.gradle.kts
+++ b/gravatar/build.gradle.kts
@@ -13,6 +13,10 @@ plugins {
     alias(libs.plugins.openapi.generator)
 }
 
+val sdkVersion = providers.exec {
+    commandLine("git", "describe", "--tags", "--abbrev=0")
+}.standardOutput.asText.get().trim()
+
 android {
     namespace = "com.gravatar"
     compileSdk = 34
@@ -22,7 +26,7 @@ android {
         minSdk = 21
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles("consumer-rules.pro")
-        buildConfigField("String", "SDK_VERSION", "\"1.1.0\"")
+        buildConfigField("String", "SDK_VERSION", "\"$sdkVersion\"")
     }
 
     buildTypes {
--- a/gravatar/build.gradle.kts
+++ b/gravatar/build.gradle.kts
@@ -16,11 +16,13 @@ plugins {
 android {
     namespace = "com.gravatar"
     compileSdk = 34
+    buildFeatures.buildConfig = true
 
     defaultConfig {
         minSdk = 21
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles("consumer-rules.pro")
+        buildConfigField("String", "SDK_VERSION", "\"1.1.0\"")
     }
 
     buildTypes {
--- a/gravatar/src/main/java/com/gravatar/services/interceptors/SdkVersionInterceptor.kt
+++ b/gravatar/src/main/java/com/gravatar/services/interceptors/SdkVersionInterceptor.kt
@@ -1,0 +1,16 @@
+package com.gravatar.services.interceptors
+
+import com.gravatar.BuildConfig
+import okhttp3.Interceptor
+import okhttp3.Response
+
+internal class SdkVersionInterceptor : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        return chain.proceed(
+            chain.request().newBuilder()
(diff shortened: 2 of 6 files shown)


AdamGrzybkowski:
@wzieba Can you take a look at this PR, especifically at the modifications to build.gradle files?

gravatar/build.gradle.kts
```diff
@@ -13,14 +13,20 @@ plugins {
     alias(libs.plugins.openapi.generator)
 }
 
+val sdkVersion = providers.exec {
+    commandLine("git", "describe", "--tags", "--abbrev=0")
+}.standardOutput.asText.get().trim()
```
wzieba, line 18: This part looks good to me, aligned with Gradle docs on configuration cache compatibility. ([docs](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:external_processes)) 👍 

This `git` operation will run on every build invocation to check if cache is up-to-date, but this is very fast operation, so I don't expect any issues.

wzieba (approved):

gravatar/build.gradle.kts
AdamGrzybkowski: Thank you!

gravatar/src/main/java/com/gravatar/services/interceptors/SdkVersionInterceptor.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import com.gravatar.BuildConfig
+import okhttp3.Interceptor
+import okhttp3.Response
+
+internal class SdkVersionInterceptor : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        return chain.proceed(
+            chain.request().newBuilder()
+                .addHeader("X-Platform", "Android")
```
hamorillo, line 11: ❓ 💡 Can it make sense to diffenciate between release and debug versions?

hamorillo (approved):
Works as described. 🚀

💡- Could adding those headers when requesting an avatar make sense? That will allow us to know how many avatar requests we have from the SDKs.

> That means I had to enable this class to be generated and it's public so it's exposed. What do you think about this?

I don't think it's a big deal.

> Can we make it internal somehow?

I don't think we can make `Build.Config` internal. We can create an internal class that we use as `Config` and set values there dynamically with a gradlew task, but I don't think the work and complexity are worth it.

AdamGrzybkowski:
>💡- Could adding those headers when requesting an avatar make sense? That will allow us to know how many avatar requests we have from the SDKs.


Do you mean when loading an avatar via `https://gravatar.com/avatar/HASH `? We could set those via ImageRequest.

gravatar/src/main/java/com/gravatar/services/interceptors/SdkVersionInterceptor.kt
AdamGrzybkowski: hmm good question. I will ask iOS folks about that. 

I assume you want to be able to filter "our" usage of the SDK, right? Another solution could be modifying how we read the last tag. Currently, it's `git describe --tags --abbrev=0` but if we got rid of the `--abbrev=0`. That way only the build from the 1.1.0 tag will have a clean version without the commit suffix.

hamorillo:
> Do you mean when loading an avatar via https://gravatar.com/avatar/HASH ? We could set those via ImageRequest.

Yes, that is what I was talking about.

gravatar/src/main/java/com/gravatar/services/interceptors/SdkVersionInterceptor.kt
hamorillo: > I assume you want to be able to filter "our" usage of the SDK, right?

Exactly.

> Currently, it's git describe --tags --abbrev=0 but if we got rid of the --abbrev=0. That way only the build from the 1.1.0 tag will have a clean version without the commit suffix.

TIL - Sounds good!

AdamGrzybkowski:
@hamorillo I will merge it for now as is - we can update it later when we agree on something.

PR #355, opened 2024-10-04
QuickEditor: Remove the URL from the selected avatar callback

### LINKED ISSUE #354: QuickEditor: Remove avatar URL from selected avatar callback

hamorillo:
Closes #354 

### Description

The idea is to use the avatar URL with the cache buster to load the most recent avatar. Since that URL is always the same, we don't want to include the `/user image` in the callback anymore.

### Testing Steps

- Smoke test the AvatarPicker in the demo-app. Verify the avatar in the demo-app is also updated after switching the Avatar.


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
```diff
@@ -71,7 +71,7 @@ public fun GravatarQuickEditorBottomSheet(
 internal fun GravatarQuickEditorBottomSheet(
     gravatarQuickEditorParams: GravatarQuickEditorParams,
     authenticationMethod: AuthenticationMethod,
-    onAvatarSelected: (AvatarUpdateResult) -> Unit,
+    onAvatarSelected: () -> Unit,
```
AdamGrzybkowski, line 73: Shouldn't the `AvatarUpdateResult` be removed as well? And we need to update the docs for this callback in all the places mentioning about loading via hash with cache buster.

hamorillo: Good catch! I've removed `AvatarUpdateResult` and added a small section about cache busting in the `Get Started`.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/GravatarQuickEditor.kt
```diff
@@ -44,7 +43,7 @@ public object GravatarQuickEditor {
      * @param fragment The fragment to launch the Gravatar Quick Editor from.
      * @param gravatarQuickEditorParams The parameters to configure the Quick Editor.
      * @param authenticationMethod The method used for authentication with the Gravatar REST API.
-     * @param onAvatarSelected The callback for the avatar update result, check [AvatarUpdateResult].
+     * @param onAvatarSelected The callback for the avatar update result.
```
AdamGrzybkowski: >* @param onAvatarSelected The callback for the avatar update.

Should we drop the `result` here as well?

AdamGrzybkowski (approved):
Just a minor comment about the docs. Other than that LGTM 👍

PR #358, opened 2024-10-07
QuickEditor: Preventing Avatar in ProfileCard to recompose twice when QE is opened

### LINKED ISSUE #357: Quick Editor: The avatar can be loaded twice when the QE bottom sheet is opened.
Introduce in #350

hamorillo:
Closes #357 

### Description

ProfileCard recomposition depended on both `uiState.emailAvatars?.selectedAvatarId` and `uiState.profile`. Therefore, when the bottom sheet was opened, we could load the avatar twice, creating a visual blink.

Depending on a different ui state that it's only updated when the avatar is updated should solve this issue.

### Testing Steps

1. Open the QE bottom sheet
2. Verify the Avatar in the ProfileCard is only loaded once (you don't see a blink)
3. Switch the selected image
4. Verify the Avatar is updated


AdamGrzybkowski (approved):
This fixes the issue 👍 

I do have mixed feelings about how we have to work around to show the latest avatar. Especially in Compose, where the UI is driven by the state, but because the Avatar URL is fixed now, we have to make those extra properties to fake a state update. I don't have a better idea for now though.

PR #359, opened 2024-10-07
Proguard rules for :gravatar module

### LINKED ISSUE #345: Test the SDK against code obfuscation
Add rules to the SDK if necessary.

AdamGrzybkowski:
Closes #345

### Description

This PR provides pro guard rules that will be shipped with our SDK so that our code won't crash if `isMinifyEnabled` in third-party app is enabled.

### Testing Steps

Run the demo app with the release build type and try to test each tab. Ideally, test `trunk` first to confirm there was an issue (crash).

### DIFF
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -68,7 +68,9 @@ android {
 
     buildTypes {
         release {
-            isMinifyEnabled = false
+            isMinifyEnabled = true
+            isShrinkResources = true
+            signingConfig = signingConfigs.findByName("debug")
             proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
         }
     }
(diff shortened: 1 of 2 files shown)


demo-app/build.gradle.kts
```diff
@@ -68,7 +68,9 @@ android {
 
     buildTypes {
         release {
-            isMinifyEnabled = false
+            isMinifyEnabled = true
+            isShrinkResources = true
+            signingConfig = signingConfigs.findByName("debug")
```
AdamGrzybkowski, line 73: Not sure we want it like that, but it's nice for testing and we don't use it for anything else right now.

hamorillo (approved):
Works as described! 🚀

💡Not directly related with this PR. For the demo-app prototype we are using the `debug` buildType. Can it make sense to use the `release` one in order to test the library with the proguard rules?

AdamGrzybkowski:
>Not directly related with this PR. For the demo-app prototype we are using the debug buildType. Can it make sense to use the release one in order to test the library with the proguard rules?

Yeah we should do it. https://github.com/Automattic/Gravatar-SDK-Android/issues/361

PR #368, opened 2024-10-08
Add X-Source header to Gravatar REST API request

### LINKED ISSUE #360: Add "X-source=app name" tag to API requests
We can get the `app name` from the application context. 

We can either get it via [App Startup](https://developer.android.com/topic/libraries/app-startup) library or ask for it directly when initializing the Context. 

Since it's not critical I think we should go with the second option and make it optional (as in not required to be passed). When available we can then get the app name and add it to the header.

hamorillo:
Closes #360 

### Description

This is a follow-up of #353. 

Having the request's source can help understand how developers use the SDK and their needs.

If the app provides the SDK with the context through the `Gravatar` object, we'll use that context to extract the application's name and send it in the `X-Source` header. 

Currently, we only use the `Context` for that purpose; it's optional to provide it. When it's not provided, we won't send the header.

### Testing Steps

1. Navigate to the QE tab in the demo-app
2. Switch through different avatars (Any QE request should work).
3. With the `Network Inspector`, verify you can see the header
image

4. In the demo-app code, remove the context from the initialization and run the demo-app
```
// Initialize the Gravatar SDK with the API key if it is available
BuildConfig.DEMO_GRAVATAR_API_KEY?.let { Gravatar.apiKey(it) }
``` 
5. Navigate to the QE tab in the demo-app
6. Switch through different avatars
7. With the `Network Inspector`, verify you CAN'T see the header
8. Everything else works as expected.

### DIFF
--- a/demo-app/src/main/java/com/gravatar/demoapp/MainActivity.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/MainActivity.kt
@@ -13,7 +13,7 @@ class MainActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
 
         // Initialize the Gravatar SDK with the API key if it is available
-        BuildConfig.DEMO_GRAVATAR_API_KEY?.let { Gravatar.apiKey(it).context(this) }
+        BuildConfig.DEMO_GRAVATAR_API_KEY?.let { Gravatar.apiKey(it).context(applicationContext) }
 
         setContent {
             DemoGravatarApp()
--- a/demo-app/src/main/java/com/gravatar/demoapp/MainActivity.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/MainActivity.kt
@@ -13,7 +13,7 @@ class MainActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
 
         // Initialize the Gravatar SDK with the API key if it is available
-        BuildConfig.DEMO_GRAVATAR_API_KEY?.let { Gravatar.initialize(it) }
+        BuildConfig.DEMO_GRAVATAR_API_KEY?.let { Gravatar.apiKey(it).context(this) }
 
         setContent {
             DemoGravatarApp()
--- a/docs/get-started/get-started.md
+++ b/docs/get-started/get-started.md
@@ -63,7 +63,7 @@ android {
 Then you can access the API key in your app's code like this:
 
 ```kotlin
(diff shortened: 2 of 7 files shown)


demo-app/src/main/java/com/gravatar/demoapp/MainActivity.kt
```diff
@@ -13,7 +13,7 @@ class MainActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
 
         // Initialize the Gravatar SDK with the API key if it is available
-        BuildConfig.DEMO_GRAVATAR_API_KEY?.let { Gravatar.initialize(it) }
+        BuildConfig.DEMO_GRAVATAR_API_KEY?.let { Gravatar.apiKey(it).context(this) }
```
AdamGrzybkowski:
```suggestion
        BuildConfig.DEMO_GRAVATAR_API_KEY?.let { Gravatar.apiKey(it).context(applicationContext) }
```

I guess it works, but I think this should be the application context, no?

docs/get-started/get-started.md
```diff
@@ -63,7 +63,7 @@ android {
 Then you can access the API key in your app's code like this:
 
 ```kotlin
-Gravatar.initialize(BuildConfig.GRAVATAR_API_KEY)
+Gravatar.apiKey(BuildConfig.GRAVATAR_API_KEY).context(appContext)
```
AdamGrzybkowski:
```suggestion
Gravatar.apiKey(BuildConfig.GRAVATAR_API_KEY)
    .context(appContext) // Optional
```

Should we add a comment that the `context` is optional?

AdamGrzybkowski (approved):
Works as described 👍 Left two minor comments.

docs/get-started/get-started.md
hamorillo: I would say it should work with any context. But let's change it to prevent issues.

demo-app/src/main/java/com/gravatar/demoapp/MainActivity.kt
hamorillo: I would say it should work with any context. But let's change it to prevent issues.

docs/get-started/get-started.md
hamorillo: 🤔 What about: 

```
// Optional but highly encouraged.
``` 

As it provides valuable information for us.

AdamGrzybkowski: 👍

PR #369, opened 2024-10-08
Release/2.0.0 rc1

AdamGrzybkowski:
### Description

2.0.0 release candidate 🎉 

### Testing Steps

Smoke test won't hurt?


AdamGrzybkowski:
@wzieba The "submit dependencies" step from Github Actions fails [here](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/gravatar/build.gradle.kts#L18). Do you have any idea why is this happening?

build.gradle.kts
```diff
@@ -54,7 +54,7 @@ apiValidation {
 }
 
 // Semantic versioning for release version
-val versionName = "1.1.0"
+val versionName = "2.0.0-rc1"
```
wzieba, line 57: Do we need this if we use git to check tag anyway?

wzieba:
I see - it's because by default, GitHub actions checkout action doesn't fetch tags and has depth of 1. I propose fix here: in #370, please merge this to this branch or trunk and rebase this branch - everything should work fine then. Or we can merge this if you prefer, as Submit dependencies failure shouldn't block release.

---

Please, consider using a separate branch for storing generated documentation. The current diff is not really readable, and if I fetch the repository, I don't want to fetch the complete, generated documentation in HTML.

wzieba:
Actually, I can't even comment on the PR because Google Chrome is crashing every time 😓 . My question for the PR - do we need to update `versionName` if we use `git tag` to determine version anyway?

build.gradle.kts
AdamGrzybkowski: This is for docs generation. The tag is created after this PR.
I haven't created this flow but I can look into it some time.

hamorillo (approved):
🚀

hamorillo:
> Please, consider using a separate branch for storing generated documentation. The current diff is not really readable, and if I fetch the repository, I don't want to fetch the complete, generated documentation in HTML.

I agree. We need to rethink how to manage the documentation. I've created a ticket #371

PR #372, opened 2024-10-09
Don't remember cacheBuster on Avatar

### LINKED ISSUE #362: Avatar in the ProfileCard is gone after configuration change
Steps:

1. Open QE
2. Go to system settings and change the language
3. Go back to QE

https://github.com/user-attachments/assets/1a0cedb7-bdc4-4230-b265-b1d6ca136141

hamorillo:
Closes #362

### Description

We were saving the `cacheBuster` value, including between configuration changes, which produced an infinite loop of recompositions. After thinking about this, when forceRefresh is `true`, we'll update the cache buster in every recomposition, so it doesn't make sense to remember the `cacheBuster`.

Also, when forceRefresh is `false`, as a developer, I would expect the component to use the default avatar URL for the given profile.

### Testing Steps

1. Open the QE
2. Change the device orientation (or follow the steps described [here](https://github.com/Automattic/Gravatar-SDK-Android/issues/362#issue-2572523808))
3. Verify the avatar is visible (you can also verify with the Network Inspector that is requested just one time)

### DIFF
--- a/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
@@ -4,10 +4,6 @@ import android.content.res.Configuration
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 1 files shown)


AdamGrzybkowski (approved):
:shipit:

PR #381, opened 2024-10-10
QuickEditor: Refreshing all info when no avatar selected and a new one is uploaded

### LINKED ISSUE #380: Quick Editor: The selected avatar isn't shown after a new upload when there wasn't a previously selected one.
If you don't have an avatar already selected, the backend automatically selects that image as the default avatar when a new avatar is uploaded. We need to reflect that in the QE.

hamorillo:
Closes #380 

### Description

When a new avatar is uploaded into a Gravatar account with no avatar selected, the backend automatically uses that image as the default avatar.

We need to reflect that in the QE UI. We can hardcode the changes in the code, modifying the uiState as required. However, if the backend behavior changes at some point, all previous versions of the QE will show the wrong status of the Gravatar account.

To mitigate that situation, we will refresh all data from the backend if an image is uploaded and no previous avatar is selected.

What do you think about this approach? Do you think it's worth it to refresh everything or assume the risk and modify the uiState locally, expecting the backend to do the same?

**Note:** I'll add some unit tests before merging if we decide to go this way.

### Testing Steps

1. Using an account with no avatar selected
2. Upload a new image using the QE
3. Verify you see, after reloading, the new image as the selected image.

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
@@ -146,18 +146,26 @@ internal class AvatarPickerViewModel(
             when (val result = avatarRepository.uploadAvatar(email, uri)) {
                 is GravatarResult.Success -> {
                     fileUtils.deleteFile(uri)
-                    _uiState.update { currentState ->
-                        val avatar = result.value
-                        currentState.copy(
-                            uploadingAvatar = null,
-                            emailAvatars = currentState.emailAvatars?.copy(
-                                avatars = buildList {
-                                    add(avatar)
-                                    addAll(currentState.emailAvatars.avatars.filter { it.imageId != avatar.imageId })
-                                },
-                            ),
-                            scrollToIndex = null,
-                        )
+                    val isAutoSelected = _uiState.value.emailAvatars?.selectedAvatarId == null
+                    if (isAutoSelected) {
+                        _uiState.update { AvatarPickerUiState(email, avatarPickerContentLayout) }
+                        refresh()
+                    } else {
+                        _uiState.update { currentState ->
+                            val avatar = result.value
+                            currentState.copy(
+                                uploadingAvatar = null,
+                                emailAvatars = currentState.emailAvatars?.copy(
+                                    avatars = buildList {
+                                        add(avatar)
+                                        addAll(
+                                            currentState.emailAvatars.avatars.filter { it.imageId != avatar.imageId },
+                                        )
+                                    },
+                                ),
+                                scrollToIndex = null,
+                            )
+                        }
                     }
                 }
 
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
@@ -155,6 +155,9 @@ internal class AvatarPickerViewModel(
                                 avatarUpdates = currentState.avatarUpdates.inc(),
                             )
                         }
+                        if (_uiState.value.emailAvatars?.selectedAvatarId != null) {
+                            _actions.send(AvatarPickerAction.AvatarSelected)
+                        }
                     } else {
                         _uiState.update { currentState ->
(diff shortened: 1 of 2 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -146,18 +146,26 @@ internal class AvatarPickerViewModel(
...
-                                    add(avatar)
-                                    addAll(currentState.emailAvatars.avatars.filter { it.imageId != avatar.imageId })
-                                },
-                            ),
-                            scrollToIndex = null,
-                        )
+                    val isAutoSelected = _uiState.value.emailAvatars?.selectedAvatarId == null
+                    if (isAutoSelected) {
+                        _uiState.update { AvatarPickerUiState(email, avatarPickerContentLayout) }
+                        refresh()
```
AdamGrzybkowski: I think it's completely reasonable to refresh the content like you said, to future proof in case of backend changes, but I wonder if we really have to clean the whole state?

Could we do this refresh action as part of the uploading indicator? With the current code, we hide the whole list, the profile card is recomposed fully and that seems like a lot to select an avatar.

hamorillo: > Could we do this refresh action as part of the uploading indicator? With the current code, we hide the whole list, the profile card is recomposed fully and that seems like a lot to select an avatar.

I don't think it's easy to keep the uploading indicator until the `fetchAvatars` finishes, at least reusing the current methods, but I'll try to improve it.

AdamGrzybkowski: Correct, you have to use suspend functions that won't start a new coroutine. We do have one like that to [fetch avatars](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt#L209) but not for the profile.

hamorillo: Done!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -217,9 +230,9 @@ internal class AvatarPickerViewModel(
                     currentState.copy(
                         emailAvatars = emailAvatars,
                         scrollToIndex = if (scrollToSelected && emailAvatars.avatars.isNotEmpty()) {
-                            emailAvatars.avatars.indexOfFirst {
+                            emailAvatars.avatars.indexOfFirstOrNull {
```
hamorillo, line 236: With the `coerceAtLeast(0)`, when there wasn't an avatar selected, we still updated the state with 0, so when we uploaded a new image and set it to 0 again, the `scrollToIndex` there wasn't a change in the state, and nothing happens. I think this is more correct and solves the issue.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -146,18 +146,31 @@ internal class AvatarPickerViewModel(
...
+                    val isAutoSelected = _uiState.value.emailAvatars?.selectedAvatarId == null
+                    if (isAutoSelected) {
+                        fetchAvatars(showLoading = false, scrollToSelected = true)
+                        _uiState.update { currentState ->
+                            currentState.copy(
+                                uploadingAvatar = null,
+                                avatarUpdates = currentState.avatarUpdates.inc(),
+                            )
+                        }
+                    } else {
```
AdamGrzybkowski: Shouldn't we emit the `AvatarPickerAction.AvatarSelected` here?

hamorillo: Good catch! Done!

AdamGrzybkowski (approved):
:shipit:

PR #383, opened 2024-10-10
Implement a custom version of CenterAlignedTopBar

### LINKED ISSUE #379: Replace CenterAlignedTopBar

hamorillo:
Closes #379 

### Description

We want to stop relying on experimental Material3 components. We use two: `ModalBottomSheet` and `CenterAlignedTopBar.` This PR removes the second one and replaces it with a custom `@Composable` that we can easily switch to the original one when the experimental annotation is removed (if they preserve the definition).

Comparison with the previous version:

com gravatar quickeditor ui components QETopBarTest qrTopBarDark_compare
com gravatar quickeditor ui components QETopBarTest qrTopBarLight_compare

### Testing Steps

Smoke test the QE including the XML activity. There should be no substantial visual differences.


AdamGrzybkowski (approved):
:shipit:

PR #384, opened 2024-10-11
Remove backup rules from quickeditor manifest

### LINKED ISSUE #374: QuickEditor: Remove backup rules from the Manifest
Some apps that already have the token don't need anything from that, yet they might end up with Manifest merge conflicts because of that. 
We shouldn't add those backup rules to the Manifest and just mention them in the docs.

AdamGrzybkowski:
Closes #374 

### Description

Manifest merger failed when I tried to add QuickEditor in WP as they had their own backup rules applied. I'm not sure why the Merger ignores those rules in the Demo App (we also have them applied) but since this  optional I think it's okey to remove them from the manifest.

### Testing Steps

CI should be fine.

### DIFF
--- a/docs/get-started/get-started.md
+++ b/docs/get-started/get-started.md
@@ -314,12 +314,12 @@ GravatarQuickEditor.logout(Email("{USER_EMAIL}"))
 #### Exclude Data Store files from Android backup (optional, but recommended)
 
 Data Store files are subject to Android backups. Encrypted files from the backup won't work when restored on a different device so we have to exclude those files.
-It is encouraged to add the below rules to your own files.
+It is encouraged to create those files or copy paste below rules to your own respective files.
 
 <details>
   <summary>Instructions</summary>
 
-In `AndroidManifest.xml` add those lines:
+In `AndroidManifest.xml` add the below lines. If you already have them, you can skip this step.
 
 ```xml
 <application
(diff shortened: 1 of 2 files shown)


docs/get-started/get-started.md
```diff
@@ -314,7 +314,7 @@ GravatarQuickEditor.logout(Email("{USER_EMAIL}"))
 #### Exclude Data Store files from Android backup (optional, but recommended)
 
 Data Store files are subject to Android backups. Encrypted files from the backup won't work when restored on a different device so we have to exclude those files.
-If your app has backup rules configured, those that are provided in the SDK won't be used so you have to copy them to your files.
+It is encouraged to add the below rules to your own files.
```
hamorillo: Third-party developers must copy/paste our rules into their files, right? Do you think we should make it more specific?

AdamGrzybkowski: Yes, correct. Do you want to use the `copy/paste` wording instead of `add`?

hamorillo: It's not a strong opinion. I've mentioned that because, at first glance, you can understand that adding the line in the manifest is enough, but maybe it's just me making it clear enough. So, if you think it is okay, feel free to merge it.

hamorillo (approved):

docs/get-started/get-started.md
AdamGrzybkowski: Updated!

PR #385, opened 2024-10-11
Add @JvmOverloads to AvatarUrl.url() function

### LINKED ISSUE #375: Add @JvmOverload to AvatarUrl.url()
Here https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/gravatar/src/main/java/com/gravatar/AvatarUrl.kt#L109

AdamGrzybkowski:
Closes #375 

### Description

Improving Java interop. `cacheBuster` param was required in Java, and the `@JvmOverloads` annotation will create a separate `url()` function without any arguments and use a default null value.

### Testing Steps

One way to test it is to create a Java file and try to use it - That's what I did.

### DIFF
--- a/gravatar/src/main/java/com/gravatar/AvatarUrl.kt
+++ b/gravatar/src/main/java/com/gravatar/AvatarUrl.kt
@@ -58,6 +58,7 @@ public class AvatarUrl {
      * @param hash Gravatar hash
      * @param avatarQueryOptions Avatar query options
      */
+    @JvmOverloads
     public constructor(hash: Hash, avatarQueryOptions: AvatarQueryOptions? = null) {
         this.hash = hash
         this.avatarQueryOptions = avatarQueryOptions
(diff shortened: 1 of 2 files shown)


gravatar/src/main/java/com/gravatar/AvatarUrl.kt
```diff
@@ -104,8 +104,10 @@ public class AvatarUrl {
     /**
      * Get the [URL] for the avatar.
      *
+     * @param cacheBuster Random string value to force a cache bust
```
hamorillo, line 110: I wonder why Detekt hasn't complained about that... 🤔

hamorillo (approved):
❓💡 Similar things happen with come constructors, for example:

```kotlin
﻿public constructor(
        email: Email,
        avatarQueryOptions: AvatarQueryOptions? = 
) : this(email.hash(), avatarQueryOptions)
```

We could create a ticket to review those cases. It should be a quick win for Java interop. Wdyt?

gravatar/src/main/java/com/gravatar/AvatarUrl.kt
AdamGrzybkowski: yeah good question

AdamGrzybkowski:
>We could create a ticket to review those cases. It should be a quick win for Java interop. Wdyt?

Good call, I will add those here.

AdamGrzybkowski:
@hamorillo done!

PR #386, opened 2024-10-11
Launch showSnackbar from a separate coroutineScope to not block collect()

### LINKED ISSUE #365: QuickEditor: `onAvatarSelected` callback seems to be delay by the Snackbar
When selecting an image for the second time while the previous confirmation snackbar is still visible it's possible to observe that the Avatar in the third-party app (demo app in this case) will be displayed only after the old snackbard is hidden. 

https://github.com/user-attachments/assets/6b10547c-b97d-421a-a16e-153560c2efd0

AdamGrzybkowski:
Closes #365

### Description

We used the scope from the `collect` function to show the snack bar. `showSnackbar` suspends until the snackbar disappears so we were blocking new events from being collected and this caused the bug where the `onAvatarSelected` would be invoked with a delay. 

### Testing Steps

Select multiple avatars while the snackbar is visible. The Avatar in the demo app should be refreshed without waiting for the snack to disappear.


hamorillo (approved):
🚀

PR #388, opened 2024-10-14
Downgrade Compose bom to 2024.02.00

AdamGrzybkowski:
### Description

Downgrading Compose to an older version to not force the SDK consumers to use the newest version. 

### Testing Steps

Smoke test the Demo app.


gradle/libs.versions.toml
```diff
@@ -21,7 +21,7 @@ activityCompose = "1.9.2"
 robolectric = "4.12.2"
 composeUiTest = "1.7.2"
 kotlinCoroutines = "1.7.3"
-composeBom = "2024.09.02"
+composeBom = "2024.02.02"
```
hamorillo: I think we should go with "2024.02.00" to use the `androidx.compose.material3:material3:1.2.0` instead of `1.2.1`

```suggestion
composeBom = "2024.02.00"
```

AdamGrzybkowski: Sure, done!

hamorillo (approved):
🚀

PR #389, opened 2024-10-15
Comment GH PRs with depenednecy diff

wzieba:
### Description

This PR introduces an integration of CI toolkit plugin, that will print any changes made to `releaseRuntimeClasspath` if PR changes any of Gradle files.

### Testing Steps

I've tested this by introducing 2089485 . The script printed following comment:

image


wzieba:
The comment above should be removed, but I realised we have a bug in script (https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/issues/129). Anyway, I don't think it's a blocking thing.

AdamGrzybkowski (approved):
:shipit:

PR #392, opened 2024-10-15
Throwing GravatarException on noncatching service methods

### LINKED ISSUE #377: Not catching service functions should throw Gravatar specific exceptions
There's a difference between catching and non-catching service functions. One returns a specific Gravatar `ErrorType` and the other throws an Exception without any mapping on our side. 

We should consider throwing our own `GravatarException` with `ErrorType`.

hamorillo:
Closes #377 

### Description

We want to wrap all exceptions that can occur inside the Gravatar service methods in our own `GravatarException` to provide the `ErrorType` as we do in the catching variants of those methods.

With this approach, we'll provide detailed information about the error, but at the same time, third-party developers know what exceptions can be expected from the Gravatar Services, basically `GravatarException`.

### Testing Steps

- Code review (including test)
- Smoke test demo-app (although we are not using `nonCatching` method there)

### DIFF
--- a/gravatar/build.gradle.kts
+++ b/gravatar/build.gradle.kts
@@ -84,6 +84,7 @@ dependencies {
     testImplementation(libs.mockk.android)
     testImplementation(libs.mockk.agent)
     testImplementation(libs.kotlinx.coroutines.test)
+    testImplementation(kotlin("test"))
 }
 
 project.afterEvaluate {
--- a/gravatar/src/main/java/com/gravatar/services/ErrorType.kt
+++ b/gravatar/src/main/java/com/gravatar/services/ErrorType.kt
@@ -20,7 +20,7 @@ internal fun HttpException.errorTypeFromHttpCode(moshi: Moshi): ErrorType = when
     }
 
     in HttpResponseCode.SERVER_ERRORS -> ErrorType.Server
-    else -> ErrorType.Unknown
+    else -> ErrorType.Unknown("HTTP Code $code - ErrorBody $rawErrorBody")
 }
 
 internal fun Throwable.errorType(moshi: Moshi): ErrorType {
@@ -28,7 +28,7 @@ internal fun Throwable.errorType(moshi: Moshi): ErrorType {
         is SocketTimeoutException -> ErrorType.Timeout
         is UnknownHostException -> ErrorType.Network
         is HttpException -> this.errorTypeFromHttpCode(moshi)
-        else -> ErrorType.Unknown
+        else -> ErrorType.Unknown(message)
     }
 }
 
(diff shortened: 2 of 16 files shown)


gravatar/src/main/java/com/gravatar/services/HttpException.kt
```diff
@@ -5,11 +5,10 @@ import retrofit2.Response
...
  *
+ * [rawErrorBody] The raw error body of the response.
  * [code] The HTTP status code.
- * [message] The HTTP status message.
  */
-public class HttpException internal constructor(response: Response<*>) : RuntimeException() {
-    internal val rawErrorBody: String? = response.errorBody()?.string()
-    public val code: Int = response.code()
-    public override val message: String = "HTTP ${response.code()} $rawErrorBody"
+internal class HttpException internal constructor(response: Response<*>) : RuntimeException() {
```
hamorillo: We'll use this exception only internally to catch HTTP errors.

gravatar/src/main/java/com/gravatar/services/ErrorType.kt
```diff
@@ -54,8 +54,18 @@ public sealed class ErrorType {
...
     public data object Unauthorized : ErrorType()
 
-    /** An unknown error occurred */
-    public data object Unknown : ErrorType()
+    /**
+     * An unknown error occurred
+     *
+     * @property errorMsg The error message, if available.
+     */
+    public class Unknown(public val errorMsg: String? = null) : ErrorType() {
```
hamorillo, line 62: I think it would be nice to provide some information when it is available.

gravatar/build.gradle.kts
```diff
@@ -84,6 +84,7 @@ dependencies {
     testImplementation(libs.mockk.android)
     testImplementation(libs.mockk.agent)
     testImplementation(libs.kotlinx.coroutines.test)
+    testImplementation(kotlin("test"))
```
hamorillo, line 87: Testing the inside of the expected exception with [JUnit 4 is not very straightforward](https://www.baeldung.com/junit-assert-exception#1-assert-exception-is-thrown), so adding this dependency here [makes our life easier.](https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-fails-with.html)

gravatar/src/main/java/com/gravatar/services/HttpException.kt
AdamGrzybkowski: We are throwing it and passing as `originalException`, no? Would it make sense to keep it public for that?

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -29,23 +30,29 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
...
+                        return@withContext null
+                    } else {
+                        throw HttpException(response)
+                    }
+                }
             }
+        } catch (ex: Exception) {
+            throw GravatarException(ex.errorType(GravatarSdkContainer.instance.moshi), ex)
         }
-    }
```
AdamGrzybkowski, line 48: Is there a reason why we can't use `runThrowingExceptionRequest` here?

AdamGrzybkowski (approved):
Demo App works fine - I've tried some network errors like 401 to make sure this still works. 

Left some minor comments.

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
hamorillo: In this case, the returned value when 404 is null instead of the exception, so we can't use `runThrowingExceptionRequest`, which requires not null.

gravatar/src/main/java/com/gravatar/services/HttpException.kt
hamorillo: Yeah, it makes sense. I'll revert the changes in this class.

AdamGrzybkowski: I think you can keep the constructor private - it's for internal use indeed.

hamorillo: I kept it as internal. 👍

PR #394, opened 2024-10-16
QuickEditor: Handle CAMERA permission

### LINKED ISSUE #393: Handle android.permission.CAMERA if present in the Manifest
When we want to take a new picture with a camera app we generally don’t need to ask for any permissions because the camera app should handle that. The problem is when a third-party app (Jetpack in this case) has 
```
<uses-permission android:name="android.permission.CAMERA" />
```
in AndroidManifest. This forces us to handle that permission even if we still ask the camera app to take the picture…

https://developer.android.com/reference/android/provider/MediaStore#ACTION_IMAGE_CAPTURE

AdamGrzybkowski:
Closes #393

### Description

It turns out we need to handle the Camera permission when a third-party app has `<uses-permission android:name="android.permission.CAMERA" />` defined in its `AndroidManifest.xml`. If we don't we will get `SecurityException` when launching the `ACTION_IMAGE_CAPTURE` intent - [source](https://developer.android.com/reference/android/provider/MediaStore#ACTION_IMAGE_CAPTURE).

**Permission granted**
https://github.com/user-attachments/assets/55dd6c8e-100d-4145-9d59-d88112d9efd8

**Permission denied**
https://github.com/user-attachments/assets/c5062aa4-25a7-47ca-9c36-dfe42f04b14c

**No permission in the third-party app**
https://github.com/user-attachments/assets/e742f451-c297-431a-98ff-7ddc9b7651ca

### Testing Steps

Camera permission is commented out in the Demo app manifest
1. Try to take a photo - all should work as usual.
2. Uncomment the Camera permission
3. Try to take a photo - you should see the permission dialog
4. Try different options - granting, denying, etc.

### DIFF
--- a/gravatar-quickeditor/src/main/res/values/strings.xml
+++ b/gravatar-quickeditor/src/main/res/values/strings.xml
@@ -34,4 +34,7 @@
     <string name="avatar_upload_failure_dialog_title">Couldn\'t upload image</string>
     <string name="avatar_upload_failure_dialog_remove_upload">Remove upload</string>
     <string name="oauth_email_associated_error_message">Something went wrong when verifying your email.</string>
+    <string name="gravatar_qe_permission_required_alert_title">Permission required</string>
+    <string name="gravatar_qe_camera_permission_rationale_message">To take a picture, you need to grant camera permission. You can grant it in the app settings.</string>
+    <string name="gravatar_qe_dismiss">Dismiss</string>
 </resources>
--- a/demo-app/src/main/AndroidManifest.xml
+++ b/demo-app/src/main/AndroidManifest.xml
@@ -3,6 +3,7 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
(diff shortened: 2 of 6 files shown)


gravatar-quickeditor/src/main/res/values/strings.xml
```diff
@@ -34,4 +34,7 @@
     <string name="avatar_upload_failure_dialog_title">Couldn\'t upload image</string>
     <string name="avatar_upload_failure_dialog_remove_upload">Remove upload</string>
     <string name="oauth_email_associated_error_message">Something went wrong when verifying your email.</string>
+    <string name="gravatar_qe_permission_required_alert_title">Permission required</string>
+    <string name="gravatar_qe_camera_permission_rationale_message">To take a picture, you need to grant camera permission. You can grant it in the app settings.</string>
+    <string name="gravatar_qe_dismiss">Dismiss</string>
```
AdamGrzybkowski, line 39: @mlumeau  We will have to wait for the translations.

hamorillo (approved):
Works as described! 🚀

PR #396, opened 2024-10-17
Make onDismiss optional in the GravatarQuickEditor.show()

AdamGrzybkowski:
### Description

Small public API update. The `onDismiss` callback can be ignored, as it's not critical to QE functioning, so I made it optional and added the `@JvmOverloads` for Java interop. 

I noticed we had an outdated doc comment for this callback so updated it as well.

### Testing Steps

Smoke test QE from the XML Activity class


hamorillo (approved):
:shipit:

PR #398, opened 2024-10-17
Create an empty GravatarUCropActivity to avoid Manifest merger fails

AdamGrzybkowski:
### Description

When a third-party app depends on UCrop already they likely have defined the `UCropActivity` in the manifest. We do that as well in the QE and this causes the Manifest merger to fail. 

It's possible to remove that declaration from a third-party app (or use `tools:replace`) and keep the one from QE, but it's not ideal because there could be custom fields set that could differ from the ones we use. 

To fix this I've added a "fake" UCrop activity to not have clashing names. This way we can define our Activity in the manifest without requiring third-party apps to remove theirs. 

### Testing Steps

I've added a potentially conflicting UCropActivity to the demo app, so the CI should be enough. 

I was also able to confirm this in the WP app with the built from this PR.

### DIFF
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -117,6 +117,7 @@ dependencies {
     implementation(libs.coil.compose)
     implementation(libs.androidx.activity)
     implementation(libs.androidx.constraintlayout)
+    implementation(libs.ucrop)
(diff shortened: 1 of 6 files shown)


hamorillo (approved):
Good catch! 🚀

PR #400, opened 2024-10-18
QuickEditor: Missing action on gravatar icon

### LINKED ISSUE #376: QuickEditor: Missing action on the Gravatar icon
When tapped we should show the user profile the same as the `View profile` button.

mlumeau:
Closes #376 

### Description

Adds a redirection to https://gravatar.com/profile in a ~~custom~~ browser tab when clicking on the bottomsheet's Gravatar icon in the top right corner.

edit: launches in a regular browser tab to align with the behavior of the "view profile" button.

### Testing Steps
- [ ] While logged out of Gravatar in your browser, click on the icon and see that you are redirected to the login page
- [ ] While logged in Gravatar in your browser, click on the icon and see that you are redirected to your profile edition page

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QETopBar.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QETopBar.kt
@@ -16,7 +16,6 @@ import androidx.compose.material3.TextButton
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.platform.LocalUriHandler
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.font.FontWeight
@@ -26,9 +25,7 @@ import com.gravatar.quickeditor.R
 import com.gravatar.ui.GravatarTheme
 
 @Composable
-internal fun QETopBar(onDoneClick: () -> Unit, gravatarIconUrl: String, modifier: Modifier = Modifier) {
-    val uriHandler = LocalUriHandler.current
-
+internal fun QETopBar(onDoneClick: () -> Unit, modifier: Modifier = Modifier, onGravatarIconClick: () -> Unit = {}) {
     GravatarCenterAlignedTopAppBar(
         modifier = modifier,
         title = {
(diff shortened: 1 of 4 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QETopBar.kt
```diff
@@ -24,7 +26,9 @@ import com.gravatar.quickeditor.R
 import com.gravatar.ui.GravatarTheme
 
 @Composable
-internal fun QETopBar(onDoneClick: () -> Unit, modifier: Modifier = Modifier) {
+internal fun QETopBar(onDoneClick: () -> Unit, gravatarIconUrl: String, modifier: Modifier = Modifier) {
```
AdamGrzybkowski:
```suggestion
internal fun QETopBar(onDoneClick: () -> Unit, onGravatarIconClick: () -> Unit = {}, modifier: Modifier = Modifier) {
```

I would expose the `onGravatarIconClick` rather than pass the URL to let the parent composable decide what behavior they want on icon tap. I think it's more flexible, WDYT?

AdamGrzybkowski (commented):
> Adds a redirection to https://gravatar.com/profile in a custom browser


I don't think the current code will use the custom browser tab. It will launch the browser.  Launcing the custom tab looks more or less like this:

```
val customTabIntent: CustomTabsIntent = CustomTabsIntent.Builder()
        .build()
customTabIntent.launchUrl(context, URL)
 ```
 
 I've also left a comment in the Issue on how to show the specific profile rather `/profile`. This could show the wrong profile if the user is logged in with a different email then used in the QE.

mlumeau:
> I don't think the current code will use the custom browser tab

You're right, I had that in mind but then decided to go with a regular tab in the code to do the same as the "view profile" button, forgot to change that in the PR description.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QETopBar.kt
mlumeau: Fixed in 187258d

hamorillo:
> I've also left a comment in the Issue on how to show the specific profile rather /profile. This could show the wrong profile if the user is logged in with a different email then used in the QE.

This is important. I think we'll need to "launch" the callback at least to `GravatarQuickEditorBottomSheet` and, from there, open the profile using the email from `gravatarQuickEditorParams`.

mlumeau:
> This is important. I think we'll need to "launch" the callback at least to GravatarQuickEditorBottomSheet and, from there, open the profile using the email from gravatarQuickEditorParams

In that case if the email is not associated to a gravatar account it will 404. That's what I did in my first commit. Also if the user is signed in a different email there will be an error message in the bottom sheet after oauth, so I think it's enough.

AdamGrzybkowski:
If you think that's a better URL to open I think it's best to discuss with iOS folks to make sure both platforms are aligned.

AdamGrzybkowski (approved):
:shipit:

mlumeau:
Permission to merge into the release branch? They're targeting this release on iOS.

Issue #401, opened 2024-10-18
Gravatar: Refactor ErrorType from a sealed class to an interface


hamorillo:
ErrorType is a sealed class, and problems could be created when adding a new ErrorType (Breaking change - It can break the compilation in a when clause).

PR #402, opened 2024-10-18
Add 413 HTTP error - CONTENT_TOO_LARGE

### LINKED ISSUE #352: Add 413 error to our SDK
The error shown to the final user should be:

“The provided image exceeds the maximum size: 10MB”

hamorillo:
Closes #352 

### Description

This PR aims to solve the same issue as https://github.com/Automattic/Gravatar-SDK-iOS/pull/467 in the iOS repo.

When an image that is too big is uploaded, the server will answer with a 413, which we should handle and show the proper error message. For that reason, we need to add a new ErrorType. 

**Note:** We've discussed ErrorType being a sealed class and the problems that could arise when adding a new ErrorType (Breaking change—it can break the compilation in a when clause). I'll open a different PR for that. #401

### Testing Steps

As we are compressing and limiting the maximum size of the final image, it's not easy to test the changes. You have two options:

**Modifying the code**

- Apply the following patch (or comment the compression/size options in `UCropCropperLauncher`)
[Gravatar-Android-12-47-55.patch](https://github.com/user-attachments/files/17434223/Gravatar-Android-12-47-55.patch)
- Build the demo-app
- Upload a huge image (you can use the [resources](https://drive.google.com/drive/u/0/folders/16bRuYugP99X3grDlh938ffJaomlfFXaT) from the iOS PR)

**Mock the server response**

- Using your favorite network tool (ex: Charles...) mock the response from the server when uploading an avatar - 413

**In both options ->** Verify you see the error on the avatar picker with the associated message.

image

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerUiState.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerUiState.kt
@@ -5,6 +5,7 @@ import com.gravatar.quickeditor.data.repository.EmailAvatars
 import com.gravatar.quickeditor.ui.editor.AvatarPickerContentLayout
 import com.gravatar.restapi.models.Avatar
 import com.gravatar.restapi.models.Profile
+import com.gravatar.services.ErrorType
 import com.gravatar.types.Email
 import com.gravatar.ui.components.ComponentState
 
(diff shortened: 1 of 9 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerUiState.kt
```diff
@@ -81,7 +82,7 @@ internal data class AvatarsSectionUiState(
 
 internal data class AvatarUploadFailure(
     val uri: Uri,
-    val error: String?,
+    val error: ErrorType?,
```
hamorillo, line 85: I've replaced this because it looks better if the view handles what string should be shown. With the `String` we would need to access `R.string` from the `ViewModel`.

AdamGrzybkowski (approved):
:shipit:

PR #403, opened 2024-10-18
Add a new ErrorType to handle possible future errors

### LINKED ISSUE #401: Gravatar: Refactor ErrorType from a sealed class to an interface

hamorillo:
Closes #401 

### Description

```kotlin
public abstract class FutureErrors : ErrorType()
```

That will force any developer to add that consideration in their current exhaustive when clauses.

During the `2.x.x` every new ErrorType that we want to add should inherit from `FutureErrors`. Any exhaustive code will still compile as they are already considering the parent class. If someone is interested in a new concrete error, they can add that branch to the when without problems while keeping the FutureErrors branch.

```kotlin
public abstract class FutureErrors : ErrorType()
public data object WTFError: FutureErrors()
```

When we release new major versions, we need to change the parent class for the new errors:

```
public data object WTFError: ErrorType()
public abstract class FutureErrors : ErrorType()
```

The exact process will be followed during the life of a major version.

The downside of this approach is that, as a third-party developer, the new type will force me to consider a `generic` error if I don't currently have one. 

Thoughts?

### Testing Steps

- Code Review

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
@@ -266,6 +266,7 @@ internal class AvatarPickerViewModel(
                 is ErrorType.Unknown,
                 is ErrorType.InvalidRequest,
                 ErrorType.ContentLengthExceeded,
+                is ErrorType.FutureError,
                 -> SectionError.Unknown
             }
         }
--- a/gravatar/api/gravatar.api
+++ b/gravatar/api/gravatar.api
@@ -661,6 +661,10 @@ public final class com/gravatar/services/ErrorType$ContentLengthExceeded : com/g
 	public fun toString ()Ljava/lang/String;
 }
 
+public abstract class com/gravatar/services/ErrorType$FutureError : com/gravatar/services/ErrorType {
+	public fun <init> ()V
+}
+
 public final class com/gravatar/services/ErrorType$InvalidRequest : com/gravatar/services/ErrorType {
 	public fun <init> (Lcom/gravatar/restapi/models/Error;)V
 	public fun equals (Ljava/lang/Object;)Z
--- a/gravatar/src/main/java/com/gravatar/services/ErrorType.kt
+++ b/gravatar/src/main/java/com/gravatar/services/ErrorType.kt
@@ -82,4 +82,10 @@ public sealed class ErrorType {
 
         override fun hashCode(): Int = Objects.hash(error)
     }
+
+    /**
+     * Future errors that may be added in the future will extend this class to ensure that all errors are handled.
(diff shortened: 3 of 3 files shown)


mlumeau:
Why not make this a non-abstract class called `GenericError` instead (or introduce this type alongside `FutureErrors`)? There are error types we'd probably want to remain treated generically.

hamorillo:
> Why not make this a non-abstract class called GenericError instead (or introduce this type alongside FutureErrors)? There are error types we'd probably want to remain treated generically.

We already have the `Unknown` type, which sounds very similar to the `GenericError, ' at least in practical terms. Also, `Unknown`/`Generic Error` has a different purpose; it lets developers know that all `FutureError` will be stable, and they need to handle them (if necessary) in the next major version.

AdamGrzybkowski:
I'm not going to lie I have mixed feelings about this solution. I don't like the fact that we are exposing abstract `FutureErrors` to the third-party devs only because we are not sure if there won't be any new `ErrorType` in the future. 

`sealed` class or interface has its pros but it seems it's not flexible enough for our use case. One of the drawbacks of using the simple `interface` is that third-party devs would have to add an `else ->` clause (or just ignore other errors), but this also is an advantage for us because we won't break the compilation by adding a new type. It's a disadvantage because there won't be any signal that a new `ErrorType` was added. 

I do think that in the majority of use cases, devs will handle only specific errors they are interested in, and do an `else ->` clause anyway, so using an interface won't be such a terrible thing for them. 

I don't want to block the above solution, so if we were to go with that I would try to figure out a different name. `FutureErorrs` sounds like there's nothing there for now to handle (which is true only until we add a new `ErrorType`). So we may end up with `FutureErrors` that are not future and they should/could be handled now.

hamorillo:
> I'm not going to lie I have mixed feelings about this solution. 

It's the same for me. That's why I opened the PR as a draft; I wanted to discuss whether it is worth it before opening it for review. It's the only way I've found to keep the benefits from the sealed class while being able to add new `ErrorType`s without a new major version. 

I'm not pushing harder with this approach because I'm not totally confident about it. 

> I do think that in the majority of use cases, devs will handle only specific errors they are interested in, and do an else -> clause anyway, so using an interface won't be such a terrible thing for them.

That can be true, and if this is true for the majority of cases, maybe we should take the `interface` approach. 

>  but this also is an advantage for us because we won't break the compilation by adding a new type.

I don't think it's an `advantage`. I mean, it is what we are trying to solve. The `else` branch would hide new errors or force you to have some extra unreachable code if you've caught all `ErrorType`s. Said that, I think it is something acceptable and probably easier to manage/understand than the `FutureError` approach. 

If we all agree on that, I'll open a PR with the interface approach to get it merged before `2.0.0`

AdamGrzybkowski:
>If we all agree on that, I'll open a PR with the interface approach to get it merged before 2.0.0

That's what I would vote for.

hamorillo:
Closing this in favor of #406

PR #406, opened 2024-10-21
Refactor ErrorType from `sealed class` to `interface`

### LINKED ISSUE #401: Gravatar: Refactor ErrorType from a sealed class to an interface

hamorillo:
Closes #401 

### Description

Adding a new ErrorType would cause a breaking change as we used a `sealed class`. Switching from the `sealed class` to an `interface` will remove the breaking change as the third-party developers will need to add an else branch into their `when` clauses. You can read more about this discussion in this PR #403.

### Testing Steps

- Smoke Test demo-app, force some errors in the QE


mlumeau (approved):
I agree that's a better and more flexible approach 👍

AdamGrzybkowski (approved):

PR #407, opened 2024-10-21
Disable the option to override QuickEditor theme

### LINKED ISSUE #332: Disable the option to override Theme in the Quick Editor

AdamGrzybkowski:
Closes #332 
### Description

We don't want the QuickEditor theme to be overridden. It should be always obvious this is an official part of Gravatar even when implemented within a third-party app.

### Testing Steps

The Demo app now tries to modify the theme, so just smoke-test it and make sure the surface color in the QE is not RED.

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
@@ -70,17 +70,15 @@ public fun GravatarQuickEditorBottomSheet(
     onAvatarSelected: () -> Unit,
     onDismiss: (dismissReason: GravatarQuickEditorDismissReason) -> Unit = {},
 ) {
-    CompositionLocalProvider(LocalGravatarTheme provides mainGravatarTheme) {
-        GravatarQuickEditorBottomSheet(
-            gravatarQuickEditorParams = gravatarQuickEditorParams,
-            authenticationMethod = authenticationMethod,
-            onAvatarSelected = onAvatarSelected,
-            onDismiss = onDismiss,
-            modalBottomSheetState = rememberGravatarModalBottomSheetState(
-                avatarPickerContentLayout = gravatarQuickEditorParams.avatarPickerContentLayout,
-            ),
-        )
-    }
+    GravatarQuickEditorBottomSheet(
+        gravatarQuickEditorParams = gravatarQuickEditorParams,
+        authenticationMethod = authenticationMethod,
+        onAvatarSelected = onAvatarSelected,
+        onDismiss = onDismiss,
+        modalBottomSheetState = rememberGravatarModalBottomSheetState(
+            avatarPickerContentLayout = gravatarQuickEditorParams.avatarPickerContentLayout,
+        ),
+    )
 }
 
 @Composable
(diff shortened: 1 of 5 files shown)


gravatar-ui/src/main/java/com/gravatar/ui/GravatarTheme.kt
```diff
@@ -52,17 +52,17 @@ public interface GravatarTheme {
         get() = MaterialTheme.shapes
 }
 
+public val mainGravatarTheme: GravatarTheme = object : GravatarTheme {
+    override val colorScheme: ColorScheme
+        @Composable
+        get() = if (isSystemInDarkTheme()) DarkColorScheme else LightColorScheme
+}
+
```
AdamGrzybkowski, line 63: I had to expose this, not sure there's a way to avoid that.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
```diff
@@ -67,15 +70,17 @@ public fun GravatarQuickEditorBottomSheet(
...
-    GravatarQuickEditorBottomSheet(
-        gravatarQuickEditorParams = gravatarQuickEditorParams,
-        authenticationMethod = authenticationMethod,
-        onAvatarSelected = onAvatarSelected,
-        onDismiss = onDismiss,
-        modalBottomSheetState = rememberGravatarModalBottomSheetState(
-            avatarPickerContentLayout = gravatarQuickEditorParams.avatarPickerContentLayout,
-        ),
-    )
+    CompositionLocalProvider(LocalGravatarTheme provides mainGravatarTheme) {
```
AdamGrzybkowski: The only idea I had was to provide the Gravatar theme here, so if there's one provided in a parent component it will be overridden anyway. 

Are there other ways to do it?

hamorillo: I don't think we have much more options.

gravatar-ui/src/main/java/com/gravatar/ui/GravatarTheme.kt
hamorillo: Exposing our Theme is not necessarily negative. It lets third-party developers modify in cases when, for example, they only want to change a concrete color. 

I think it doesn't hurt. 👍

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
hamorillo: ❓ I wonder if we should move this to the internal composable as it is also used by the `GravatarQuickEditorBottomSheetWrapper`. I mean, we are only applying the `CompositionLocalProvider` to the public composable but not to the one used by the old view system extensions. 

I've been trying to modify that in the test activity, and I couldn't, but it looks safer in the internal composable. Wdty?

AdamGrzybkowski: Oh yes, good catch!

hamorillo (approved):
🚀

PR #408, opened 2024-10-22
Use `gravatar_` prefix in all SDK resources (strings and drawables)

### LINKED ISSUE #329: Add `gravatar_` prefix to all resources

hamorillo:
Closes #329 

### Description

Adding `gravatar_` as a prefix in all our resources will prevent conflicts between apps and the SDK strings/drawables.

One thing I don't like is that, this PR is "creating" a bunch of strings, which are the same but are going to send to the translation service anyway. However,  I don't know how to prevent this. Fortunately, we don't have a lot of them at the moment...

### Testing Steps

- Smoke test demo-app


AdamGrzybkowski:
>One thing I don't like is that, this PR is "creating" a bunch of strings, which are the same but are going to send to the translation service anyway. However, I don't know how to prevent this. Fortunately, we don't have a lot of them at the moment...


Not much we can do about it. I'm not even sure if modifying those translations here makes sense. It would probably be better to manually update them one by one in Glotpress and then download them again if we don't want them to be sent to the translation team.

gravatar-quickeditor/src/main/res/values/strings.xml
```diff
@@ -1,40 +1,40 @@
...
+    <string name="gravatar_avatar_picker_network_error_title">No internet</string>
+    <string name="gravatar_avatar_picker_network_error_message">Uh oh! Looks like you\'re not connected to the internet.</string>
+    <string name="gravatar_login_required">Login required</string>
+    <string name="gravatar_login_required_message">To modify your Gravatar profile you need to log&#160;in first</string>
+    <string name="gravatar_oauth_wrong_email_authenticated_error_message">It looks like you used the wrong email to log in. Please try again using %s this time. Thanks!</string>
+    <string name="gravatar_happy_face_image_content_description">Happy face image</string>
+    <string name="gravatar_upload_image_content_description">Upload image</string>
+    <string name="gravatar_failed_to_load_avatar_content_description">Upload failed, tap to see actions.</string>
+    <string name="gravatar_avatar_upload_failure_dialog_title">Couldn\'t upload image</string>
+    <string name="gravatar_avatar_upload_failure_dialog_remove_upload">Remove upload</string>
```
AdamGrzybkowski: I've started using `gravatar_qe` for the quick editor module (you can see below). I thought it would be safer to use different prefixes per module. WDYT?

hamorillo:
> It would probably be better to manually update them one by one in Glotpress and then download them again if we don't want them to be sent to the translation team.

That was my first idea, but I haven't found a way to update the keys in GlotPress. Even if we can do it, I doubt our jobs won't detect them as a new string and send them anyway.

gravatar-quickeditor/src/main/res/values/strings.xml
hamorillo: Good idea. So `gravatar_qe` and `gravatar_ui` then?

AdamGrzybkowski:
>That was my first idea, but I haven't found a way to update the keys in GlotPress. Even if we can do it, I doubt our jobs won't detect them as a new string and send them anyway.


The keys should be updated once Glotpress pulls the strings from the `trunk` -> After that we could manually do the translations ourselves. Once we have all of them translated, only the untranslated ones will likely be sent to the team. It's just a theory though.

hamorillo:
> Once we have all of them translated, only the untranslated ones will likely be sent to the team. It's just a theory though.

Based on previous conversations, I thought the new keys would be sent anyway, but we can try that and see what happens. :)

gravatar-quickeditor/src/main/res/values/strings.xml
AdamGrzybkowski: Sounds good 👍

hamorillo: Done!

AdamGrzybkowski (approved):
:shipit:

PR #410, opened 2024-10-23
QuickEditor: Preventing queue the Snackbars when selecting an image

### LINKED ISSUE #364: QuickEditor: Don't queue the Snackbars when selecting an image
Snackbars are queued, so if the user quickly selects different images those will be shown at a later time (unless manually dismissed by the user).

We should be smarter about this and skip queuing the Snack if there's one visible already.

hamorillo:
Closes #364 

### Description

As described in issue #364, we are queueing snack bars when switching between avatars. If a previous snackbar is still visible, the next one is added to the queue and shown after the first one. 

We want to avoid that situation for both success and error snackbars. However, I think it makes sense to show a success snackbar after an error one, so we'll let snackbars to queue if they are from a different event.

<details>
<summary> Before </summary>

[Your text here...](https://github.com/user-attachments/assets/98b74234-bf94-4bf9-a178-c9a506d5a4d9)
</details>

<details>
<summary> After </summary>

https://github.com/user-attachments/assets/a8b0be91-9e7c-495c-8103-93377ee3a8ec
</details>

### Testing Steps

1. Open the QE with an account that has several avatars
2. Quickly switch between them
3. Verify snackbars are not queued
4. Turn off the device Internet connection
5. Quickly switch between them
6. Verify error snackbars are not queued

### DIFF
--- a/demo-app/src/main/java/com/gravatar/demoapp/MainActivity.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/MainActivity.kt
@@ -13,6 +13,7 @@ class MainActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
 
         // Initialize the Gravatar SDK with the API key if it is available
+        @Suppress("UNNECESSARY_SAFE_CALL")
         BuildConfig.DEMO_GRAVATAR_API_KEY?.let { Gravatar.apiKey(it).context(applicationContext) }
 
         setContent {
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
@@ -66,7 +66,6 @@ import com.yalantis.ucrop.UCrop
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.withContext
 import java.net.URI
 
@@ -86,8 +85,6 @@ internal fun AvatarPicker(
     val context = LocalContext.current
     val uiState by viewModel.uiState.collectAsState()
     val scope = rememberCoroutineScope()
-    val avatarSelectedSnackbarMutex by remember { mutableStateOf(Mutex()) }
-    val avatarSelectedFailureSnackbarMutex by remember { mutableStateOf(Mutex()) }
 
     val uCropLauncher = rememberLauncherForActivityResult(ActivityResultContracts.StartActivityForResult()) {
         it.data?.let { intentData ->
@@ -109,8 +106,6 @@ internal fun AvatarPicker(
                         context = context,
                         uCropLauncher = uCropLauncher,
                         scope = scope,
-                        avatarSelectedSnackbarMutex = avatarSelectedSnackbarMutex,
-                        avatarSelectedFailureSnackbarMutex = avatarSelectedFailureSnackbarMutex,
                     )
                 }
             }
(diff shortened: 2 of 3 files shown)


demo-app/src/main/java/com/gravatar/demoapp/MainActivity.kt
```diff
@@ -13,6 +13,7 @@ class MainActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
 
         // Initialize the Gravatar SDK with the API key if it is available
+        @Suppress("UNNECESSARY_SAFE_CALL")
```
hamorillo, line 16: This is not related to this PR, is just to avoid the warning message as we want to keep the safe call.

AdamGrzybkowski:
Hmm I'm confused with the `After` video - it seems like the queueing is still there - is that a correct recording?

hamorillo:
> Hmm I'm confused with the After video - it seems like the queueing is still there - is that a correct recording?

🤦 Nop, it wasn't the correct recording. Updated! Sorry!

AdamGrzybkowski:
I think there's a simpler and more generic solution. Instead of playing with Mutex and handling each snack separately we can modify the `SnackbarHostState.showQESnackbar` extension function to sth like this: 
```
internal suspend fun SnackbarHostState.showQESnackbar(
    message: String,
    actionLabel: String? = null,
    withDismissAction: Boolean = false,
    snackbarType: SnackbarType = SnackbarType.Info,
    duration: SnackbarDuration =
        if (actionLabel == null) SnackbarDuration.Short else SnackbarDuration.Indefinite,
): SnackbarResult {
    val visuals = QESnackbarVisuals(
        message = message,
        withDismissAction = withDismissAction,
        actionLabel = actionLabel,
        duration = duration,
        snackbarType = snackbarType,
    )
    return if (currentSnackbarData?.visuals != visuals) {
        showSnackbar(visuals)
    } else {
        SnackbarResult.Dismissed
    }
}
```

This code will skip each Snackbard if it has the same visuals as the current one - we can decide if we want to compare all of them or only the message, etc.

The advantage here is that this will work across the `QuickEditor`, because the logic is in the `SnackHost` rather than in the AvatarPicker. 

WDYT?

hamorillo:
I like the idea; it's much simpler. However, I think with that solution, when the currantSnackbar is, let's say, a successful one, we'll queue all the error ones that we try to show. Am I wrong?

Also, we'll be returning `SnackbarResult.Dismissed,`, which is not really true, but I don't think that's an important point as it is just for internal use.

AdamGrzybkowski:
> However, I think with that solution, when the currantSnackbar is, let's say, a successful one, we'll queue all the error ones that we try to show. Am I wrong?

Yes, but we could check the `SnackbarType` only with some casting if that's the preference. 

> Also, we'll be returning SnackbarResult.Dismissed, which is not really true, but I don't think that's an important point as it is just for internal use.

True, we could return a custom result type that would have `Skipped` for example.

hamorillo:
> Yes, but we could check the SnackbarType only with some casting if that's the preference.

Can you explain this a bit more? I'm not following the idea 😅

AdamGrzybkowski:
Sure, sth like this:
```
 return if ((currentSnackbarData?.visuals as? QESnackbarVisuals)?.snackbarType != snackbarType) {
        showSnackbar(
            QESnackbarVisuals(
                message = message,
                withDismissAction = withDismissAction,
                actionLabel = actionLabel,
                duration = duration,
                snackbarType = snackbarType,
            )
        )
} else {
        SnackbarResult.Dismissed
}
```

but I think we should at least check the message so that only "the same" information is skipped.

AdamGrzybkowski:
I don't think we should skip all the new snackbars while another is shown, we should skip only those that show the same information.

hamorillo:
> I don't think we should skip all the new snackbars while another is shown, we should skip only those that show the same information.

Yeah I agree. I think I haven't explained well previously.

What I was trying to say with: 

> However, I think with that solution, when the currantSnackbar is, let's say, a successful one, we'll queue all the error ones that we try to show. Am I wrong?

is that, with your proposed solution, we are only comparing with the currently shown snackbar. So, If I'm not wrong, we prevent queueing the same snackbar that is already visible, but we can queue different ones. For example:

Alt Text

If I perform a successful avatar switch (snack bar visible) and tap several times to switch the avatar, which, for some reason, all fail, I think all those failures will be queued. Is this possible with that approach?

AdamGrzybkowski:
ohh I get it now! I don't think we can do it with what I proposed.

The way I see it is that we can keep it simple and see if we get feedback about such situations, the end goal was to try and minimize the amount of Snacks we show when quickly selecting Avatars. Alternatively, we could try to extend the `SnackHostState` and implement your logic somewhere there to keep it centralized rather than scattered across individual pages/composables. This is just an idea though, not sure what's possible here.

hamorillo:
> The way I see it is that we can keep it simple and see if we get feedback about such situations, the end goal was to try and minimize the amount of Snacks we show when quickly selecting Avatars.

Maybe I've overthinked this. The truth is that I don't think we would have several cases where what I described will happen.  I've reverted my approach in favor of yours. 

Thanks for giving a different point of view! 🙇

AdamGrzybkowski (approved):

AdamGrzybkowski:
> Maybe I've overthinked this. The truth is that I don't think we would have several cases where what I described will happen. I've reverted my approach in favor of yours.


I guess KISS is my motto 😅 
Your solution was a bit more detailed, but would be a bit harder to scale it to other pages in the future.

PR #412, opened 2024-10-23
Refactor how MediaPickerPopup position is calculated

### LINKED ISSUE #367: QuickEditor: MediaPickerPopup has wrong offset on some devices
Image

AdamGrzybkowski:
Closes #367

### Description

On some devices, the `MediaPickerPopup` wouldn't be shown in the correct position on the screen. It's unclear why there would be a different behavior but it seemed like it could be caused by window insets. 

Anyway, I couldn't figure out how to fix the calculation with the previous logic, so I've decided to change how the position is calculated.  

Instead of doing the calculation outside of the `MediaPickerPopup` I have now put that logic inside the component. We now have to pass the anchor bounds rather than an already calculated offset. The alignment of the `Popup` is now hardcoded so we always know from where the offset will be calculated. In the new calculations, we include the `Popup` size as well, and this is the crucial part here. Additionally, I have allowed for basic customization to meet our needs. `anchorAlignment` can be passed (the difference between horizontal and vertical avatar picker content layout) to define how the horizontal anchor offset should be calculated. 

You can also notice that I replaced the `Dialog` from material library with the one from Composable Unstyled. The reason for that is because the `anchorBounds` we get are in relation to the whole screen (status and navigation bars included) but the material `Dialog` is only drawn in the safe area, meaning the `anchorBounds` could actually be bigger the whole available space of the dialog. This makes it impossible to properly draw the popup in a correct position. The one from the Composable Unstyled will be drawn fully unless limited by proper modifiers. 

| Before | After |
|---|---|
| Screenshot 2024-10-23 at 15 49 03 | Screenshot 2024-10-23 at 15 45 09 |
| Screenshot 2024-10-23 at 15 49 11 | Screenshot 2024-10-23 at 15 45 26 |
| Screenshot 2024-10-23 at 15 48 49 | Screenshot 2024-10-23 at 15 46 43 |

### Testing Steps

It's possible to test this with Pixel 6 Pro emulator (with device frame - not sure if that matters). I can repro this on all physical devices I have, only some emulators worked fine with the previous logic.

1. Go to the QE and tap `Upload image` 
2. Confirm the popup is aligned properly - it should be slightly above the upload button
3. Repeat with Vertical layout with images and an empty list.

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/HorizontalAvatarsSection.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/HorizontalAvatarsSection.kt
@@ -13,24 +13,19 @@ import androidx.compose.material3.Surface
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.geometry.Rect
+import androidx.compose.ui.geometry.Size
+import androidx.compose.ui.layout.boundsInRoot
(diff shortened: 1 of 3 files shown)


hamorillo (approved):
Tested in: 

- Nexus 5X (Emulator) 👍 
- Pixel 6 PRO (Emulator) 👍 
- Samsung S21 (Physical device) 👍 

🚀

PR #413, opened 2024-10-24
Update to the latest OpenAPI spec

AdamGrzybkowski:
### Description

This PR updates to the latest OpenAPI spec and updates the QE code to match the previous behavior of the upload avatar endpoint. 

### Testing Steps

There should be no visual or behavioral changes. 

Verify that the Avatar is selected automatically when a new photo is uploaded while no avatar is selected.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -77,7 +77,7 @@ comments:
     searchProtectedProperty: false
   OutdatedDocumentation:
     active: true
-    excludes: [ '**/test/**', '**/androidTest/**', '**/demoapp/**', '**/api/models/**', '**/restapi/models/**' ]
+    excludes: [ '**/test/**', '**/androidTest/**', '**/demoapp/**', '**/restapi/models/**', '**/restapi/apis/**' ]
     matchTypeParameters: true
     matchDeclarationsOrder: true
     allowParamOnConstructorProperties: false
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
@@ -143,37 +143,37 @@ internal class AvatarPickerViewModel(
                     failedUploadDialog = null,
                 )
             }
-            when (val result = avatarRepository.uploadAvatar(email, uri)) {
+            val selectAvatar = _uiState.value.emailAvatars?.selectedAvatarId == null
+            when (val result = avatarRepository.uploadAvatar(email, uri, selectAvatar)) {
                 is GravatarResult.Success -> {
                     fileUtils.deleteFile(uri)
-                    val isAutoSelected = _uiState.value.emailAvatars?.selectedAvatarId == null
-                    if (isAutoSelected) {
-                        fetchAvatars(showLoading = false, scrollToSelected = true)
-                        _uiState.update { currentState ->
-                            currentState.copy(
-                                uploadingAvatar = null,
-                                avatarUpdates = currentState.avatarUpdates.inc(),
-                            )
-                        }
-                        if (_uiState.value.emailAvatars?.selectedAvatarId != null) {
-                            _actions.send(AvatarPickerAction.AvatarSelected)
-                        }
-                    } else {
-                        _uiState.update { currentState ->
-                            val avatar = result.value
-                            currentState.copy(
-                                uploadingAvatar = null,
-                                emailAvatars = currentState.emailAvatars?.copy(
-                                    avatars = buildList {
-                                        add(avatar)
-                                        addAll(
-                                            currentState.emailAvatars.avatars.filter { it.imageId != avatar.imageId },
-                                        )
-                                    },
-                                ),
-                                scrollToIndex = null,
-                            )
-                        }
+                    val avatar = result.value
+                    if (avatar.selected == true) {
+                        _actions.send(AvatarPickerAction.AvatarSelected)
+                    }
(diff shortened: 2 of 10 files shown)


config/detekt/detekt.yml
```diff
@@ -77,7 +77,7 @@ comments:
     searchProtectedProperty: false
   OutdatedDocumentation:
     active: true
-    excludes: [ '**/test/**', '**/androidTest/**', '**/demoapp/**', '**/api/models/**', '**/restapi/models/**' ]
+    excludes: [ '**/test/**', '**/androidTest/**', '**/demoapp/**', '**/restapi/models/**', '**/restapi/apis/**' ]
```
AdamGrzybkowski, line 80: Detekt was complaining about `OutdatedDocumentation` but we have no control over those files, so added `apis` as well here.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -143,37 +143,37 @@ internal class AvatarPickerViewModel(
...
                 )
             }
-            when (val result = avatarRepository.uploadAvatar(email, uri)) {
+            val selectAvatar = _uiState.value.emailAvatars?.selectedAvatarId == null
+            when (val result = avatarRepository.uploadAvatar(email, uri, selectAvatar)) {
                 is GravatarResult.Success -> {
                     fileUtils.deleteFile(uri)
-                    val isAutoSelected = _uiState.value.emailAvatars?.selectedAvatarId == null
-                    if (isAutoSelected) {
-                        fetchAvatars(showLoading = false, scrollToSelected = true)
```
AdamGrzybkowski, line 151: We no longer need this. The Avatar in the response will now have a proper `selected` value set so we can update the UI based on that.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -143,37 +143,37 @@ internal class AvatarPickerViewModel(
                     failedUploadDialog = null,
                 )
             }
-            when (val result = avatarRepository.uploadAvatar(email, uri)) {
+            val selectAvatar = _uiState.value.emailAvatars?.selectedAvatarId == null
+            when (val result = avatarRepository.uploadAvatar(email, uri, selectAvatar)) {
```
hamorillo: > "description": "Determines if the uploaded image should be set as the avatar for the email. If not passed, the image is only selected as the email's avatar if no previous avatar has been set. Accepts '1'/'true' to always set the avatar or '0'/'false' to never set the avatar.",

❓ Based on that, I wonder If we should not pass any value to "selectAvatar" and let the backend handle the default action. With that code, we are forcing to select the avatar but I think that's unnecessary as, if there is no avatar selected, the backend will select it by default.

config/detekt/detekt.yml
hamorillo: 🤔  I wonder why it hasn't complained before.

In fact, it's just:

```
gravatar/src/main/java/com/gravatar/restapi/apis/AvatarsApi.kt:73:17: Documentation of uploadAvatar is outdated [OutdatedDocumentation]
```

Not sure why it's only that method.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
AdamGrzybkowski: Makes sense! I realized I should have make the `hash` and `selectAvatar` nullable in the `AvatarService` so I've updated that as well.

config/detekt/detekt.yml
AdamGrzybkowski: yeah, but still we don't control that file, so I think it's valid to exclude it.

hamorillo: Yeah, just trying to understand the reason behind the error.

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
```diff
@@ -28,15 +28,28 @@ public class AvatarService(private val okHttpClient: OkHttpClient? = null) {
...
         withContext(GravatarSdkDI.dispatcherIO) {
             val service = instance.getGravatarV3Service(okHttpClient, oauthToken)
 
             val filePart =
                 MultipartBody.Part.createFormData("image", file.name, file.asRequestBody())
 
-            val response = service.uploadAvatar(filePart)
+            val response = service.uploadAvatar(
+                data = filePart,
+                selectedEmailHash = hash.toString(),
```
hamorillo: ⚠️ We need to use a safe call here. If not, when the hash is `null`, `toString` will return the `"null"` string.
```suggestion
                selectedEmailHash = hash?.toString(),
```

hamorillo (changes requested):

gravatar/src/main/java/com/gravatar/services/AvatarService.kt
hamorillo: 💡 Maybe we can add a unit test for that?

AdamGrzybkowski: Good catch! I will add a test 👍

AdamGrzybkowski: Done!

hamorillo (approved):
Thanks for the changes! 🚀

PR #414, opened 2024-10-24
Quick Editor: More options button in avatar picker

### LINKED ISSUE #411: Quick Editor: More options button with empty dialog menu
Image

hamorillo:
Closes #411 

### Description

Part of the milestone `2.1.0` is to add the possibility to modify the alternative text, delete an avatar, or define its rating.

The common entry point for those three options is a button in each avatar in the avatar picker.  

Image

This PR implements the more options button to allow the parallelized implementation of those three features.

I've added an "Alt Text" button for the sake of define all the architecture and how it should look, however it doesn't execute any action under the hood.

I've generalized our previous `Popup` menu, as this one is pretty similar, so we can keep all the code centralized.

For now, the popup is aligned with `CenterHorizontally`; we can iterate on this if we think that's not the best option.

image

### Testing Steps

1. Open the QE in the demo-app
2. Verify you can see the more options button in each of the avatars
3. Click any of the more options buttons and verify the Popup with the menu is opened
4. Smoke test the upload new avatar popup menu


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/PickerPopup.kt
```diff
@@ -100,9 +103,15 @@ private fun PickerPopup(
...
                         LazyColumn {
-                            itemsIndexed(popupButtons) { index, item ->
-                                item()
-                                if (index < popupButtons.size - 1) {
+                            itemsIndexed(popupItems) { index, item ->
+                                PopupButton(
+                                    text = stringResource(item.text),
+                                    iconRes = item.iconRes,
+                                    contentDescription = stringResource(item.contentDescription),
+                                    shape = RoundedCornerShape(topStart = 8.dp, topEnd = 8.dp),
```
AdamGrzybkowski: This should be different for the first, middle, and last items. You can see it's incorrect when you open the `MediaPickerPopup` and tap&hold the `Take Photo` button. The ripple will have rounded corners at the top not matching the shape of the button.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
```diff
@@ -64,10 +80,25 @@ internal fun SelectableAvatar(
...
+                moreOptionsPopupVisible = true
+            }, Modifier.align(Alignment.BottomEnd))
+
             AvatarLoadingState.Loading -> LoadingOverlay()
             is AvatarLoadingState.Failure -> FailureOverlay()
         }
+
+        if (moreOptionsPopupVisible) {
+            AvatarMoreOptionsPickerPopup(
+                anchorAlignment = Alignment.CenterHorizontally,
```
AdamGrzybkowski, line 93: It seems like sth is off with the alignment - it's always centered on the screen. Shouldn't it be centered to the anchor? 

It might not be related to this PR, but we might have to work on that calculation to fit this usage a bit better.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/PickerPopup.kt
hamorillo: Good catch! Updated!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
hamorillo: I mentioned this in the PR description; I think we can iterate over the alignment in a different PR to keep this one small. But if you think that's a blocker for this one, I can work on it.

AdamGrzybkowski: No, no, not a blocker!

AdamGrzybkowski (approved):
:shipit:

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
hamorillo: 👍 Ticket created -> #419

PR #418, opened 2024-10-25
Resolve conflict between `2.1.0` and `trunk`

hamorillo:
### Description

Branch where I resolved the conflict between `2.1.0` and `trunk` in order to merge #417.

⚠️ To solve the string conflicts, I've removed all strings from all files. We'll need to update strings when we have the new versions from Glotpress. I think this should be safe enough.

### Testing Steps

- Smoke test the demo-app


AdamGrzybkowski (approved):
The diff is almost impossible to view here because of the docs, but I checked it locally and it looks good.

PR #425, opened 2024-10-31
QuickEditor: Implement GravatarQuickEditorActivity

### LINKED ISSUE #424: QuickEditor: Remove the `singleTask` launchMode limitation from the QE

hamorillo:
Closes #424 

### Description

To simplify the integration we need to remove the limitation of using the QuickEditor from an activity using the launchMode="singleTask".

The approach is to create an intermediate activity that can assume the complexity and the launch mode required by the QE. The behavior should be as much as possible as it is using the Composable directly.

https://github.com/user-attachments/assets/a0b8c091-ee85-4cf7-8e49-520f8b090d11

### Testing Steps

* In the `demo-app` manifest uncomment the `Gravatar QE Activity` part and comment the intent filter in `MainActivity` (or apply this [patch](https://github.com/user-attachments/files/17733845/Gravatar-Android-15-17-02.patch))
* Launch the `demo-app`
* Tap over `Test with Activity without Compose`
* Tap over the second button, `Update Avatar with QuickEditor Activity`
* Verify everything works as expected. 
  - OAuth flow with different browsers and using custom tabs (Firefox would be nice)
  - Avatar switch (Note that with this approach, we only notify the avatar change when our activity finishes)
  - Close the bottom sheet without switching avatars
  - You can also test rotating the device orientation with the bottom sheet (GravatarQuickEditorActivity) visible

### DIFF
--- a/docs/get-started/get-started.md
+++ b/docs/get-started/get-started.md
@@ -364,7 +364,7 @@ getQEResult.launch(
 )
 ```
 
-It's important to note that using the `GravatarQuickEditorActivity` you'll only receive the result of the Quick Editor when it's dismissed not instantly as with using the `@Composable` component from your `singleTask` activity.
+It's important to note that using the `GravatarQuickEditorActivity` you'll only receive the result of the Quick Editor when it's dismissed not instantly as with using the `@Composable` component from your `singleTask` activity (see [Section 1.1](#11-using-you-own-activity-with-androidlaunchmodesingletask-recommended)).
 
 In the `demo-app` you can find a detailed implementation showing how to use the provided activity. See `QuickEditorTestActivity`.
 
--- a/docs/get-started/get-started.md
+++ b/docs/get-started/get-started.md
@@ -244,6 +244,8 @@ To do that the QuickEditor needs an authorization token to perform requests on b
 
 ### 1. Let the Quick Editor handle the OAuth flow
 
+#### 1.1 Using you own activity with `android:launchMode="singleTask"` (Recommended)
+
 Quick Editor can handle the heavy lifting of running the full OAuth flow, so you don't have to do that. We will still need a few things from you.
 First, you have to go to [OAuth docs](https://docs.gravatar.com/oauth/) and create your Application. Define the `Redirect URLs`.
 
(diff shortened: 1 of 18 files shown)


docs/get-started/get-started.md
```diff
@@ -311,6 +313,61 @@ When the user logs out form the app, make sure to run:
...
+    GravatarQuickEditorActivity.GravatarEditorActivityArguments(
+        GravatarQuickEditorParams { ... },
+        AuthenticationMethod.OAuth(
+            OAuthParams { ... },
+        ),
+    ),
+)
+```
+
+It's important to note that using the `GravatarQuickEditorActivity` you'll only receive the result of the Quick Editor when it's dismissed not instantly as with using the `@Composable` component from your `singleTask` activity.
```
AdamGrzybkowski: I would link to the 1.1 above just in case someone missed it. WDYT?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/GravatarQuickEditorActivity.kt
```diff
@@ -0,0 +1,142 @@
...
+                authenticationMethod = arguments.authenticationMethod,
+                onAvatarSelected = { avatarHasChanged = true },
+                onDismiss = { finishWithResult() },
+            )
+        }
+    }
+
+    override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+        outState.putBoolean("avatarHasChanged", avatarHasChanged)
```
AdamGrzybkowski: ⛏️ Can we extract `avatarHasChanged` to a property?

AdamGrzybkowski (approved):
Great work! 🎉 
I tested it with: Opera, Chrome, Firefox, DuckDuckGo, and Brave. All worked fine.

Left some minor comments, but the functionality itself is 🟢

docs/get-started/get-started.md
hamorillo: Done! 👍

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/GravatarQuickEditorActivity.kt
hamorillo: Sure! [02dd869](https://github.com/Automattic/Gravatar-SDK-Android/pull/425/commits/02dd869b162f3eed0400189f75a3282ca1e448b6)

PR #438, opened 2024-11-14
Update string value

### LINKED ISSUE #423: Clipped copy in the avatar selected confirmation Snack
It's `May take a few minutes to appear everywhere` instead of `It may take a few minutes to appear everywhere`.

AdamGrzybkowski:
Closes #423

### Description

The copy shown after selecting an avatar seems to be clipped. It was taken from the Figma, but it could have been a typo. 

### Testing Steps

1. Open the QE
2. Select an avatar
3. Confirm the copy in the Snackbar


hamorillo (approved):
🚀

PR #441, opened 2024-11-14
Update README.md with mention to iOS repo

mlumeau:
Added a mention to the iOS repository for devs who are looking for an iOS equivalent.


hamorillo (approved):
🚀

PR #450, opened 2024-11-20
QuickEditor: Delete avatar

### LINKED ISSUE #236: QuickEditor: Delete Avatar
Given the application has internet connectivity and I am authenticated
As a user with at least one avatar
I want to delete an avatar
So that I can clean up unused avatars

**Acceptance Criteria:**

- The option to delete an avatar must be available in the avatar's contextual menu.
- Users must be able to quickly delete an avatar.
- The avatar must be removed from the UI immediately without waiting for the response.
- If avatar deletion fails:
  - Show an error toast "Error deleting the image" with a "Try again" action.
- Bring back the removed avatar.
- A confirmation dialog must be shown to the user before deleting the avatar. "Are you sure you want to delete this image?"

Alt Text

hamorillo:
Closes #236

### Description

This PR implements the base of the delete avatar flow. 

**Note:** When opening this PR, the endpoints are not working yet, so you'll always receive a `404`; however, I think testing the whole flow is possible. However, it's possible that we'll need to make minor adjustments when the endpoint is released.

https://github.com/user-attachments/assets/7cc13327-c440-469f-8d13-61aa6ffb3d42

This PR will be followed by two more to add the confirmation dialog (#449)  and the error toast (#448).

### Testing Steps

1. Open the QE
2. In any of your uploaded avatars, tap the options button (...)
3. Select `Delete`
4. You should see how the avatar disappears immediately (as we are assuming the flow will work)
5. You should see the avatar reappear when we receive the `404` (that's the work part of not having the endpoints yet :) )
6. If you want, you can check logcat a see the error:
```
2024-11-20 12:05:40.816 10320-10465 AvatarService           com.gravatar.demoapp                 W  Network call unsuccessful trying to delete Gravatar avatar: Response{protocol=h2, code=404, message=, url=https://api.gravatar.com/v3/me/avatars/bfb35130743eb019b309a3c219ee77b5}.body
```
7. You can also play with a network interceptor and mock a 200 answer (I don't think it's completely necessary)

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
@@ -45,6 +45,7 @@ import androidx.lifecycle.viewmodel.compose.viewModel
 import com.gravatar.extensions.defaultProfile
 import com.gravatar.quickeditor.R
 import com.gravatar.quickeditor.data.repository.EmailAvatars
+import com.gravatar.quickeditor.ui.components.AvatarOption
 import com.gravatar.quickeditor.ui.components.AvatarsSection
 import com.gravatar.quickeditor.ui.components.EmailLabel
 import com.gravatar.quickeditor.ui.components.ErrorSection
@@ -188,12 +189,11 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onEvent: (AvatarPickerEv
                                 is AvatarUi.Uploaded -> onEvent(AvatarPickerEvent.AvatarSelected(avatarUi.avatar))
                             }
                         },
-                        onAltTextSelected = { /* Future implementation */ },
-                        onDeleteSelected = { avatarUi ->
-                            when (avatarUi) {
-                                is AvatarUi.Uploaded -> onEvent(AvatarPickerEvent.AvatarDeleteSelected(avatarUi.avatar))
-                                is AvatarUi.Local -> {
-                                    // Local images are not uploaded yet, so we don't need to delete them
+                        onAvatarOptionClicked = { avatar, avatarOption ->
+                            when (avatarOption) {
+                                AvatarOption.ALT_TEXT -> Unit
+                                AvatarOption.DELETE -> {
+                                    onEvent(AvatarPickerEvent.AvatarDeleteSelected(avatar))
                                 }
                             }
                         },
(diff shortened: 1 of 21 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AvatarsSection.kt
```diff
@@ -37,6 +37,7 @@ internal fun AvatarsSection(
     state: AvatarsSectionUiState,
     onAvatarSelected: (AvatarUi) -> Unit,
     onAltTextSelected: (AvatarUi) -> Unit,
+    onDeleteSelected: (AvatarUi) -> Unit,
```
AdamGrzybkowski: 💡 The number of callbacks will grow quite a bit. I had the same problem when adding "Download" option and figured we could have one callback with an extra param. Here's the commit, that you can check or even cherry-pick: https://github.com/Automattic/Gravatar-SDK-Android/commit/ceb89ace810962d769b29a0ff43903c6a702e3c9


Also, we don't have to pass the `AvatarUi` here - we only need these callbacks for the images that are uploaded to the server. The above commit tackles that as well.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -189,6 +189,14 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onEvent: (AvatarPickerEv
                             }
                         },
                         onAltTextSelected = { /* Future implementation */ },
+                        onDeleteSelected = { avatarUi ->
+                            when (avatarUi) {
+                                is AvatarUi.Uploaded -> onEvent(AvatarPickerEvent.AvatarDeleteSelected(avatarUi.avatar))
+                                is AvatarUi.Local -> {
+                                    // Local images are not uploaded yet, so we don't need to delete them
+                                }
+                            }
```
AdamGrzybkowski: This would be cleaned with the suggestion from the previous comment. We can pass a different object here, as we don't need this for `Local` avatars.

AdamGrzybkowski (commented):
Looks good! 

I left one comment about the callback that I think would be great to do now, given you're the first to add another option to the contextual menu.

On the UI side, I would explore some item insertion/removal animation. I believe this can be achieved by adding `Modifier.animateItem()` to each item in the Lazy row/column. This should help a bit with the UI changes being so fast (no progress indicator).

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AvatarsSection.kt
hamorillo: Great suggestion @AdamGrzybkowski! It looks much better with your refactor. I've cherry-picked your commit and resolved the conflicts. Thanks!

hamorillo:
> On the UI side, I would explore some item insertion/removal animation. I believe this can be achieved by adding Modifier.animateItem() to each item in the Lazy row/column. This should help a bit with the UI changes being so fast (no progress indicator).

Done! [8709436](https://github.com/Automattic/Gravatar-SDK-Android/pull/450/commits/870943683597f9c513885fc8d29aff0c71c3b7fa)

AdamGrzybkowski (approved):
:shipit:

PR #451, opened 2024-11-20
Quick Editor: Show error toast with retry action when delete avatar fails

### LINKED ISSUE #448: Show error toast with retry button when delete flow fails

hamorillo:
Closes #448 

### Description

When the avatar deletion fails, apart from adding the avatar again to the list, we want to show a toast notifying the user about what happened and giving them the possibility of retry.

**Note**: When opening this PR, the endpoints are not working yet, so you'll always receive a 404; however, I think testing the whole flow is possible. However, it's possible that we'll need to make minor adjustments when the endpoint is released.

https://github.com/user-attachments/assets/34051fce-41aa-46c6-afde-7c6b6cc27af2

### Testing Steps

1. Open the QE
2. In any of your uploaded avatars, tap the options button (...)
3. Select `Delete`
4. You should see how the avatar disappears immediately (as we are assuming the flow will work)
5. You should see the avatar reappear when we receive the 404 (that's the work part of not having the endpoints yet :) ) along with the error toast
6. Tap `Try again` and verify steps 4 and 5.

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
@@ -57,6 +57,7 @@ import com.gravatar.quickeditor.ui.editor.AvatarPickerContentLayout
 import com.gravatar.quickeditor.ui.editor.GravatarQuickEditorParams
 import com.gravatar.quickeditor.ui.editor.bottomsheet.DEFAULT_PAGE_HEIGHT
 import com.gravatar.quickeditor.ui.extensions.QESnackbarHost
+import com.gravatar.quickeditor.ui.extensions.QESnackbarResult
 import com.gravatar.quickeditor.ui.extensions.SnackbarType
 import com.gravatar.quickeditor.ui.extensions.showQESnackbar
 import com.gravatar.restapi.models.Avatar
@@ -100,6 +101,7 @@ internal fun AvatarPicker(
             lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
                 viewModel.actions.collect { action ->
                     action.handle(
+                        viewModel = viewModel,
                         cropperLauncher = cropperLauncher,
                         onAvatarSelected = onAvatarSelected,
                         onSessionExpired = onSessionExpired,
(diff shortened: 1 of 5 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -252,6 +255,20 @@ private fun AvatarPickerAction.handle(
         }
 
         AvatarPickerAction.InvokeAuthFailed -> onSessionExpired()
+
+        is AvatarPickerAction.AvatarDeletionFailed -> {
+            scope.launch {
+                if (snackState.showQESnackbar(
```
hamorillo, line 262: ❓ Our default behavior when the snackbar has action is to don't dismiss. I wonder if we should add a limited duration, maybe long enough for the user to tap "Try again" but not to keep that toast on the screen. Wdyt?

AdamGrzybkowski: Yeah, sounds good to me!

gravatar-quickeditor/src/main/res/values/strings.xml
```diff
@@ -19,6 +19,7 @@
     <string name="gravatar_qe_photo_library_icon_description">Photo library</string>
     <string name="gravatar_qe_capture_photo_icon_description">Capture Photo</string>
     <string name="gravatar_qe_avatar_selection_error">Ooops, there was an error selecting the avatar.</string>
+    <string name="gravatar_qe_avatar_delete_avatar_error">Error deleting the image.</string>
```
AdamGrzybkowski, line 22: Do we get any meaningful errors from the server that we could use? Not necessarily directly, but to make this message less generic?

AdamGrzybkowski (approved):
LGTM 👍

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
hamorillo: 👍 Applied `SnackbarDuration.Long`

gravatar-quickeditor/src/main/res/values/strings.xml
hamorillo: As far as I know, we don't have concrete errors directly related to the avatar deletion at the moment. However, your question made me think about general errors (ex. token error). I think we should handle, at least, the expiration token error. Wdyt?

AdamGrzybkowski: Yeah, we should. But we need that one to be handled more globally so that we don't have to repeat this on each request. So likely a separate issue anyway.

hamorillo: Agree. #453

PR #454, opened 2024-11-21
Quick Editor: Fix popup alignment above anchor

### LINKED ISSUE #419: QuickEditor: Improve avatar more options button alignment
Improve/review button alignment as per discussion [here](https://github.com/Automattic/Gravatar-SDK-Android/pull/414#discussion_r1818729018)

mlumeau:
Closes #419 

### Description

The avatar contextual menu popup was previously centered on screen, a behavior inherited from the other popup menu for the "upload" button (`MediaPickerPopup`). 

I changed the `PickerPopup` so it's always horizontally centered on the anchor, thus removing a bunch of unneeded code (mainly parameters). The "Upload" button is centered on screen anyway, so the popup should also be centered. If for some reason we have the Gravatar bottomsheet appear on the side (tablet split-screen mode, side-by-side fragments? idk), I'd prefer that the popup appears above the button and not in the middle of the screen.

https://github.com/user-attachments/assets/a00e8da0-2b47-42c8-8a2a-4de3b6e936e5

### Testing Steps

1. Open the QE
2. Tap the options button (...) on an avatar on the left
3. See that the popup appears on the left above the avatar
2. Tap the options button (...) on an avatar on the right
3. See that the popup appears on the right above the avatar
2. Tap the options button (...) on an avatar on the center
3. See that the popup appears centered above the avatar
2. Tap the Upload button
3. See that the popup appears centered above the button


AdamGrzybkowski:
Unfortunately, this breaks the alignment of the `MediaPickerPopup` in the `vertical` content layout. Here's the comparison (including the Figma design).

| Before | After | Figma |
|---|---|---|
| Screenshot 2024-11-21 at 12 42 27 | Screenshot 2024-11-21 at 12 43 04 | Screenshot 2024-11-21 at 12 43 21 |

mlumeau:
> Unfortunately, this breaks the alignment of the MediaPickerPopup in the vertical content layout. 

I know, but the popup menu of the avatar directly below it would be centered, so the behavior would be inconsistent...

AdamGrzybkowski:
>I know, but the popup menu of the avatar directly below it would be centered, so the behavior would be inconsistent...

That true. Now I can see that we don't have a spec for how this should work. 

In my mind, the Popup alignment should change depending on where on the horizontal axis the Avatar is. 

So the column on the left would have the popup aligned to the left edge of the avatar, the center one would have a centered aligned Popup, and the one on the right would have the Popup aligned to the right edge. 

This would have to be calculated in a smarter way for the `horizontal` layout as we don't have the columns there, but the same rules should apply. 

but that's just my vision... I will ask Andrei :D

mlumeau:
Meanwhile I reverted some of my changes and instead added support for Start and End alignment. 

> In my mind, the Popup alignment should change depending on where on the horizontal axis the Avatar is.

Maybe we can achieve this now by changing the alignment based on apparent position in list? But if it proves too complicated, I wouldn't spend too much time on this.

AdamGrzybkowski:
>Maybe we can achieve this now by changing the alignment based on apparent position in list? But if it proves too complicated, I wouldn't spend too much time on this.

I played a bit with this now, and to be honest we might be ok with just the `start` alignment, but we would have to restrict the area where the Popup is drawn so there's always some padding on the left and right.

mlumeau:
I'll try to make the popup clamp to edges, with padding.

mlumeau:
Here you go. I didn't pass the padding, I think its feasible though.

mlumeau:
Now with 100% more padding. I added the compute function in a new `UIUtils` file under the `extensions` package, I'm open to suggestions for this.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/LazyAvatarList.kt
```diff
@@ -36,9 +38,14 @@ internal fun LazyAvatarRow(
...
 
+    val density = LocalDensity.current
+    val layoutDirection = LocalLayoutDirection.current
+
     LazyRow(
         horizontalArrangement = horizontalArrangement,
-        modifier = modifier.onGloballyPositioned { coordinates -> parentBounds = coordinates.boundsInRoot() },
+        modifier = modifier.onGloballyPositioned { coordinates ->
+            parentBounds = coordinates.boundsInRoot().getPaddedBounds(contentPadding, density, layoutDirection)
+        },
```
AdamGrzybkowski: ⛏️ Maybe I'm nitpicking, but I think for the horizontal list we could have no padding because the Avatar themselves can be drawn from edge to edge of the Parent composable. 

We use `contentPadding` on the `LazyRow` which means only the first and last element will have the padding applied, but when you scroll it will look like there's no padding. 

Screenshot 2024-11-22 at 10 31 09

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/PickerPopup.kt
```diff
@@ -42,12 +41,14 @@ import com.composables.core.rememberDialogState
 internal fun PickerPopup(
     anchorAlignment: Alignment.Horizontal,
     anchorBounds: Rect,
+    maxBounds: Rect? = null,
```
AdamGrzybkowski: 💡 Although it's very obvious what this name is about now, I'm not sure I would feel the same way sometime in the future. What do you think about renaming to something like `popupDrawArea` or similar that would more cleary describe what is this used for?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
```diff
@@ -90,8 +91,9 @@ internal fun SelectableAvatar(
 
         if (moreOptionsPopupVisible) {
             AvatarMoreOptionsPickerPopup(
-                anchorAlignment = Alignment.CenterHorizontally,
+                anchorAlignment = Alignment.Start,
                 anchorBounds = popupAnchorBounds,
+                maxBounds = parentBounds,
```
AdamGrzybkowski: Completely unrelated to this PR, but just wanted to share that I'm having doubts about whether this Popup should be inside the `SelectableAvatar`. It feels like too much for such a small composable, and now with the extra params that have to be passed just for that, this makes it clearer I think. 

@hamorillo Thoughts on this?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/extensions/UIUtils.kt
```diff
@@ -0,0 +1,25 @@
+package com.gravatar.quickeditor.ui.extensions
+
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.ui.geometry.Rect
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.LayoutDirection
+
+internal fun Rect.getPaddedBounds(
```
AdamGrzybkowski: Not a big deal at this point as we only have. this, but I would try to avoid having a file with all possible UI utils. Those can be for very different things. So smaller, specialized utils would be my preference. `RectExtensions/Utils` for example.

AdamGrzybkowski (approved):
LGTM 👍 

There's one comment about the padding values in the horizontal layout that I would maybe change, but other than that this works much better now!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/extensions/UIUtils.kt
mlumeau: I'm conflicted about having "too" specialized utils files too 🤔. I'd refactor this later if we have other utils joining the list.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/LazyAvatarList.kt
mlumeau: But then the first element will still have padding because it's aligned with the `Start` of the avatar, which is drawn with padding. And for the last element, the popup won't have padding at the `End`, although the last avatar has padding.

AdamGrzybkowski: Hmm not sure I'm following. The Popup is drawn also based on the anchor (Avatar) so with the `Alignment.Start` it should be drawn properly for the first element even if it has some extra padding.

mlumeau: Exactly, so in regard to the list itself it has padding.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
hamorillo: I have the same feeling. In fact, when I implemented the Popup, I had several doubts about whether it should be outside the SelecrtableAvatar. 

The benefit of having the Popup within the SelectableAvatar is that it's encapsulated in just one place. However, I feel now that we are adding a responsibility that Composable should be outside and managed by a parent composable. 

Said that, I would create an issue with the lowest priority for now.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/LazyAvatarList.kt
hamorillo: Good point @AdamGrzybkowski . I agree that we can remove the padding for the horizontal list.

hamorillo (approved):
Looks great @mlumeau ! 💯

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/LazyAvatarList.kt
mlumeau: Alright, if you both agree, I'll do it! 😃

mlumeau: Then the util function won't be needed outside of the vertical list, so we can make it private to that file

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/SelectableAvatar.kt
AdamGrzybkowski: Done https://github.com/Automattic/Gravatar-SDK-Android/issues/457

PR #455, opened 2024-11-21
QuickEditor: Avatar image download option

### LINKED ISSUE #447: QuickEditor: Download an Avatar
Given the application has internet connectivity and I am authenticated
As a user with at least one avatar
I want to download the avatar image in its original resolution
So that I can use the image outside the application in high quality.

Acceptance Criteria:
- Downloading an avatar image is accessible in the avatar’s contextual menu.
- The image is downloaded in its original resolution.
- The image is downloaded in a publicly accessible directory.
- A message confirms download success or failure, and failed downloads can be retried.

AdamGrzybkowski:
Closes #447

### Description

There's more than one way to solve this task. But there's no solution (that I know of) that would drop the requirement for the `WRITE_EXTERNAL_STORAGE` permission on API 28 and lower (pre-SAF).

1. We could handle the download ourselves (to a temp private directory) and then request to create a file in a public one [via Intent](https://developer.android.com/training/data-storage/shared/documents-files#create-file). 

This would require from us to do all the heavy work of showing notifications, handling retries, etc. We would likely need extra dependency for this (WorkManager), so that wouldn't be ideal. On top of that, the flow would be longer, as the user would have to pick the directory for the file to be saved. Although some would prefer that, in general, I think it's enough to save it in `Downloads` folder. 

2. Use `DownloadManager`.

This still requires the `WRITE_EXTERNAL_STORAGE` on API 28 and lower, but it handles the download, notifications, and retries. The flow on AP I29+ is extremely simple as there are no extra actions required from the user. No extra dependency as well.

There are some cons. 
- The user might disable the `DownloadManager`. Thankfully, we can detect that and direct the user to a proper settings screen to enable it.
- We don't have full control, but we can listen for updates via `BroadcastReceiver`. For now, I've skipped listening for updates. There's a notification added once the download is completed, and I didn't see much value in duplicating that in the QE via Snackbar. 

I've decided to pick option nr 2 as it's simple and offers a quicker flow for the user. We can always evaluate when we encounter issued with the `DownloadManager`.
 
| API 28 | API 28 Permission rationale dialog | API 29+ | 
|---|---|---|
| download_api28 | Screenshot 2024-11-21 at 11 29 08 | download_api_29+ |

### Testing Steps

1. Launch the QE on API 29+
2. Tap on the`...` button on one Avatar
3. Tap Download image
4. Confirm that a Snackbar with information about queued download was shown
5. Check the notification for the image download

Repeat on a device with API 28 or lower. Verify the flow with granted and denied permission.

`DownloadManager` disabled scenario:

1. Go to System settings -> Apps  -> All Apps -> Menu -> Show system apps
2. Look for DownloadManager
3. Disable it
4. Go back to QE and try downloading the avatar


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/DownloadManagerDisabledAlertDialog.kt
```diff
@@ -0,0 +1,29 @@
...
+import androidx.compose.material3.AlertDialog
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.stringResource
+import com.gravatar.quickeditor.R
+
+@Composable
+internal fun DownloadManagerDisabledAlertDialog(
+    isVisible: Boolean,
```
hamorillo: ❓ Why did you add the `isVisible` param here? This way, we'll add the composable to the graph (despite it internally being nothing).

.idea/kotlinc.xml
```diff
@@ -1,5 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="Kotlin2JvmCompilerArguments">
```
hamorillo: ❓Maybe we should remove this file from the repo.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/ImageDownloader.kt
```diff
@@ -0,0 +1,71 @@
...
+            else -> {
+                val request = DownloadManager.Request(Uri.parse(imageUrl.withMaxSizeQueryParam().toString()))
+                    .addRequestHeader("X-Platform", "Android")
+                    .addRequestHeader("X-SDK-Version", BuildConfig.SDK_VERSION)
+                    .addRequestHeader("X-Source", appName)
+                    .setNotificationVisibility(DownloadManager.Request.VISIBILITY_VISIBLE_NOTIFY_COMPLETED)
+                    .setMimeType("image/*")
+                    .setDestinationInExternalPublicDir(
+                        Environment.DIRECTORY_DOWNLOADS,
+                        "gravatar_image_${System.currentTimeMillis()}.png",
```
hamorillo, line 37: ⛏ Totally nitpicking. Maybe we can try to use the same name for the same image so in the download folder, we'll see that it's the same image (the system adds the file-name-2.png

As I said, we can consider this in a different PR if we think it looks better.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -135,6 +144,38 @@ internal fun AvatarPicker(
...
+    val writeExternalStoragePermissionLauncher = rememberLauncherForActivityResult(
+        ActivityResultContracts.RequestPermission(),
+    ) { isGranted: Boolean ->
+        if (isGranted) {
+            avatarToDownload?.let { onEvent(AvatarPickerEvent.DownloadAvatarTapped(it)) }
+            avatarToDownload = null
+        } else {
+            storagePermissionRationaleDialogVisible = true
+        }
+    }
```
hamorillo: ❓ Shouldn't we set to `null` the `avatarToDownload` in both cases?

```suggestion
        if (isGranted) {
            avatarToDownload?.let { onEvent(AvatarPickerEvent.DownloadAvatarTapped(it)) }
        } else {
            storagePermissionRationaleDialogVisible = true
        }
        avatarToDownload = null
    }
```

hamorillo (commented):
Works great! 👏  Left some minor comments.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/ImageDownloader.kt
AdamGrzybkowski: I thought about it and tried it, but it didn't work like that. 
The image had the same file name... Instead of sth like `filename(1).png` it was `filename.png(1)`. I can try on different devices, but still, the end result was the same, multiple files were created.

.idea/kotlinc.xml
AdamGrzybkowski: Say no more 😅

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/DownloadManagerDisabledAlertDialog.kt
AdamGrzybkowski: I've used this already in some Dialogs, eg. `FailedAvatarUploadAlertDialog`. 

The idea is that you don't have to put all those `if ()` conditions in the main composable. I can already see the number of dialogs growing, so the less code we have to put them the better I think.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/ImageDownloader.kt
hamorillo: Maybe different devices handle this in other ways. Nevermind then. 👍

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/DownloadManagerDisabledAlertDialog.kt
hamorillo: I agree that it is convenient but has some drawbacks. As I mentioned, with this approach, the `DownloadManagerDisabledAlertDialog` will always be added, and any updates to other state variables within the composable would still trigger recomposition, even though the content isn’t displayed. 

In this case, the other params are lambdas, so the undesired recompositions are unlikely to happen (I guess), but I still have some doubts about this approach. Anyway, not a blocker.

hamorillo (approved):

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/DownloadManagerDisabledAlertDialog.kt
AdamGrzybkowski: Not sure that is true. I've checked it with the layout inspector and here's the result:

| Dialog visible | Dialog hidden |
|---|---|
| Screenshot 2024-11-22 at 09 39 22 | Screenshot 2024-11-22 at 09 39 48 |

I can't find anything related to the dialog when it's hidden.

hamorillo: 🤯 🤔 I've checked I have the same result; however, I don't understand why the dialog composable is not there. I would expect it to be there despite it being "empty". Maybe Compose is smart enough to get rid of it in those cases 🤷‍♂️ . 

I've asked Perplexity, and it thinks the same as me (Could I bias its answer with my question?) 😅 but the truth is the composable is not there, so we are probably both wrong. 😄 

Please feel free to merge the PR. I'll investigate how this works when I have some free time, just out of curiosity. 

:shipit:

PR #458, opened 2024-11-22
QuickEditor: No avatar seleted alert banner

### LINKED ISSUE #443: QuickEditor: Warning when there is not avatar selected
## User Story

**Given** the application has internet connectivity and I am authenticated
**As** a user with at least one avatar
**I want** to inform the user that they have no selected avatars
**So** they know they need to select one

## Acceptance Criteria

- When there's no selected avatar:
  - Show a warning view at the top of the page saying: "No image is assigned to this address. Please select an image otherwise the default will be used."
  - The profile card should show whatever the default avatar returns from the backend (at the moment it's the Gravatar logo).
  - The alert should be dismissible and re-appear when the QE is closed and opened

Image

hamorillo:
Closes #443 

### Description

This PR adds an informative banner to encourage users to select an avatar instead of using the default one. Per the definition, the banner should be dismissable but must reappear if the condition occurs again. (Ex: selecting an avatar and deleting it).

https://github.com/user-attachments/assets/0b4f3727-3d46-4d86-a00e-6821e9c880d9

### Testing Steps

**Note:** As the deletion flow is not working yet in the SDK, you'll need a browser to remove the selected avatar.

1. Using an account without a selected avatar, open the QE.
2. Verify you can see the alert banner
3. Tap in the X to dismiss the banner
4. Verify the banner is dismissed
5. Close the QE and open it again
6. Verify the banner is visible again
7. Upload a new image 
8. Verify the banner disappears.
9. Close the QE
10. Using the browser, remove the last uploaded (and selected) avatar
11. Open the QE and select any of the already uploaded avatars
12. Verify the banner disappears.

Please verify the UI in both light and dark themes.


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -147,6 +149,11 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onEvent: (AvatarPickerEv
     var loadingSectionHeight by remember { mutableStateOf(DEFAULT_PAGE_HEIGHT) }
     var storagePermissionRationaleDialogVisible by rememberSaveable { mutableStateOf(false) }
     var avatarToDownload: Avatar? by remember { mutableStateOf(null) }
+    var nonSelectedAvatarAlertVisible by rememberSaveable { mutableStateOf(uiState.showNonSelectedAvatarAlert) }
+
+    LaunchedEffect(uiState.showNonSelectedAvatarAlert) {
+        nonSelectedAvatarAlertVisible = uiState.showNonSelectedAvatarAlert
+    }
```
AdamGrzybkowski: Would you mind explaining why we need this? I believe this is done so that the banner reappears when the state changes in the ViewModel. Is that it? 

I'm worried though, that we now have UI state in two places and we should avoid that.

Would it be possible to keep the state fully in ViewModel (including the dismissed state)?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -197,6 +204,13 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onEvent: (AvatarPickerEv
                     .fillMaxWidth()
                     .padding(bottom = 10.dp),
             )
+            AnimatedVisibility(nonSelectedAvatarAlertVisible) {
```
AdamGrzybkowski: I wonder if we could customize this animation. Currently when this banner gets hidden the whole UI is moved up, rather than the top of the bottom sheet just going down. Feels unnatural.

hamorillo: I've been playing with the `AnimatedVisibility` params and options. It's not easy to achieve, or at least it's not pretty straightforward. 

I agree it's nice to have, but I don't want to dedicate too much time to this now. What do you think about creating a ticket and returning to it when we have more "free time"?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
hamorillo: Thanks for pointing this out. 🙇  I agree that the approach wasn't correct. I've refactored it.

AdamGrzybkowski: Looks better now! I'm worried about the current logic though. Considering a scenario like this:

1. User deletes the currently selected Avatar
2. Alert banner appears
3. User dismisses the banner
4. User tries to delete another avatar, but the request fails
5. Alert banner reappears   -- this shouldn't happen
6. User dismisses the banner
7. User tries to delete another avatar again, this time successfully
8. Alert banner reappears   -- this shouldn't happen

Here's the recording showing alert banner reappearing:


https://github.com/user-attachments/assets/ac0ca506-f161-4014-8840-91850d52eede



I do believe we should respect the user's action to dismiss the banner for longer. I would propose to reset that user action only if the user went from `avatar selected` to `no avatar` state. Here's an updated scenario where the alert banner would reappear:

1. User deletes the currently selected Avatar
2. Alert banner appears
3. User dismisses the banner
4. User tries to delete another avatar, but the request fails
5. Alert banner DOESN'T reappear
7. User tries to delete another avatar again, this time successfully
8. Alert banner DOESN'T reappear
9. User selects an avatar
10. User deletes the currently selected Avatar
11. Alert banner appears - Only here

WDYT?

hamorillo: It makes sense. I've refactored that and added `DeleteAvatarAlertStatus` to make it clearer the current status of the AlertBanner.

AdamGrzybkowski: The behavior looks good now! 


💡 💡 💡 💡 

I do feel like we could simply the implementation and make it more future proof though. 

The Alert banner visibility is defined by two values `isAvatarSelected (emailAvatars.selectedAvatarId == null)` and `isBannerDismissed`. 
The visibility condition would look like this: `banner.visible = !isAvatarSelected && !isBannerDismissed`

The first value `isAvatarSelected` is already in the UI state so technically we don't need to do anything extra here.
The second one `isBannerDissmed` should change based on the behavior in two places:
1. When the users dismisses it manually it should be set to `true`
2. When the `isAvatarSelected` changes from `true` to `false` it should be reset to `false`

Point 1 is already handled in your implementation. Point 2 is to, kinda, but manually in each potential place where the `emailAvatars.selectedAvatarId` value could change. This is where I think we might end up with some bugs in the future as we could simply forget to update it in new places.

As I mentioned above we already have the `emailAvatars.selectedAvatarId` in the UI state so we could simply observe it and react when the value changes. Here's an example code: 

```
_uiState
            .filter { it.emailAvatars != null } //We don't care about this until we get this value
            .map { it.emailAvatars?.selectedAvatarId != null }
            .distinctUntilChanged()
            .onEach { isAvatarSelected ->
                if (isAvatarSelected) {
                    _uiState.update { currentState ->
                        currentState.copy(
                            nonSelectedAvatarAlertVisible = DeleteAvatarAlertStatus.HIDDEN,
                        )
                    }
                } else {
                    _uiState.update { currentState ->
                        currentState.copy(
                            nonSelectedAvatarAlertVisible = DeleteAvatarAlertStatus.VISIBLE,
                        )
                    }
                }
            }
            .launchIn(viewModelScope)
```

I've used `DeleteAvatarAlertStatus` from your implementation but not sure we need it. 

With the above code, we can get rid of the `shouldShowNonSelectedAvatarAlert()` function completely and the `isBannerDismissed` should be properly reset whenever the `emailAvatars.selectedAvatarId` value changes no matter what triggered it.  

Sorry, for not catching this earlier, but I was focused on the behavior rather than the code! Let me know what you think!

hamorillo: I was looking for a similar approach, but I couldn't come up with anything. Looks like I could have been more inspired.  

I've updated the code, removing the `DeleteAvatarAlertStatus`, which is totally unnecessary right now, and observing the flow to update the `nonSelectedAvatarAlertVisible`, as you suggested. 

Thanks for your review, @AdamGrzybkowski, I really appreciate it. I hope it looks better now.

AdamGrzybkowski (approved):
Thanks for all the updates @hamorillo!! Looks good now 🎉 

I believe, we need to switch the target branch, as this won't make it into 2.2.0, right?

hamorillo:
> we need to switch the target branch, as this won't make it into 2.2.0, right?

Yeah, as we've discussed, this should be in `2.3.0`. I've created [release/2.3.0](https://github.com/Automattic/Gravatar-SDK-Android/tree/release/2.3.0) from trunk and switched the target.

PR #463, opened 2024-11-27
Apply tint to verified services remote icons

### LINKED ISSUE #462: LinkedIn icon is black when using dark theme
When using Gravatar SDK by `LargeProfile` composable and default `LocalGravatarTheme` I've noticed that in dark mode, LinkedIn icon is black. It should match other icons and be white.

Image

AdamGrzybkowski:
Closes #462

### Description

There's an issue with icon tint for verified services that Icon are from a URL rather than a local resource. Those icons are not properly tinted in dark themes.

This PR solves that by applying a `LocalContentColor.current` to them.

| Before | After |
|---|---|
| Screenshot 2024-11-27 at 14 07 07 | Screenshot 2024-11-27 at 14 04 50 |
| Screenshot 2024-11-27 at 14 06 57 | Screenshot 2024-11-27 at 14 04 58 |

### Testing Steps

1. Add one of the services we don't have a local icon for to your profile: LinkedIn, BlueSky works.
2. Alternatively you can test other public profiles that have those services added.
3. Run the demo app.
4. On the profile tab test the components.
5. Repeat in the other UI mode (dark/light).

### DIFF
--- a/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/SocialMedia.kt
@@ -12,9 +12,11 @@ import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
+import androidx.compose.material3.LocalContentColor
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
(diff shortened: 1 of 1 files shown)


hamorillo (approved):
🚀 

As we plan to release `2.2.0` soon, maybe this is not a big deal, but should we consider releasing this fix also in `2.1.1`?

AdamGrzybkowski:
>As we plan to release 2.2.0 soon, maybe this is not a big deal, but should we consider releasing this fix also in 2.1.1?

I think 2.2.0 is ok. This is not a crash, only certain icons are affected and only on the dark theme so I think it is quite limited.

Issue #371, opened 2024-10-09
Rethink how to manage generated documentation


hamorillo:
[Discussed here](https://github.com/Automattic/Gravatar-SDK-Android/pull/369#issuecomment-2399855516)

PR #464, opened 2024-11-28
Remove docs from trunk

AdamGrzybkowski:
Part of #371

### Description

This part removes the generated API documentation from the `trunk` branch. 

To be more precise, I removed:
- `docs/dokka/` directory
- docs/index.html` file

Those docs are already available in a separate branch `docs` and that is where the gh pages are configured to build from. 

Next, I will work on the automation, to build those docs automatically on each tag push.

### Testing Steps

There is nothing to test really, you can check the `docs` folder and see if the gh pages are working.


hamorillo (approved):
🚀

PR #466, opened 2024-11-28
Docs: Fix redirection URL

hamorillo:
### Description

When using the https://automattic.github.io/Gravatar-SDK-Android link to access the documentation we are redirected to: `https://automattic.github.io/Gravatar-SDK-Android/dokka/current/index.html` which is the old path. Now, we should redirect to: https://automattic.github.io/Gravatar-SDK-Android/current/index.html

image

### Testing Steps

- Review changes (This won't work until it's merged)

**Note:** CI is failing in this PR as it tries to execute all the usual tasks... We can omit that for this PR.


AdamGrzybkowski:
Huh, sorry! I've just noticed that and fixed it directly :D Thanks for catching it though!!

PR #467, opened 2024-11-28
Create docs.yml workflow to generate API documentation

### LINKED ISSUE #371: Rethink how to manage generated documentation

AdamGrzybkowski:
Closes #371

### Description

This PR adds a `docs.yml` file that will be triggered on each "stable" tag push and will build the API documentation. 

This should be a fully automated process now so the release steps have been updated accordingly. 

I did test this workflow first on branch pushes and a different docs branch - you can the commits generated here https://github.com/Automattic/Gravatar-SDK-Android/commits/adam/only_docs/. Note that they have the wrong version in the commit due to some issues with `git describe` - but we don't use this here, it was only for testing. 

### Testing Steps

It's worth testing that the `sdkVersion` is properly set in the QE, so run it and inspect the network requests.

As for the workflow itself - not sure. I tested it as mentioned above, so now, I guess we will see once we create a new release :D


.github/workflows/docs.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: ./gradlew dokkaHtmlMultiModule
+      - name: Update Docs Branch
+        run: |
+          git checkout docs
+          cp -R docs/dokka/history .
+          cp -R docs/dokka/current .
+          git add history
+          git add current
+          git config user.name github-actions
+          git config user.email github-actions@github.com
```
AdamGrzybkowski, line 50: Is that ok? Do we have some other values we could use here?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,47 @@
...
+          distribution: 'temurin'
+          java-version: '17'
+      - name: Copy docs to docs/dokka directory
+        run: |
+          git fetch origin docs:docs
+          git checkout docs
+          mkdir -p docs/dokka || true
+          cp -R current/ docs/dokka/current
+          cp -R history/ docs/dokka/history
+          git checkout ${{ github.ref_name }}
```
AdamGrzybkowski: Checking out to the tag to build the docs from it, rather than trunk.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,47 @@
+name: Build and Deploy API Documentation
+
+on:
+  push:
+    tags:
+      # Match 1.0.0 but not 1.0.0-rc1 or other suffixes
+      - '[0-9]+\.[0-9]+\.[0-9]'
+      - '![0-9]+\.[0-9]+\.[0-9]-[a-z]*'
+      - '![0-9]+\.[0-9]+\.[0-9]-[A-Z]*'
```
AdamGrzybkowski: One regex like this `^[0-9]+\.[0-9]+\.[0-9]+$` would work but I'm not sure if we can use special characters that are not listed here https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

.github/workflows/docs.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+          git checkout docs
+          cp -R docs/dokka/history .
+          cp -R docs/dokka/current .
+          git add history
+          git add current
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git commit -m "Update API documentation for ${{ github.ref_name }}"
+          git push origin docs
```
AdamGrzybkowski, line 52: I could technically deploy to `gh-pages` at the end - but not sure this adds any value when we can publish the docs from the `docs` branch as well.

.github/workflows/docs.yml
hamorillo: ❓ Why do we need the following?

```
- '![0-9]+\.[0-9]+\.[0-9]-[a-z]*'
- '![0-9]+\.[0-9]+\.[0-9]-[A-Z]*'
```

💡 Semantic versioning website has some [suggested regex](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string). Those validate all the possibilities of SemVer, so we can't apply them directly. However, I like the fact that they validate errors like: `01.09.01`. However, after checking the documentation, I'm not sure if the `or` is available on GitHub.

What we can do, at least, is limit mayor to 1>=, as we know we won't release anything as (`0.X.X`). I think, we also need to allow patch version 9>, something like:

`'[1-9]+\.[0-9]+\.[0-9]+'`

.github/workflows/docs.yml
hamorillo: That's OK in my opinion.

gravatar/build.gradle.kts
```diff
@@ -11,9 +11,7 @@ plugins {
     alias(libs.plugins.dokka)
 }
 
-val sdkVersion = providers.exec {
-    commandLine("git", "describe", "--tags", "--abbrev=0")
-}.standardOutput.asText.get().trim()
+val sdkVersion: String by rootProject.extra
```
hamorillo, line 15: It's not a problem with this PR because we did the same before, but we are sending the previous version while developing the next iteration. It's not a big deal but wdyt about creating a ticket with low priority to improve this?

.github/workflows/docs.yml
hamorillo: 🤷‍♂️ I don't have an answer. In fact, I have a question just out of curiosity. Can we use any mail there?

.github/workflows/docs.yml
AdamGrzybkowski: >❓ Why do we need the following?

These are to prevent running that workflow on anything like: `1.1.1-rc1` or `1.1.1-ALPHA1`. We removed docs for versions like that recently so I thought we don't want to have docs for those.

>'[1-9]+\.[0-9]+\.[0-9]+'

Sounds good to me. I wasn't too worried about using 0 as a major. Technically this is correct, and if that error would happen on our side we could just remove the last commit. But yeah, at this point, 0 as major, shouldn't be used.

gravatar/build.gradle.kts
AdamGrzybkowski: Hmmm would it be ok if the version during progress be sth like `2.1.0-{hash}`? Where 2.1.0 is the current version? That should be simple and likely only require removing the `--abbrev=0` part.

.github/workflows/docs.yml
hamorillo: But `1.1.1-rc1` or `1.1.1-ALPHA1` don't match with the regex `[1-9]+.[0-9]+.[0-9]+`, right?

gravatar/build.gradle.kts
hamorillo: 👍 Sounds good to me.

.github/workflows/docs.yml
AdamGrzybkowski: They would I think. This `^[1-9]+.[0-9]+.[0-9]+$` with the`^` and `$` would work better, but like I said, I'm not sure we can use all the characters as they are not mentioned in the docs.

gravatar/build.gradle.kts
AdamGrzybkowski: Done

.github/workflows/docs.yml
hamorillo: If they look for a full match in the tag (which makes sense to me), it shouldn't work:

image

Anyway, it won't hurt those lines, so feel free to keep them.

hamorillo (approved):
🚀

.github/workflows/docs.yml
AdamGrzybkowski: Right, so I did test it by pushing some tags and indeed they do a full match. I have updated the workflow to use `[1-9]+.[0-9]+.[0-9]+`.

Thanks!

.github/workflows/docs.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+      - name: Copy docs to docs/dokka directory
+        run: |
+          git fetch origin docs:docs
+          git checkout docs
+          mkdir -p docs/dokka || true
+          cp -R current/ docs/dokka/current
+          cp -R history/ docs/dokka/history
```
wzieba: If you want, you can use `actions/checkout` for these checkouts as well

```
- uses: actions/checkout@v4
  with:
    ref: 'docs'
    path: 'docs/dokka'
```

which might simplify the setup (I haven't tested it but, something like the suggestion above)

.github/workflows/docs.yml
```diff
@@ -0,0 +1,45 @@
...
+          git checkout ${{ github.ref_name }}
```
wzieba: and then another checkout action:

```
- uses: actions/checkout@v4
  with:
    ref: ${{ github.ref_name }}
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,45 @@
...
+  generate_docs:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
wzieba: As we perform checkouts/clones in later steps, maybe this is not necessary?

.github/workflows/docs.yml
wzieba: LGTM - you can also use wpmobilebot account, but this can be any mail or username

wzieba (approved):
Looks good, I shared some suggestions that might simplify the workflow file.

Thanks also for applying fix for not keeping the version static in repository!

.github/workflows/docs.yml
AdamGrzybkowski: Hmm I have to test that. I think that would require some extra modification to how I copy the docs between the branch switches. Currently, I'm using git uncommitted changes and work within the same directory, but the suggestion would use a completely separate folder for that right?

AdamGrzybkowski: The workflow was updated. Now we will checkout the latest tag at `code/` directory and the `docs` branch at `docs/`. Docs are then copied to the `code/docs/dokka/` for the Dokka generation and once generated moved back to the `docs/`. 

This workflow no longer relies on git uncommitted changes so I believe it should be less error prone.

.github/workflows/docs.yml
AdamGrzybkowski: Done

.github/workflows/docs.yml
AdamGrzybkowski: It's not needed, removed.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: Checkout the branch with the latest code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.ref_name }}
+          path: 'code'
+      - name: Checkout the branch with the docs
+        uses: actions/checkout@v4
+        with:
+          ref: 'adam/only_docs'
```
hamorillo: You forgot to replace your test branch:

```suggestion
          ref: 'docs'
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,52 @@
...
+          cp -R code/docs/dokka/current docs/
+      - name: Push updated docs
+        working-directory: ./docs
+        run: |
+          git add history
+          git add current
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git commit -m "Update API documentation for ${{ github.ref_name }}"
+          git push origin adam/only_docs
```
hamorillo: You forgot to replace your test branch:

```suggestion
          git push origin docs
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  generate_docs:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout the branch with the latest code
+        uses: actions/checkout@v3
```
hamorillo: Should we use the v4 here as well?
```suggestion
        uses: actions/checkout@v4
```

hamorillo (changes requested):
Great improvement! I left some comments.

.github/workflows/docs.yml
AdamGrzybkowski: 🤦 Yes, thanks!

hamorillo (approved):
🚀

PR #472, opened 2024-11-29
Remove Alt text option from Avatar's contextual menu

AdamGrzybkowski:
### Description

This clears the unused option. It was added temporarily, but we're planning to make a 2.2.0 release with just the `download` 
 and `delete` option. 
 
 Should we remove the icon and string resources as well? 
 
 
Screenshot 2024-11-29 at 14 59 47

### Testing Steps

Run the QE and confirm `Alt text` option is not available.


AdamGrzybkowski:
>Should we remove the icon and string resources as well?

~Lint answered that question.~

EDIT: I ignored Lint for those resources, so we can easily add them later (especially the strings with translations).

hamorillo (approved):
🚀

PR #475, opened 2024-12-03
QuickEditor: Make ColorScheme configurable via GravatarColorScheme param

### LINKED ISSUE #473: QuickEditor: Allow third-party app specify light/dark/system mode

AdamGrzybkowski:
Closes #473 

### Description

This PR adds an additional param `GravatarColorScheme` that allows forcing Light or Dark color schemes within the Quick Editor. 

The passed param is used to set a proper `Configuration.uiMode` that's provided via `CompositionLocalProvider`. 
Thanks @hamorillo for the tip.

https://github.com/user-attachments/assets/f1672fe3-88aa-443f-b9c7-13a402980f5d

### Testing Steps

1. Start the demo app
2. Select a different ColorScheme than is currently set system-wise
3. Launch the QE
4. Confirm the color scheme is correct
5. Repeat with different color scheme

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarColorScheme.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarColorScheme.kt
@@ -1,0 +1,10 @@
+package com.gravatar.quickeditor.ui.editor
+
+/**
+ * The color scheme to be used in the Quick Editor.
+ */
+public enum class GravatarColorScheme {
+    LIGHT,
+    DARK,
+    SYSTEM,
+}
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
@@ -1,5 +1,6 @@
 package com.gravatar.quickeditor.ui.editor.bottomsheet
 
+import android.content.res.Configuration
 import android.graphics.Color
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
@@ -25,6 +26,7 @@ import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.platform.LocalUriHandler
 import androidx.compose.ui.unit.dp
 import androidx.core.view.WindowInsetsControllerCompat
@@ -43,6 +45,7 @@ import com.gravatar.quickeditor.ui.components.QEDragHandle
 import com.gravatar.quickeditor.ui.components.QETopBar
 import com.gravatar.quickeditor.ui.editor.AuthenticationMethod
 import com.gravatar.quickeditor.ui.editor.AvatarPickerContentLayout
+import com.gravatar.quickeditor.ui.editor.GravatarColorScheme
 import com.gravatar.quickeditor.ui.editor.GravatarQuickEditorDismissReason
 import com.gravatar.quickeditor.ui.editor.GravatarQuickEditorPage
 import com.gravatar.quickeditor.ui.editor.GravatarQuickEditorParams
@@ -93,6 +96,7 @@ internal fun GravatarQuickEditorBottomSheet(
         GravatarModalBottomSheet(
             onDismiss = onDismiss,
             modalBottomSheetState = modalBottomSheetState,
+            colorScheme = gravatarQuickEditorParams.colorScheme,
         ) {
             when (authenticationMethod) {
                 is AuthenticationMethod.Bearer -> {
@@ -120,6 +124,7 @@ internal fun GravatarQuickEditorBottomSheet(
 @Composable
 private fun GravatarModalBottomSheet(
     onDismiss: (dismissReason: GravatarQuickEditorDismissReason) -> Unit = {},
+    colorScheme: GravatarColorScheme,
     modalBottomSheetState: ModalBottomSheetState,
     content: @Composable () -> Unit,
 ) {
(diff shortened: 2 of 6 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
```diff
@@ -132,53 +137,73 @@ private fun GravatarModalBottomSheet(
...
-                        !isDarkTheme
-                }
-                Surface(
-                    modifier = Modifier
-                        .fillMaxWidth(),
-                    tonalElevation = 1.dp,
+                // Modal content must be taking the uiMode from Activity and doesn't respect the above set CompositionLocalProvider
+                CompositionLocalProvider(
+                    LocalConfiguration provides configuration,
                 ) {
```
AdamGrzybkowski, line 158: If I only keep this here, without the one at the top the `QETopBar` and the navigation bars are not properly themed. If I remove this one here, the content of the BottomSheet is not themed. 

That's why we need it in both places. To be honest I'm quite puzzled why the `QETopBar` is not themed because it's part of the content of the BottomSheet...

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarColorScheme.kt
```diff
@@ -0,0 +1,7 @@
+package com.gravatar.quickeditor.ui.editor
+
+public enum class GravatarColorScheme {
```
AdamGrzybkowski: I'm open about the naming, so let me know if you think there's a better one.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
```diff
@@ -132,53 +137,74 @@ private fun GravatarModalBottomSheet(
...
-                    WindowInsetsControllerCompat(window, window.decorView).isAppearanceLightNavigationBars =
-                        !isDarkTheme
-                }
-                Surface(
-                    modifier = Modifier
-                        .fillMaxWidth(),
-                    tonalElevation = 1.dp,
+                // Modal content must be taking the uiMode from Activity and doesn't respect
+                // the above set CompositionLocalProvider
+                CompositionLocalProvider(
```
hamorillo, line 156: I've been investigating why we need to add this again here. 

The [ModalBottomSheet](https://github.com/composablehorizons/compose-unstyled/blob/c62bab5babdabceb634ec66de05f5370c161b66d/core/src/commonMain/kotlin/ModalBottomSheet.kt#L171) pass the content to another composable; [Modal](https://github.com/composablehorizons/compose-unstyled/blob/c62bab5babdabceb634ec66de05f5370c161b66d/core/src/androidMain/kotlin/Modal.android.kt#L40) which is using the legacy view system under the hood. 

I think we are losing the configuration that was already provided at that point. That is why we need to set the `LocalConfiguration` inside the `ModalBottomSheet` `content` too.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
hamorillo: 🤯 I've been trying to understand what happens there. No idea. I'll try again tomorrow with the brain fresh.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarColorScheme.kt
hamorillo: ⚠️ I don't like `ColorScheme` because it can be easily confused with [material/ColorScheme](https://developer.android.com/reference/kotlin/androidx/compose/material3/ColorScheme).

What about:

```suggestion
public enum class GravatarUiMode {
```

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
hamorillo: Also, mention that this is a pain because it forces us to add the `colorScheme: GravatarColorScheme` param to the composable instead of providing a `LocalConfiguration` as third-party apps can do to override the UI components style.

hamorillo (commented):
Works great! 

While reviewing this PR I've noticed [Compose-unstyled 1.20.0 has been released](https://github.com/composablehorizons/compose-unstyled/releases/tag/1.20.0). (We are currently using `1.15.0`). 

I've seen some changes in the classes we are using. Maybe we can bump the version in a different PR.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
AdamGrzybkowski: Yes, that is true. The thing I don't get is why I need the outer `CompositionLocalProvider`, because technically, all the content is wrapped with the inner one already.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarColorScheme.kt
AdamGrzybkowski: Makes sense! I will update it.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
hamorillo: Yeah, same here. I don't understand it.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarColorScheme.kt
AdamGrzybkowski: Done.

hamorillo (approved):
🚀

PR #476, opened 2024-12-03
Switch serialization library from `moshi-reflection` to `moshi-codegen`

### LINKED ISSUE #480: Switch from moshi-reflection to moshi-codegen
We want to switch from [moshi-reflection](https://github.com/square/moshi?tab=readme-ov-file#codegen) to [moshi-codegen](https://github.com/square/moshi?tab=readme-ov-file#codegen) to not increase the final APK size.

hamorillo:
Closes #480

### Description

Our SDK uses [Moshi](https://href.li/?https://github.com/square/moshi) for serialization/deserialization duties. The problem is that we were using the [reflection](https://href.li/?https://github.com/square/moshi?tab=readme-ov-file#reflection) version of Moshi which, as the documentation mention:

> The reflection adapter transitively depends on the kotlin-reflect library which is a 2.5 MiB .jar file.

That makes the final APK notably increase its size just by adding our library as a dependency. 

Moshi documentation also specifies:

> prefer reflection to convert both private and protected properties

As we don't need that, we can switch the [codegen](https://href.li/?https://github.com/square/moshi?tab=readme-ov-file#codegen) version and remove the size increase in the final APK. Also, Moshi documentation says:

> Prefer codegen for better performance and to avoid the kotlin-reflect dependency

### Testing Steps

Everything should work exactly as before.

- Smoke test all tabs of the `demo-app`.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -17,6 +17,7 @@ plugins {
     // Roborazzi
     alias(libs.plugins.roborazzi) apply false
     alias(libs.plugins.binary.compatibility.validator)
+    alias(libs.plugins.ksp) apply false
 }
 
 buildscript {
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -17,6 +17,7 @@ dokka = "1.9.20"
 junit = "4.13.2"
 kotlin = "1.9.25"
 kotlinCoroutines = "1.7.3"
+ksp = "1.9.25-1.0.20"
 ktlint = "12.1.0"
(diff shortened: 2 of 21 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

AdamGrzybkowski (approved):
:shipit:

PR #477, opened 2024-12-03
QuickEditor: Remove delay before notifying the avatar change

### LINKED ISSUE #474: QuickEditor: Remove delay after avatar change
The backend has been updated, and we don't need the delay anymore.

hamorillo:
Closes #474

### Description

The backend has been updated, and we don't need to wait any longer before notifying the avatar switch. The avatar switch is immediate now. We are ready to remove the delay introduced in #350. More concretely in this [commit](https://github.com/Automattic/Gravatar-SDK-Android/pull/350/commits/1aca9c43b7f134334f27aaa248bf14d3ea8f0782).

### Testing Steps

- Open the QE and switch between avatars. Verify everything works smoothly; you can see the new avatar in the profile card.


AdamGrzybkowski (approved):
Thanks for keeping an eye on that! With each new selection, the Avatar was properly refreshed!

:shipit:

PR #479, opened 2024-12-04
Notify 3rd party when selected avatar is deleted

### LINKED ISSUE #478: Notify third-party app when selected avatar was removed.
When the selected avatar is deleted, the third-party app should be notified to refresh the Avatar.

AdamGrzybkowski:
Closes #478

### Description

When the selected avatar is deleted we should notify the third-party app about that change. 

This PR fixes that. 

| Before | After |
|---|---|
| notify_before | notify_after |

### Testing Steps
1. Run the QE
2. Select an avatar
3. Delete that avatar
4. Confirm the Avatar was updated in the Demo app


hamorillo (approved):
Good catch! Works great! 🚀

PR #482, opened 2024-12-04
QuickEditor: Fail silently when we get a 404 during avatar deletion

### LINKED ISSUE #460: Handle 404 when deleting an avatar
When deleting an avatar, if we get a 404 error, I'd expect to fail silently and keep the image deleted.

hamorillo:
Closes #460 

### Description

When we receive a 404 when deleting an avatar, we can assume that it has already been deleted, so we can continue as if the deletion had worked.

With this approach, we are improving the UX the final user feels.

### Testing Steps

- Open the QE and a browser with the same Gravatar account
- In the browser, delete an avatar
- In the QE, try to delete the same avatar
- Verify that the avatar is removed from the list as expected (You can also verify that the request has finished with a 404)
- Repeat the previous steps with the selected avatar and verify the avatar is removed from the list and the profile card refreshed.
- Verify you can delete as expected, non already deleted avatars.


AdamGrzybkowski (approved):
Great addition! :shipit:

PR #484, opened 2024-12-10
QuickEditor: Modify the way PickerPopup calculates its position

### LINKED ISSUE #457: QuickEditor: Extract AvatarMoreOptionsPickerPopup from SelectableAvatar
Discussion https://github.com/Automattic/Gravatar-SDK-Android/pull/454#discussion_r1853638756

AdamGrzybkowski:
Closes #457

### Description

The issue mentions that we should extract the `AvatarMoreOptionsPickerPopup` from `SelectableAvatar`. The root of that idea was all the arguments we had to pass for the popup position calculation. 
While working on #237 I noticed the issue and how it could be changed. This PR is a result of that. 

In short, we now use a version of `PopupPositionProvider`, it is a modified version of the code from the Composable Unstyled to make sure we prioritize showing the Popup above the anchor not below it (the original is for the Dropdown component). To make it work I had to also remove the `Dialog` from the `PickerPopup` as we need to have it in the same Box parent as the anchor, so we can't do the absolute positioning like before.  

Removing the `Dialog` means we no longer have the dimmed background, but this is not something that is in the spec for Menu components in Material 3, so this should be ok. 

With the updated code we can now have the button together with the Popup without any extra arguments like `anchorbounds` or `maxDrawingArea`.  Thanks to this we could have the `UploadImageButton` wrapped with the Popup and placed wherever we need without duplicating the Popup code. 

### Testing Steps
Test the `AvatarMoreOptionsPickerPopup` and the `MediaPickerPopup` in all possible scenarios: 
- horizontal, 
- vertical (empty list and non-empty).

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/PickerPopup.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/PickerPopup.kt
@@ -141,12 +141,14 @@ internal data class PickerPopupPositionProvider(
             IntOffset(offset.x.toPx().toInt(), offset.y.toPx().toInt())
         }
 
+        val displayPadding = with(density) { 16.dp.toPx().toInt() }
+
         // Compute horizontal position.
         val toRight = anchorBounds.left
         val toLeft = anchorBounds.right - popupContentSize.width
 
-        val toDisplayRight = windowSize.width - popupContentSize.width
-        val toDisplayLeft = 0
+        val toDisplayRight = windowSize.width - popupContentSize.width - displayPadding
+        val toDisplayLeft = 0 + displayPadding
 
         val x = (
             if (alignment == Alignment.Start) {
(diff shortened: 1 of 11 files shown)


hamorillo (approved):
Great work, @AdamGrzybkowski! 

Everything looks great in the vertical layout and the upload image menu. 

However, in the horizontal layout, I think the following cases look weird:

imageimage

I know those are corner cases, so maybe we can merge this PR and create a ticket to improve those. Wdyt?

AdamGrzybkowski:
Yes, this is also the behavior from the `DropdownMenu`. When the anchor view is outside of the screen it will anchor to the edge of the screen: left or right. I will check if there's a smart way to use parent view edges instead without passing those values.

AdamGrzybkowski:
@hamorillo I could only set a fixed display padding, let me know what you think about this.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/PickerPopup.kt
```diff
@@ -141,12 +141,14 @@ internal data class PickerPopupPositionProvider(
             IntOffset(offset.x.toPx().toInt(), offset.y.toPx().toInt())
         }
 
+        val displayPadding = with(density) { 16.dp.toPx().toInt() }
```
hamorillo: ❓ What do you think about passing the padding as a parameter?

hamorillo (approved):
Looks great! 🤩  Thanks for the improvement!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/PickerPopup.kt
AdamGrzybkowski: Done. I kept the default value for simplicity.

PR #487, opened 2024-12-11
Rename imageId to avatarId in AvatarService to match with other functions

AdamGrzybkowski:
### Description

I've noticed the name `imageId` in the `AvatarService` delete function is inconsistent with other functions. This hasn't been released yet, so it's the last time to change it. 

This is not something that would be caught by the binary check plugin, but if the named param was used this would be a breaking change. 

### Testing Steps

CI should be enough, smoke test Avatar deletion in the QE.


hamorillo (approved):
Nice catch! 👏 

It's a pity that other methods use `avatarId` because I think it looks better with `imageId` as that's what we receive in the `Avatar` model class. Said that, I think it makes sense to modify this for coherence with the other method. 

Using `imageId` could be a task for a future `3.0.0` 🤷 .

Thanks for taking care of this.

AdamGrzybkowski:
>It's a pity that other methods use avatarId because I think it looks better with imageId as that's what we receive in the Avatar model class.

Yes, I agree. It's just that now we have a total mix, so I think this is slightly better.

PR #488, opened 2024-12-12
QuickEditor: Implement the Avatar's rating change in the AvatarPicker

### LINKED ISSUE #237: QuickEditor: Update the rating of the Avatar

AdamGrzybkowski:
Closes #237

### Description

The 2nd PR integrates the API request with the Avatar rating changes coming from the UI. 

I've made the code in a way that will be easily expandable by the next `altText` modification without adding much code in the ViewModel, since those are handled by one API request. 

https://github.com/user-attachments/assets/e103d589-fdc4-4de3-8f89-8c6a83fb8e4b

### Testing Steps
1. Run the QE
2. Tap on the Avatar's `...` button and then `Change rating`
3. Tap a different one that's not selected
4. Confirm the Snack was shown informing about successful change
5. Tap on the Avatar's `...` button and then `Change rating`
6. Confirm the rating from step 3 is selected
7. Mock the PATCH network request so it fails
8. Tap on the Avatar's `...` button and then `Change rating`
9. Tap a different one that's not selected
10. Confirm the Snack was shown informing about an error
11. Tap on the Avatar's `...` button and then `Change rating`
12.. Confirm the rating is still the same as it was before step 9

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
@@ -77,7 +77,47 @@ internal class AvatarPickerViewModel(
             AvatarPickerEvent.DownloadManagerDisabledDialogDismissed -> hideDownloadManagerAlert()
             is AvatarPickerEvent.AvatarDeleteSelected -> deleteAvatar(event.avatarId)
             AvatarPickerEvent.AvatarDeleteAlertDismissed -> hideNonSelectedAvatarAlert()
-            is AvatarPickerEvent.AvatarRatingSelected -> Unit
+            is AvatarPickerEvent.AvatarRatingSelected -> updateAvatar(event.avatarId, event.rating)
+        }
+    }
+
+    private fun updateAvatar(avatarId: String, rating: Avatar.Rating? = null, altText: String? = null) {
+        viewModelScope.launch {
+            val oldAvatar: Avatar? = _uiState.value.emailAvatars?.avatars?.find { it.imageId == avatarId }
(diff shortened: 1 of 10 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -77,7 +77,47 @@ internal class AvatarPickerViewModel(
...
             AvatarPickerEvent.AvatarDeleteAlertDismissed -> hideNonSelectedAvatarAlert()
-            is AvatarPickerEvent.AvatarRatingSelected -> Unit
+            is AvatarPickerEvent.AvatarRatingSelected -> updateAvatar(event.avatarId, event.rating)
+        }
+    }
+
+    private fun updateAvatar(avatarId: String, rating: Avatar.Rating? = null, altText: String? = null) {
+        viewModelScope.launch {
+            val oldAvatar: Avatar? = _uiState.value.emailAvatars?.avatars?.find { it.imageId == avatarId }
+            val updateType = AvatarUpdateType.RATING
```
hamorillo, line 87: We need to handle the `updateType` when adding the `altText` but we'll do it in its own PR. :)

hamorillo (approved):
Great work! 🚀

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
AdamGrzybkowski: Yes, I had it here, but I noticed I was going out of the scope of this PR, so I removed it.

PR #489, opened 2024-12-12
QuickEditor: Fix the animation when showing/hiding AlertBanner

### LINKED ISSUE #461: QuickEditor: Improve no-avatar selected banner exit animation
https://github.com/Automattic/Gravatar-SDK-Android/pull/458#discussion_r1853948678

AdamGrzybkowski:
Closes #461

### Description

The AvatarPicker's size would weirdly animate when showing or hiding the alert banner. This was cause by the `animateContentSize()` modifier on the top composable. We need this content animation when switching between loading/error/success states, but we can move this modifier lower where that content change is actually happening. 

With the top `animateContentSize()` removed the appearance of the alert banner properly animates without any extra jumps.

| Before | After |
|---|---|
| animation_before | animation_after |

### Testing Steps

1. Open the QE
2. Confirm the switch from loading to success animated
3. Delete the currently selected Avatar
4. Confirm the alert banner appeared nicely without extra movement of the whole bottom sheet
5. Select an avatar
6. Confirm the alert banner disappeared nicely without extra movement of the whole bottom sheet


hamorillo (approved):
Thanks for taking care of this. 🤩 🚀

PR #492, opened 2024-12-16
Do not fire API call when rating and altText unchanged

### LINKED ISSUE #491: QuickEditor: Do not make API request when an existing Avatar rating is tapped

AdamGrzybkowski:
Closes #491 

### Description

We don't need to fire the PATCH API call when the rating and the alt text are the same. This PR implements an early check to return before making that call. 

### Testing Steps

1. Run the QE
2. Open the Avatar's menu and tap `Change rating`
3. Tap the already selected rating
4. Confirm no confirmation Snack is shown


hamorillo (approved):
Nice improvement! 🚀

PR #493, opened 2024-12-16
Release/2.3.0

AdamGrzybkowski:
### Description

2.2.0 was shipped so merging the 2.3.0 to trunk.

### Testing Steps

Test around avatar deletion flows. We had updates to it in both `trunk` and the `relese/2.3.0` branches. General smoke testing is welcome.


hamorillo (approved):
🚀

Issue #404, opened 2024-10-21
QuickEditor: Update the ALT text for avatar


AdamGrzybkowski:
![Image](https://github.com/user-attachments/assets/b5e41d9a-5c7c-42d3-a65f-07b4e50689e5)

Issue #499, opened 2024-12-17
Quick Editor: Alt Text - Improve behaviour on the Alt text input text


hamorillo:
The behavior of the placeholder can be improved. 

**Reminder:** We can't use experimental composables.

AdamGrzybkowski:
@hamorillo I wonder if we need this. There's always a default alt text set, so the only way to show the placeholder is to delete the text manually. At that point, the placeholder doesn't provide much value.

hamorillo:
Yeah, I agree. With the new default `altText`, the placeholder doesn't add much value. We can reduce the priority to `low` (just in case, at some point, we have time) or remove the ticket.

PR #501, opened 2024-12-17
Quick Editor - Alt text feature foundations 

### LINKED ISSUE #497: Quick Editor: Alt Text - Base implementation of the Alt Text functionality

hamorillo:
Closes #497 

### Description

This PR implements the foundations of the `Alt Text` feature. It doesn't include all of the functionality, as some parts still need to be implemented/improved. [See sub-issues](https://github.com/Automattic/Gravatar-SDK-Android/issues/404)

I'm opening this PR for two reasons:
- I think we can parallelize the missing work
- To keep this PR as small as possible

The basic functionality, which is changing the alternative text for an image, is implemented. 
You should be able to see the confirmation toast and the error toast when something fails.

image

### Testing Steps

1. Open the QE and the options for any avatar
2. Modify the alt text
3. Verify the `Save` is enabled when the value is different from the original alt text
4. Save a new alt text
5. Verify the QE navigates to the Avatar picker
6. Close the QE bottom sheet and reopen it
7. Open the alt text for the previous image and verify you see the new alt text
8. Try to change it again but force an error (network error is enough)
9. Verify you see the error toast
10. Open the alt text again and verify you see the previous alt text (not the one that produces the error)
11. Verify dark/light mode
12. Verify orientation changes


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AltTextSection.kt
```diff
@@ -0,0 +1,174 @@
...
+                                RoundedCornerShape(cornerRadius),
+                            )
+                            .clip(RoundedCornerShape(cornerRadius)),
+                    )
+                    val placeholder = stringResource(R.string.gravatar_qe_avatar_alt_text_section_placeholder)
+                    /*
+                     * TextField is still Experimental so using this workaround to avoid its use.
+                     * This needs to be improved.
+                     */
+                    BasicTextField(
```
hamorillo: `TextField` is experimental, so we need to use `BasicTextField`. The major problem with this composable is the "placeholder". It enabled the placeholder by using:

```
decorator = TextFieldDefaults.TextFieldDecorationBox(
```

Which is also experimental, so this is a pain for us.

I've implemented a workaround but also created a [task](https://github.com/Automattic/Gravatar-SDK-Android/issues/499) to improve this.

Current behavior:

https://github.com/user-attachments/assets/e57aa03a-90f9-4e5a-85be-9dd81f5cf450

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AltTextSection.kt
```diff
@@ -1,19 +1,138 @@
...
+                        } else {
+                            LocalTextStyle.current
+                        },
+                        modifier = Modifier
+                            .fillMaxWidth()
+                            .padding(start = 16.dp, top = 8.dp, bottom = 8.dp),
+                    )
+                }
+                QEButton(
+                    buttonText = "Save",
```
AdamGrzybkowski: 👀

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
```diff
@@ -205,11 +206,13 @@ internal fun AvatarPicker(uiState: AvatarPickerUiState, onEvent: (AvatarPickerEv
...
-                    profile = uiState.profile,
-                    modifier = Modifier.padding(horizontal = 16.dp),
-                )
+            if (uiState.mainSection is MainContentSection.AvatarsSectionUiState) {
+                key(uiState.avatarUpdates) {
+                    ProfileCard(
+                        profile = uiState.profile,
+                        modifier = Modifier.padding(horizontal = 16.dp),
+                    )
+                }
```
AdamGrzybkowski: ❓ Have you considered creating a new page for the `Alt text` functionality? We could use [nested graphs](https://developer.android.com/guide/navigation/design/nested-graphs#compose) for this. Although it might look more time-consuming, I think it could lead to:
1. Simplified UI code. There wouldn't be a need for several checks like this to show/hide section.
2. Potentially simpler integration with the `Close/Cancel` button from the `TopBar`. This is a very top-level component, so having a nested destination would simplify working with it as it's closer. 

If you prefer to keep it like this, maybe we should wrap the whole AvatarPicker (card and the section) and only do one of those checks (if section) in one place. This would also simplify any type of enter/exit animations I think.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AltTextSection.kt
AdamGrzybkowski: Hmm this doesn't work at all like a placeholder, right? I mean, when I tap on the placeholder it will place the cursor wherever I tap not only in the beginning.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AltTextSection.kt
```diff
@@ -0,0 +1,174 @@
...
+
+@Composable
+internal fun AltTextSection(
+    state: MainContentSection.AltTextUiState,
+    onAltTextChanged: (String) -> Unit,
+    onWhatIsAltTextClicked: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val sectionPadding = 16.dp
+    var altText by remember { mutableStateOf(state.avatar.altText) }
```
AdamGrzybkowski: ⚠️ We should at least use `rememberSavable` here, I've lost the text on the screen orientation change.

AdamGrzybkowski: Ideally, we should have this in the viewModel, so we can use this for any logic that depends on it (max limit for example).

AdamGrzybkowski (commented):
I've left some comments/ideas that I think are worth discussing. Additionally, I think we need to handle the system back button to hide the Alt text section only, not the whole bottom sheet. There's no way to exit this state now without saving the new Alt text.

AdamGrzybkowski:
One more and this could be left as a subtask for further improvements, but I think it would be great to set the max height of the input field and make the content of it scrollable in a way the the `Save` button would remain visible even with a longer text.

Screenshot 2024-12-17 at 14 10 29

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AltTextSection.kt
hamorillo: Exactly, that's the problem. If you prefer, we can remove this and work on it in the subtask.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
hamorillo: I'll take a look at the nested graphs. Thanks for the suggestion.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AltTextSection.kt
AdamGrzybkowski: Yeah, I think this is no better than no placeholder so maybe makes sense to not have it.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AltTextSection.kt
hamorillo: 🤔 Why did nothing fail in CI? I'll check.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
hamorillo: I've done a lot of refactoring in this PR using a nested graph. As you mentioned, the code looks cleaner (at least in my opinion). 

> Potentially simpler integration with the Close/Cancel button from the TopBar.

This still will cause some "problems" so we'll need to figure out how to solve this. One option could be to move the TopBar inside each section, but I haven't thought deeply about this. 

Anyway, let me know what you think about the current approach. 

I think we can still do another step and create a different ViewModel for the AltText section. Extracting the code related to the AltText should be "easy", the problem here is that we'll need to also make a way to sync the data between them. We can iterate over this in the following PRs.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/AltTextSection.kt
hamorillo: Everything is in the viewmodel now.

hamorillo:
> I think it would be great to set the max height of the input field and make the content of it scrollable in a way the the Save button would remain visible even with a longer text.

We have an issue to limit the number of characters #502, when we implement that feature, I would say this won't be a huge problem, but we'll see.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextSection.kt
```diff
@@ -0,0 +1,238 @@
...
+
+@Composable
+internal fun AltTextSection(
+    onBackPressed: () -> Unit,
+    viewModel: AvatarPickerViewModel,
+    modifier: Modifier = Modifier,
+) {
+    BackHandler {
+        onBackPressed()
+    }
```
hamorillo: ⚠️ This works well before the keyboard is opened. However, after that, the bottom sheet is dismissed instead of being handled here. If that's OK with you, I'll create a ticket for this.

AdamGrzybkowski: That's weird... we can ticket this.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -49,21 +65,42 @@ internal fun GravatarQuickEditorPage(
...
+            }
+            composable(
+                route = EditorNavDestinations.ALT_TEXT.name,
+            ) {
+                AltTextSection(
+                    onBackPressed = { navController.popBackStack() },
+                    viewModel = editorViewModel,
+                    modifier = Modifier.padding(16.dp),
+                )
+            }
```
AdamGrzybkowski: 💡  Could we write an extension function for this to not repeat it twice? Something like: `NavGraphBuilder.avatarPickerGraph()`

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -49,21 +65,42 @@ internal fun GravatarQuickEditorPage(
...
+                            },
+                        )
+                    },
+                    viewModel = editorViewModel,
+                )
+            }
+            composable(
+                route = EditorNavDestinations.ALT_TEXT.name,
+            ) {
+                AltTextSection(
```
AdamGrzybkowski: ⛏️ I see I wasn't consistent with the naming but I planned to add a `Page` suffix to each destination. Could we rename this to `AltTextPage`?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -49,21 +65,42 @@ internal fun GravatarQuickEditorPage(
...
-                gravatarQuickEditorParams = gravatarQuickEditorParams,
-                handleExpiredSession = true,
-                onAvatarSelected = onAvatarSelected,
-                onSessionExpired = { navController.navigate(QuickEditorPage.OAUTH.name) },
-            )
+        navigation(
+            route = QuickEditorPage.EDITOR.name,
+            startDestination = EditorNavDestinations.AVATAR_SELECTION.name,
+        ) {
+            composable(route = EditorNavDestinations.AVATAR_SELECTION.name) {
```
AdamGrzybkowski: 💡 In a separate PR, we could work a bit on the enter/exit animations.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -33,14 +43,20 @@ internal fun GravatarQuickEditorPage(
     onDismiss: (dismissReason: GravatarQuickEditorDismissReason) -> Unit = {},
 ) {
     val navController = rememberNavController()
+    val editorViewModel: AvatarPickerViewModel = viewModel(
+        factory = AvatarPickerViewModelFactory(
+            gravatarQuickEditorParams = gravatarQuickEditorParams,
+            handleExpiredSession = true,
+        ),
+    )
```
AdamGrzybkowski: ⚠️ There's a slight change in behavior here. This will make the `AvatarPickerViewModel` service QE relaunches. You can see that when you close/open the QE, there's no initial loading. 

I wonder if this would be problematic for apps that provide the token. In a scenario where the user changes but the ViewModel survives and persists the previous data.

AdamGrzybkowski: My suggestion would be to opt for a separate ViewModel anyway. So maybe it's not worth thinking about how to fix that.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextSection.kt
```diff
@@ -0,0 +1,238 @@
...
+                }
+
+                Row(
+                    verticalAlignment = Alignment.Top,
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .padding(top = 16.dp),
+                ) {
+                    val cornerRadius = 8.dp
+                    val avatarSize = 96.dp
```
AdamGrzybkowski: 💡 We have this size defined in `LazyAvatarRow.kt` - maybe it's worth using it if it's the same value.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextSection.kt
```diff
@@ -0,0 +1,238 @@
...
+                    val sizePx = with(LocalDensity.current) { avatarSize.roundToPx() }
+                    AsyncImage(
+                        model = altTextState.avatar.imageUrlWithSize(sizePx),
+                        contentDescription = stringResource(
+                            id = R.string.gravatar_qe_selectable_avatar_content_description,
+                        ),
+                        modifier = Modifier
+                            .size(96.dp)
```
AdamGrzybkowski:
```suggestion
                            .size(avatarSize)
```

⛏️ Shouldn't it be that?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextSection.kt
hamorillo: [Done](https://github.com/Automattic/Gravatar-SDK-Android/issues/513)

AdamGrzybkowski (commented):
That looks better now I think! I've left a few comments that are worth addressing first. 

I noticed that the Avatar in the profile card refreshes after coming back from the AltText section. I believe it's not directly related to this PR, but something that we should take a look at.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
AdamGrzybkowski: >I think we can still do another step and create a different ViewModel for the AltText section. Extracting the code related to the AltText should be "easy", the problem here is that we'll need to also make a way to sync the data between them. We can iterate over this in the following PRs.


I think we should do that. I left a comment related to that.

AdamGrzybkowski:
I created an issue for the Avatar reload - https://github.com/Automattic/Gravatar-SDK-Android/issues/514

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
hamorillo: I already have a [ticket](https://github.com/Automattic/Gravatar-SDK-Android/issues/498) for that 😃

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
hamorillo: I have a sticker physically in front of me to do this, and I still forgot it. 🤦‍♂️  Thanks!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextSection.kt
hamorillo: I am not sure if we should reuse that value. This is a different context and screen, and changing the size of the avatar in the list doesn't necessarily imply that we want to change it here.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
hamorillo: Working on it ⚙️

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextSection.kt
AdamGrzybkowski: Right, that makes sense!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPicker.kt
hamorillo: I've extracted the AltText related code to it's own ViewModel. The `AvatarPickerViewModel` handles only the rating changes so I've also refactored that code to remove all the unnecessary stuff. 

> we'll need to also make a way to sync the data between them

This PR doesn't accomplish that part. I think we should work on that in a different PR. For the moment, the AltText won't be updated in the local data models (it's updated in the backend) until the bottomsheet is closed and reopened again. (Which forces the data to be loaded from the server). #516

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -72,48 +72,52 @@ internal class AvatarPickerViewModel(
...
             is AvatarPickerEvent.AvatarDeleteSelected -> deleteAvatar(event.avatarId)
             AvatarPickerEvent.AvatarDeleteAlertDismissed -> hideNonSelectedAvatarAlert()
-            is AvatarPickerEvent.AvatarRatingSelected -> updateAvatar(event.avatarId, event.rating)
+            is AvatarPickerEvent.AvatarRatingSelected -> updateAvatarRating(event.avatarId, event.rating)
+            is AvatarPickerEvent.AvatarAltTextTapped -> launchAltTextEditor(event.avatarId)
         }
     }
 
-    private fun updateAvatar(avatarId: String, rating: Avatar.Rating? = null, altText: String? = null) {
+    private fun updateAvatarRating(avatarId: String, rating: Avatar.Rating? = null) {
```
hamorillo, line 80: The AvatarPicker modifies only the `Rating` so there is no need for the generic method that allow the `altText` update. I've remove/renamed all the related code.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -110,12 +122,64 @@ internal fun GravatarQuickEditorPage(
...
+            val avatarId = requireNotNull(it.arguments?.getString("avatarId"))
+            val altText = requireNotNull(it.arguments?.getString("altText"))
+            val avatarUrl = requireNotNull(it.arguments?.getString("avatarUrl"))
+            AltTextPage(
+                email = email,
+                avatarId = avatarId,
+                altText = altText,
+                avatarUrl = avatarUrl,
+                onBackPressed = { navController.popBackStack() },
+                modifier = Modifier.padding(16.dp),
```
AdamGrzybkowski: ❓  All other pages define their own paddings, rather than it being set here. Is there a reason you did it here?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextViewModel.kt
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+    }
+
+    private fun updateUiStateWithNewAltText(newAltText: String) {
+        viewModelScope.launch {
+            _uiState.update { currentState ->
+                currentState.copy(
+                    altText = newAltText,
+                    isSaveButtonEnabled = newAltText != altText,
```
AdamGrzybkowski, line 78: When the `newAltTest` is empty the API call returns 200, but it doesn't update it. Looks like it quietly fails. Should we disable the save button in such cases?

AdamGrzybkowski (approved):
Thanks for all the updates! Tested it and all worked fine!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
hamorillo: I thought about it as if it wasn't part of the section. If, at some point, we want to reuse the composable in a different place, we may require a different padding. However, you are right, and for the moment, I'll move it inside the composable to follow the same style as on the other pages.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextViewModel.kt
hamorillo: 🤔 Nice catch! I haven't noticed that.

The web is applying a "default" `AltText` when trying to save it as empty (Example: `Héctor Abraham's avatar`). The endpoint we are using is not working the same way. I'll check with the backend folks. If we want to set the default `AltText`, we need the backend to return the avatar object with that data.

hamorillo: It looks like this will be fixed in the backend and should be transparent for us (as we use the returned avatar object) so I think we can merge the PR.

AdamGrzybkowski: Sounds good!

PR #509, opened 2024-12-19
QuickEditor: Only invoke onAvatarSelected when selected avatar is deleted

### LINKED ISSUE #507: QuickEditor: onAvatarSelected is invoked when deleting unselected Avatar
https://github.com/user-attachments/assets/737b9da3-a024-46c0-8b42-2cb902e193b7

AdamGrzybkowski:
Closes #507

### Description

The `onAvatarSelected` callback would be invoked regardless of whether the deleted avatar was selected or not. This is not correct, we should only invoke it for the selected avatar.

### Testing Steps
1. Run the QE
2. Delete an unselected Avatar
3. Confirm the callback was not invoked (there shouldn't be an update in the demo app and no snack)


hamorillo (approved):
Thanks for taking care of this. 🚀

PR #510, opened 2024-12-19
Fix ViewProfileButton paddding when in Empty state

### LINKED ISSUE #504: Gravatar UI: Review ProfileCard Empty state
The empty state is bigger than the loaded state. Should we both have the same height?

Image

AdamGrzybkowski:
Closes #504

### Description

The `ViewProfile` button in the `ProfileSummary` component had an off padding when in Empty state. It turns out that when in Empty state, the external Modifier was not set on the button, but it was in non-empty state leading to different look.

| Before | After |
|---|---|
| Screenshot 2024-12-19 at 11 50 44 | Screenshot 2024-12-19 at 11 50 33 |

### Testing Steps

1. Use the profile tab in the demo app
2. Display the profile card for a random email that doesn't have an associated Gravatar account 
3. Confirm that the 2nd card in the list looks correct


hamorillo (approved):
👨‍🎨 🚀

PR #512, opened 2024-12-19
QuickEditor: Increase opacity of the avatar's menu button

### LINKED ISSUE #511: QuickEditor: Increase the opacity of the Avatar's menu button
The button can be easily missed if the background doesn't play with the current opacity. We want to try to mitigate this a bit.

hamorillo:
Closes #511 

### Description

Avatar's menu button can be missed depending on the avatar behind it. We are increasing the opacity to try to make it more visible.

The visibility of the button improves a bit but still can be missed depending on the avatar: 

image

### Testing Steps

- CI (Screenshot test)
- Smoke the avatar picker to review the opacity change


AdamGrzybkowski (approved):
LGTM 👍

PR #520, opened 2025-01-07
QuickEditor: AltText - Save the avatars list locally in the AvatarsRespository instead of in the `AvatarPickerViewModel`

### LINKED ISSUE #516: Quick Editor: Alt Text - Sync data between `AltTextViewModel` and `AvatarPickerViewModel`
For the moment, the AltText won't be updated in the local data models (it's updated in the backend) until the bottomsheet is closed and reopened again. (Which forces the data to be loaded from the server).

_Originally posted by @hamorillo in https://github.com/Automattic/Gravatar-SDK-Android/pull/501#discussion_r1899592046_

hamorillo:
Closes #516 

### Description

With the new `AltTextPage`, we need to save the avatar's data in a shared place as we need to access that info from both screens. We already have the `AvatarsRepository`, so that's the best place to keep the avatars list locally. We need to create a flow for each email and update that flow in every repository operation.

We were doing a similar thing in the `AvatarPickerViewModel`. We had a variable where we kept the avatars list and updated it in the different operations (avatar upload, deletion, selection...). This PR moves that logic to the repository while sharing the avatar flow for each email. 

### Testing Steps

Everything should work as before except the AltText, where this PR should solve the issue of reopening the AltText page after changing it. In that situation, the new AltText should be visible, not as it was before when the old AltText was visible.

- Smoke test all the operations in the QE (avatar upload, deletion, selection, rating)
- Test that after changing the `AltText` and reopening the AltTextPage for the same image, the new AltText is visible instead of the old one.

https://github.com/user-attachments/assets/9257df1d-8c47-410a-8717-df6c2cf61b52

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/QuickEditorContainer.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/QuickEditorContainer.kt
@@ -91,12 +91,13 @@ internal class QuickEditorContainer private constructor(
         FileUtils(context)
     }
 
-    val avatarRepository: AvatarRepository
-        get() = AvatarRepository(
+    val avatarRepository: AvatarRepository by lazy {
+        AvatarRepository(
             avatarService = avatarService,
             tokenStorage = tokenStorage,
             dispatcher = Dispatchers.IO,
         )
+    }
 
     val imageDownloader: ImageDownloader by lazy { ImageDownloader(context = context) }
 
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/QuickEditorContainer.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/QuickEditorContainer.kt
@@ -13,6 +13,7 @@ import com.gravatar.quickeditor.data.FileUtils
 import com.gravatar.quickeditor.data.ImageDownloader
 import com.gravatar.quickeditor.data.datastore.createEncryptedFileWithFallbackReset
 import com.gravatar.quickeditor.data.repository.AvatarRepository
+import com.gravatar.quickeditor.data.storage.AvatarStorage
 import com.gravatar.quickeditor.data.storage.DataStoreProfileStorage
 import com.gravatar.quickeditor.data.storage.DataStoreTokenStorage
 import com.gravatar.quickeditor.data.storage.InMemoryTokenStorage
@@ -91,13 +92,17 @@ internal class QuickEditorContainer private constructor(
         FileUtils(context)
     }
 
-    val avatarRepository: AvatarRepository by lazy {
-        AvatarRepository(
+    private val avatarStorage: AvatarStorage by lazy {
+        AvatarStorage()
+    }
+
+    val avatarRepository: AvatarRepository
+        get() = AvatarRepository(
             avatarService = avatarService,
             tokenStorage = tokenStorage,
+            avatarStorage = avatarStorage,
             dispatcher = Dispatchers.IO,
         )
-    }
 
     val imageDownloader: ImageDownloader by lazy { ImageDownloader(context = context) }
 
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/repository/AvatarRepository.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/repository/AvatarRepository.kt
@@ -9,6 +9,8 @@ import com.gravatar.services.AvatarService
 import com.gravatar.services.GravatarResult
 import com.gravatar.types.Email
 import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.withContext
 
 internal class AvatarRepository(
@@ -16,18 +18,26 @@ internal class AvatarRepository(
     private val tokenStorage: TokenStorage,
     private val dispatcher: CoroutineDispatcher,
 ) {
+    fun getAvatarsFlow(email: Email) = avatarsFlow(email).asSharedFlow()
+
+    private val avatarsFlows = mutableMapOf<String, MutableSharedFlow<EmailAvatars>>()
+
+    private fun avatarsFlow(email: Email): MutableSharedFlow<EmailAvatars> =
+        avatarsFlows.getOrPut(email.hash().toString()) { MutableSharedFlow(replay = 1) }
+
     suspend fun getAvatars(email: Email): GravatarResult<EmailAvatars, QuickEditorError> = withContext(dispatcher) {
         val token = tokenStorage.getToken(email.hash().toString())
         token?.let {
             when (val avatarsResult = avatarService.retrieveCatching(token, email.hash())) {
                 is GravatarResult.Success -> {
-                    val avatars = avatarsResult.value
-                    GravatarResult.Success(
+                    val emailAvatars = avatarsResult.value.let { avatars ->
                         EmailAvatars(
                             avatars,
                             avatars.firstOrNull { it.selected == true }?.imageId,
-                        ),
-                    )
+                        )
+                    }
+                    avatarsFlow(email).emit(emailAvatars)
+                    GravatarResult.Success(emailAvatars)
                 }
 
                 is GravatarResult.Failure -> {
@@ -84,10 +94,32 @@ internal class AvatarRepository(
         val token = tokenStorage.getToken(email.hash().toString())
         token?.let {
             when (val result = avatarService.updateAvatarCatching(avatarId, token, rating, altText)) {
-                is GravatarResult.Success -> GravatarResult.Success(result.value)
+                is GravatarResult.Success -> {
+                    updateAvatarFlow(email, result.value)
+                    GravatarResult.Success(result.value)
+                }
+
                 is GravatarResult.Failure -> GravatarResult.Failure(QuickEditorError.Request(result.error))
             }
         } ?: GravatarResult.Failure(QuickEditorError.TokenNotFound)
+    }
+
+    private suspend fun updateAvatarFlow(email: Email, avatarToUpdate: Avatar) {
+        avatarsFlow(email).let { avatarsFlow ->
+            val emailAvatars = avatarsFlow.replayCache.lastOrNull()?.let {
+                it.copy(
+                    avatars = it.avatars.map { avatar ->
+                        if (avatar.imageId == avatarToUpdate.imageId) {
+                            avatarToUpdate
+                        } else {
+                            avatar
+                        }
+                    },
+                )
+            }
+
+            emailAvatars?.let { avatarsFlow.emit(it) }
+        }
     }
 }
 
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/repository/AvatarRepository.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/repository/AvatarRepository.kt
@@ -133,7 +133,25 @@ internal class AvatarRepository(
         val token = tokenStorage.getToken(email.hash().toString())
         token?.let {
             when (val result = avatarService.deleteAvatarCatching(avatarId, token)) {
-                is GravatarResult.Success -> GravatarResult.Success(Unit)
+                is GravatarResult.Success -> {
+                    avatarsFlow(email).let { avatarsFlow ->
+                        val emailAvatars = avatarsFlow.replayCache.lastOrNull()?.let {
+                            it.copy(
+                                avatars = it.avatars.filter { avatar ->
+                                    avatar.imageId != avatarId
+                                },
+                                selectedAvatarId = if (it.selectedAvatarId == avatarId) {
+                                    null
+                                } else {
+                                    it.selectedAvatarId
+                                },
+                            )
+                        }
+
+                        emailAvatars?.let { avatarsFlow.emit(it) }
+                    }
+                    GravatarResult.Success(Unit)
+                }
                 is GravatarResult.Failure -> GravatarResult.Failure(QuickEditorError.Request(result.error))
             }
         } ?: GravatarResult.Failure(QuickEditorError.TokenNotFound)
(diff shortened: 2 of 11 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/QuickEditorContainer.kt
```diff
@@ -91,12 +91,13 @@ internal class QuickEditorContainer private constructor(
         FileUtils(context)
     }
 
-    val avatarRepository: AvatarRepository
-        get() = AvatarRepository(
+    val avatarRepository: AvatarRepository by lazy {
```
hamorillo: The `avatarRepository` needs to be a singleton that can be shared between screens.

AdamGrzybkowski: I think we should create `AvatarsStorage` to offload the repository. That way only the storage needs to be a Singleton and the Repository will make sure we get the latest data (from remote or local).

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -443,6 +428,16 @@ internal class AvatarPickerViewModel(
             .launchIn(viewModelScope)
     }
 
+    private fun avatarsObserver() {
```
AdamGrzybkowski: ⛏️ I would rename it to `collectAvatars` or `observe` - the current name sounds like we would return or create the Observer.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/repository/AvatarRepository.kt
```diff
@@ -4,30 +4,42 @@ import android.net.Uri
...
+                    val emailAvatars = avatarsResult.value.let { avatars ->
                         EmailAvatars(
                             avatars,
                             avatars.firstOrNull { it.selected == true }?.imageId,
-                        ),
-                    )
+                        )
+                    }
+                    avatarsFlow(email).emit(emailAvatars)
+                    GravatarResult.Success(emailAvatars)
```
AdamGrzybkowski: 💡  If we are going to have a separate flow with avatars I think we should rename this function to `refreshAvatars` and not return them here. This could be confusing, and unclear whether the avatars returned should be used or those emitted in the flow.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/repository/AvatarRepository.kt
```diff
@@ -4,30 +4,42 @@ import android.net.Uri
...
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.withContext
 
 internal class AvatarRepository(
     private val avatarService: AvatarService,
     private val tokenStorage: TokenStorage,
     private val dispatcher: CoroutineDispatcher,
 ) {
+    fun getAvatarsFlow(email: Email) = avatarsFlow(email).asSharedFlow()
```
AdamGrzybkowski:
```suggestion
    fun getAvatars(email: Email) = avatarsFlow(email).asSharedFlow()
```

⛏️ I think I would shorten the name and skip the `flow` suffix. The return type is specified so not sure having it in the name adds much value.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/repository/AvatarRepository.kt
```diff
@@ -4,30 +4,42 @@ import android.net.Uri
...
+
+    private val avatarsFlows = mutableMapOf<String, MutableSharedFlow<EmailAvatars>>()
```
AdamGrzybkowski: The `EmailAvatars` object was created based on how the initial API was structured but also to serve the presentation layer. I don't think this is how we should store it. The `Avatar` object already has the `selected` field so it should be enough to store a plain list of Avatars and mapped to the `EmailAvatars` in the ViewModel.

AdamGrzybkowski (commented):
Works great! But I would like to hear from you what you think about the comment about extracting all the local storage from the repository. The Repository would stay cleaner and without any knowledge of how objects are stored. It would make sure the data is returned from a proper source.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/QuickEditorContainer.kt
hamorillo: Good point @AdamGrzybkowski. I agree that creating `AvatarsStorage` will make the `AvatarRepository` cleaner. I'll refactor this.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/repository/AvatarRepository.kt
```diff
@@ -84,14 +95,13 @@ internal class AvatarRepository(
...
+                    GravatarResult.Success(result.value)
+                }
+
                 is GravatarResult.Failure -> GravatarResult.Failure(QuickEditorError.Request(result.error))
             }
         } ?: GravatarResult.Failure(QuickEditorError.TokenNotFound)
     }
 }
-
-internal data class EmailAvatars(
```
hamorillo, line 94: `AvatarRepository` will work only with `Avatar`s, the viewmodel will map the Avatars to EmailAvatars, so moving this class to the PickersViewModel.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/QuickEditorContainer.kt
hamorillo: I've refactored the code as discussed. `AvatarsStorage` handles the avatar list for each email locally.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/repository/AvatarRepository.kt
hamorillo: I renamed the function; however, it still returns the avatar lists because we use it to change the scroll position in the viewmodel. If we don't return the avatar list here we need to refactor how we scroll to the selected avatar in that case. I'm not saying I don't want to do it, but I would prefer to work on it in a different PR because this one is growing quite a bit and doesn't look like a blocker to me. Wdyt?

AdamGrzybkowski: hmm, I see. I think we can do it later. We will have two places where we get the data, so I see this as a potential source for problems in the future.

hamorillo: > We will have two places where we get the data, so I see this as a potential source for problems in the future.

Yeah, I agree.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/storage/AvatarStorage.kt
```diff
@@ -0,0 +1,86 @@
...
+import com.gravatar.types.Email
+import kotlinx.coroutines.flow.MutableSharedFlow
+
+internal class AvatarStorage {
+    private val avatarsFlows = mutableMapOf<String, MutableSharedFlow<List<Avatar>>>()
+
+    internal fun avatarsFlow(email: Email): MutableSharedFlow<List<Avatar>> =
+        avatarsFlows.getOrPut(email.hash().toString()) { MutableSharedFlow(replay = 1) }
+
+    internal suspend fun storeAvatars(avatars: List<Avatar>, email: Email) {
```
AdamGrzybkowski, line 14: ⛏️ Totall nitpick, but email param is first in other functions :D

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
```diff
@@ -51,6 +51,7 @@ internal class AvatarPickerViewModel(
     init {
         refresh()
         nonAvatarSelectedAlertObserver()
+        collectAvatars()
```
AdamGrzybkowski, line 54: 💡 Not sure but I think we might be able to combine these two collectors now. It's not something for this PR though.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/storage/AvatarStorage.kt
```diff
@@ -0,0 +1,86 @@
...
+        avatarsFlow(email).emit(avatars)
+    }
+
+    internal suspend fun markAvatarAsSelected(email: Email, avatarId: String) {
```
AdamGrzybkowski:
```suggestion
    internal suspend fun selectAvatar(email: Email, avatarId: String) {
```

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/data/storage/AvatarStorageTest.kt
```diff
@@ -0,0 +1,128 @@
...
+class AvatarStorageTest {
+    private lateinit var avatarStorage: AvatarStorage
+
+    @Before
+    fun setUp() {
+        avatarStorage = AvatarStorage()
+    }
+
+    @Test
+    fun `given avatars when storeAvatars then avatarsFlow emits a avatars`() = runTest {
```
AdamGrzybkowski:
```suggestion
    fun `given avatars when storeAvatars then avatarsFlow emits avatars`() = runTest {
```

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/data/storage/AvatarStorageTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        avatarStorage.storeAvatars(avatars, email)
+
+        // Then
+        avatarStorage.avatarsFlow(email).test {
+            assertEquals(avatars, awaitItem())
+        }
+    }
+
+    @Test
+    fun `given avatarId when markAvatarAsSelected then avatarsFlow emits a avatars with selected avatar`() = runTest {
```
AdamGrzybkowski:
```suggestion
    fun `given avatarId when markAvatarAsSelected then avatarsFlow emits avatars with selected avatar`() = runTest {
```

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/data/storage/AvatarStorageTest.kt
```diff
@@ -0,0 +1,128 @@
...
+                    createAvatar("otherAvatar", isSelected = false),
+                    createAvatar(id = "imageId", isSelected = true),
+                ),
+                awaitItem(),
+            )
+        }
+    }
+
+    @Test
+    fun `given avatar when addAvatar then avatarsFlow emits a avatars with updated avatar`() = runTest {
```
AdamGrzybkowski:
```suggestion
    fun `given avatar when addAvatar then avatarsFlow emits avatars`() = runTest {
```

This one adds new one, not updates it, right?

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/data/storage/AvatarStorageTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        avatarStorage.avatarsFlow(email).test {
+            assertEquals(
+                listOf(updatedAvatar, createAvatar("otherAvatar", isSelected = false)),
+                awaitItem(),
+            )
+        }
+    }
+
+    @Test
+    fun `given avatarId when deleteAvatar then avatarsFlow emits a avatars without deleted avatar`() = runTest {
```
AdamGrzybkowski:
```suggestion
    fun `given avatarId when deleteAvatar then avatarsFlow emits avatars without deleted avatar`() = runTest {
```

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/data/storage/AvatarStorageTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        avatarStorage.avatarsFlow(email).test {
+            assertEquals(
+                listOf(createAvatar("otherAvatar", isSelected = false)),
+                awaitItem(),
+            )
+        }
+    }
+
+    @Test
+    fun `given avatar when updateAvatar then avatarsFlow emits a avatars with updated avatar`() = runTest {
```
AdamGrzybkowski:
```suggestion
    fun `given avatar when updateAvatar then avatarsFlow emits avatars with updated avatar`() = runTest {
```

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModelTest.kt
```diff
@@ -351,69 +351,8 @@ class AvatarPickerViewModelTest {
...
-                ),
-                awaitItem(),
-            )
-        }
-        verify { fileUtils.deleteFile(uri) }
-    }
-
-    @OptIn(ExperimentalCoroutinesApi::class)
-    @Test
-    fun `given avatar returned when avatar with the same id then uiState is updated`() = runTest {
```
AdamGrzybkowski, line 359: ❓ This one is not needed now?

AdamGrzybkowski (approved):
It works great, and the alt text is synchronized! 

We could slim down the amount of data we pass to the AltTextPage now.

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModelTest.kt
hamorillo: This logic is no longer part of the ViewModel as it's handled by the AvatarStore/Repository, so I think this test is unnecessary.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
hamorillo: Issue #522

hamorillo:
> We could slim down the amount of data we pass to the AltTextPage now.

Issue #523

PR #521, opened 2025-01-08
Fix no avatar selected banner shown when avatar list empty

### LINKED ISSUE #519: QuickEditor: No avatar selected banner is visible when no Avatars are uploaded
When the avatar list is empty we show a special message, the banner is unnecessary. We should also make sure this isn't shown when uploading the initial avatar.

Image

AdamGrzybkowski:
Closes #519

### Description

This PR fixes a bug where the alert banner about no selected avatar is visible when no avatars are uploaded.

| Before | After |
|---|---|
| Screenshot 2025-01-07 at 11 35 03 | Screenshot 2025-01-08 at 11 36 24 |

### Testing Steps

1. Make sure you have no avatars uploaded
2. Launch the QE
3. The banner shouldn't be visible
4. You can upload two avatars (make sure the banner is not visible when uploading the first one), select one, and remove: 
5. Selected one first
6. Then select the other
7. Make sure the banner visibility is correct with each step


hamorillo (approved):
🚀

PR #524, opened 2025-01-09
Handle the back navigation manually in the QuickEditor

### LINKED ISSUE #513: Quick Editor: Alt Text - Fix back handler after keyboard is opened
⚠️ This works well before the keyboard is opened. However, after that, the bottom sheet is dismissed instead of being handled here. If that's OK with you, I'll create a ticket for this.

_Originally posted by @hamorillo in https://github.com/Automattic/Gravatar-SDK-Android/pull/501#discussion_r1892941247_

AdamGrzybkowski:
Closes #513

### Description

The system back button wouldn't work properly after showing the keyboard. For some reason, the `AltTextPage`'s back handler would lose the priority and the Bottom Sheet would take over and dismiss itself. 

I could figure out why, but rather than trying to fix this, I went ahead and implemented the back behavior manually, by properly clearing the navigation stack and hiding the bottom sheet when nothing is left. I think this is the way to go now, given we have an actual pages in the stack that should be kept. 

### Testing Steps

1. Test the back button behavior within the Quick Editor. Test OAuth flow, Alt text modifications, etc.

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
@@ -12,7 +12,6 @@ import androidx.navigation.compose.composable
 import androidx.navigation.compose.navigation
 import androidx.navigation.compose.rememberNavController
 import androidx.navigation.navArgument
-import androidx.navigation.navOptions
 import com.gravatar.quickeditor.QuickEditorContainer
 import com.gravatar.quickeditor.ui.alttext.AltTextPage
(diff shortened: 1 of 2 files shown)


hamorillo (commented):
Works great! 

❓ In line 77 we navigate to OAUTH page when the sessionExpires, shouldn't we use the `navigateAndClean` too?

❓ In line 155 (it's not directly related to this PR) I think we should remove the `navOptions`:

```
navOptions = navOptions {
    popUpTo(EditorNavDestinations.AVATAR_SELECTION.name) { saveState = true }
},
```

AdamGrzybkowski:
@hamorillo Good call! Fixed it.

hamorillo (approved):
Thanks for the changes! 🚀

PR #527, opened 2025-01-10
QuickEditor: Fix `AvatarStorage.updateAvatar` to not override the `selected` value 

### LINKED ISSUE #526: QuickEditor: AvatarStore.updateAvatar should omit selected value
When updating an avatar, the `selected` information is not returned. Therefore, the `AvatarStorage` can't just use the returned object; we need to use the `selected` value from the previously stored object.

hamorillo:
Closes #526 

### Description

The `PATCH` to update avatars doesn't receive the email, so, in the backend, we can't decide if that's the avatar selected for the email we are working with.

In the `AvatarStorage.updateAvatar` method, we should only use the value from the `updatedAvatar` if it's not `null`. Otherwise, we should keep the stored avatar's selected value.

<details>
<summary> Before </summary>

https://github.com/user-attachments/assets/3b2fe41b-13a7-41c8-b585-4d33588540f8
</details>

<details>
<summary> After </summary>

https://github.com/user-attachments/assets/f30f84e1-ff05-4b5c-853d-1fb5803c0a7e
</details>

### Testing Steps

1. Open the QE
2. Select an avatar (if you don't have one selected already)
3. Modify the rating of the selected avatar
4. Verify that you still see the avatar as selected when the rating is modified
5. Open the AltTextPage and modify the AltText for the selected avatar
6. Verify that you still see the avatar as selected when the AltText is modified

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/storage/AvatarStorage.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/storage/AvatarStorage.kt
@@ -73,7 +73,9 @@ internal class AvatarStorage {
             val avatars = avatarsFlow.replayCache.lastOrNull()?.let {
                 it.map { avatar ->
                     if (avatar.imageId == avatarToUpdate.imageId) {
-                        avatarToUpdate
+                        // If the avatarToUpdate has a selected value, use it.
+                        // Otherwise, use the stored avatar selected value.
+                        avatarToUpdate.copy(selected = avatarToUpdate.selected ?: avatar.selected)
                     } else {
                         avatar
                     }
--- a/gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/data/storage/AvatarStorageTest.kt
+++ b/gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/data/storage/AvatarStorageTest.kt
@@ -2,6 +2,7 @@ package com.gravatar.quickeditor.data.storage
 
 import app.cash.turbine.test
 import com.gravatar.quickeditor.createAvatar
+import com.gravatar.quickeditor.ui.avatarpicker.copy
 import com.gravatar.types.Email
 import junit.framework.TestCase.assertEquals
 import kotlinx.coroutines.test.runTest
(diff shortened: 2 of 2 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/storage/AvatarStorage.kt
```diff
@@ -73,7 +73,9 @@ internal class AvatarStorage {
             val avatars = avatarsFlow.replayCache.lastOrNull()?.let {
                 it.map { avatar ->
                     if (avatar.imageId == avatarToUpdate.imageId) {
-                        avatarToUpdate
+                        // If the avatarToUpdate has a selected value, use it.
+                        // Otherwise, use the stored avatar selected value.
+                        avatarToUpdate.copy(selected = avatarToUpdate.selected ?: avatar.selected)
```
AdamGrzybkowski, line 78: 💡 This solution works but I see a potential issue. This would stop working if backend started sending back the `selcted` value as false instead of null. If we only copied here the values that can actually change (rating and the alt_text) we would be much safer.

This might be unlikely scenario, but technically could happen. 

WDYT?

AdamGrzybkowski (commented):
Great this requires modification only at the storage level!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/data/storage/AvatarStorage.kt
hamorillo: > This would stop working if backend started sending back the selcted value as false instead of null.

In my opinion, that would be a bug on the backend because, as we've seen, we don't know if it is selected or not. If the backend sends that value as `false/true,` it is because that's the correct value, and we should update it.

Anyway, I'm not entirely against only updating rating and alt_text, but I was looking at the `updateAvatar` method as a more abstract one that, if, for example, the URL changes, can update the URL too. (Just as an example). 

I can refactor this if you still feel that approach is much safer. 🙂

AdamGrzybkowski: >In my opinion, that would be a bug on the backend because, as we've seen, we don't know if it is selected or not. If the backend sends that value as false/true, it is because that's the correct value, and we should update it.

I agree. 

>I can refactor this if you still feel that approach is much safer.

No need to, let's stick with this one.

AdamGrzybkowski (approved):

PR #528, opened 2025-01-10
QuickEditor: Clickable alt text help link brings to support page

### LINKED ISSUE #500: Quick Editor: Alt Text - Implement "What is Alt Text?" feature

mlumeau:
Closes #500

### Description

This PR adds a navigation to the [support page article regarding alt text](https://support.gravatar.com/profiles/avatars/#add-alt-text-to-avatars) to the ~~"What is alt text?" label~~ "?" button in the alt text dialog.

Navigation happens in a new browser tab, to be consistent with the behavior of other links.

### Testing Steps
1. Open the quick editor
2. Open the three-dot menu on an avatar and click on "Alt text"
3. Click on ~~"What is alt text?"~~ the "?" button
4. - [ ] Make sure a new browser tab opens at the [support page article regarding alt text](https://support.gravatar.com/profiles/avatars/#add-alt-text-to-avatars)


gravatar/src/main/java/com/gravatar/GravatarConstants.kt
```diff
@@ -18,4 +18,8 @@ public object GravatarConstants {
 
     /** Gravatar Sign-in URL */
     public const val GRAVATAR_SIGN_IN_URL: String = "$GRAVATAR_BASE_URL/profile"
+
+    /** Gravatar Alt Text help URL */
+    public const val GRAVATAR_ALT_TEXT_HELP_URL: String =
+        "https://support.gravatar.com/profiles/avatars/#add-alt-text-to-avatars"
```
AdamGrzybkowski: ❓ Do we really want this as a public value exposed from the core module? This is needed in the QuickEditor only, so I think having it there as an internal value makes more sense. We can always expose it, but we can't hide it.

gravatar/src/main/java/com/gravatar/GravatarConstants.kt
```diff
@@ -18,4 +18,8 @@ public object GravatarConstants {
 
     /** Gravatar Sign-in URL */
     public const val GRAVATAR_SIGN_IN_URL: String = "$GRAVATAR_BASE_URL/profile"
+
+    /** Gravatar Alt Text help URL */
+    public const val GRAVATAR_ALT_TEXT_HELP_URL: String =
```
hamorillo: ❓ 💡 I'm unsure if we should add that URL to the core module. It looks like something that only the QE needs, and making that `public` will break the API if we decide to rename/remove it.  What do you think about moving that somewhere to the QE and keeping it `internal`?

gravatar/src/main/java/com/gravatar/GravatarConstants.kt
mlumeau: Yeah I realized this while updating the .api file. We don't have a constants file in the QuickEditor yet so I'll move this url to the AltText composable directly, since it's the only place where we'll use this for now.

hamorillo (commented):
💡 I remember discussing whether we should replace the label "What is alt text?" with just the icon ❔ . Maybe we can double-check that before merging.

gravatar/src/main/java/com/gravatar/GravatarConstants.kt
mlumeau: Fixed in [2d322aa](https://github.com/Automattic/Gravatar-SDK-Android/pull/528/commits/2d322aa5564c58710adb7d24c6be0d76af6112c8)

gravatar/src/main/java/com/gravatar/GravatarConstants.kt
mlumeau: Fixed in [dadad6e](https://github.com/Automattic/Gravatar-SDK-Android/pull/528/commits/dadad6ef656460ca449ca23aa63378df6dee7010)

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextPage.kt
```diff
@@ -138,9 +147,13 @@ internal fun AltTextPage(
...
                     )
-                    Text(
-                        text = stringResource(id = R.string.gravatar_qe_avatar_alt_text_section_what_is),
-                        color = MaterialTheme.colorScheme.primary,
+                    Icon(
+                        painter = painterResource(id = R.drawable.gravatar_alt_text_help),
+                        tint = MaterialTheme.colorScheme.primary,
+                        contentDescription = stringResource(id = R.string.gravatar_qe_avatar_alt_text_section_what_is),
+                        modifier = Modifier.clickable {
+                            uriHandler.openUri(GRAVATAR_ALT_TEXT_HELP_URL)
```
hamorillo, line 155: ⛏️ Nitpicking: We can try to increase the clickable area slightly (maybe with padding? But, indeed, we don't have a lot of space, so feel free to omit this comment.

hamorillo (approved):
🚀

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextPage.kt
mlumeau: Added a 4dp padding to the button, that's all I could do without increasing the height of the page.

hamorillo: With the new padding, the "?" button is not aligned with the Save button and looks weird. I didn't realize about that when I wrote my first comment. Maybe we should keep it as it was (without the padding) and not modify all paddings in the composables. Sorry for bothering you. 😞

mlumeau: As you prefer 😅 I didn't notice the misalignment until your remark. I'll revert and merge!

PR #529, opened 2025-01-10
Alt text: Alt text counter

### LINKED ISSUE #502: Quick Editor: Alt Text - Limit to 125 the number of characters

AdamGrzybkowski:
Closes #502

### Description

This PR adds a little counter to show the remaining characters count for the alt text. Once the limit is reached the alt text won't be modified.

https://github.com/user-attachments/assets/3b06012d-72b2-4b89-bf42-240e07892f0a

### Testing Steps

1. Launch the QE
2. Open the avatar's menu and tap Alt text
3. Add new characters to the input field to reach the limit
4. Confirm no more chars can be added
5. Remove the text completely
6. Confirm the counter shows 125
7. Paste a text that's longer than 125 chars
8. Confirm only the first 125 chars were copied


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextPage.kt
```diff
@@ -175,7 +185,26 @@ internal fun AltTextPage(
                         textStyle = LocalTextStyle.current.copy(color = MaterialTheme.colorScheme.onSurface),
                         modifier = Modifier
                             .fillMaxWidth()
-                            .padding(start = 16.dp, top = 8.dp, bottom = 8.dp),
+                            .fillMaxHeight(),
+                        decorationBox = { innerTextField ->
```
AdamGrzybkowski, line 203: This makes it so that the whole `decorationBox` acts as a touch are - meaning we can tap everywhere to focus the text field.

We could also use this for the placeholder I think.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextPage.kt
```diff
@@ -175,7 +185,26 @@ internal fun AltTextPage(
...
+                            Column(
+                                modifier = Modifier
+                                    .padding(start = 16.dp, top = 8.dp, bottom = 8.dp),
+                            ) {
+                                Box(
+                                    modifier = Modifier.weight(1f),
+                                ) {
+                                    innerTextField()
+                                }
```
AdamGrzybkowski, line 212: The Box is used to take the available space via a weight modifier. I couldn't find a way to apply modifier to the inner text field.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextPage.kt
```diff
@@ -175,7 +185,26 @@ internal fun AltTextPage(
                         textStyle = LocalTextStyle.current.copy(color = MaterialTheme.colorScheme.onSurface),
                         modifier = Modifier
                             .fillMaxWidth()
-                            .padding(start = 16.dp, top = 8.dp, bottom = 8.dp),
+                            .fillMaxHeight(),
+                        decorationBox = { innerTextField ->
+                            Column(
```
AdamGrzybkowski, line 204: This is not ideal as the counter is technically below the input field, not inside it, but there's no easy way to handle that so that the text doesn't overlap with the counter.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextUiState.kt
```diff
@@ -2,7 +2,11 @@ package com.gravatar.quickeditor.ui.alttext
...
     val avatarUrl: String,
-    val isSaveButtonEnabled: Boolean,
     val isUpdating: Boolean,
     val altText: String,
-)
+    val altTextMaxLength: Int,
+    private val initialAltText: String = altText,
+) {
+    val isSaveButtonEnabled: Boolean
+        get() = altText != initialAltText && !isUpdating
```
AdamGrzybkowski, line 13: We don't need to set this manually, so I moved it here and set it conditionally.

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/alttext/AltTextViewModelTest.kt
```diff
@@ -117,22 +139,23 @@ class AltTextViewModelTest {
 
         viewModel.onEvent(AltTextEvent.AvatarAltTextChange(newAltText))
 
-        viewModel.onEvent(AltTextEvent.AvatarAltTextSaveTapped)
+        advanceUntilIdle()
 
         viewModel.uiState.test {
-            skipItems(2)
+            expectMostRecentItem()
+            viewModel.onEvent(AltTextEvent.AvatarAltTextSaveTapped)
```
AdamGrzybkowski: I think this is a bit more future-proof as we are only testing the states after `AvatarAltTextSaveTapped` event without skipping fixed numbers of states.

mlumeau (commented):
I noticed a couple of things while testing:

- When I hit 0 in the counter, move the caret to the middle of the paragraph and start typing again, it erases the end of the text. Is that on purpose? 
- This is maybe worth creating another issue but when adding multiple new lines at some point the "save" button disappears with no way to bring it back.

https://github.com/user-attachments/assets/06ca6bdb-5554-41e1-af49-cd3e22f55f22

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextPage.kt
hamorillo: It's acceptable to me.

AdamGrzybkowski:
>When I hit 0 in the counter, move the caret to the middle of the paragraph and start typing again, it erases the end of the text. Is that on purpose?

Not on purpose, but that's what the `take(125)` does now. I'm open to whether this is ok for us or not. 

>This is maybe worth creating another issue but when adding multiple new lines at some point the "save" button disappears with no way to bring it back.

Yes, this is unrelated. https://github.com/Automattic/Gravatar-SDK-Android/issues/533

hamorillo:
> Not on purpose, but that's what the take(125) does now. I'm open to whether this is ok for us or not.

That behavior can create some misunderstandings. However, I agree it's not easy to handle; for example, what if I have five left characters and I paste 20 characters in the middle? We have to consider those cases.

For example, Revolut handles that with a negative counter. When you reach the limit, you are not allowed to add any more characters.

AdamGrzybkowski:
I wouldn't complicate things too much. Here's my suggestion:
- Disallow adding extra characters when the alt text reaches the limit regardless of the cursor position.
- Keep the first 125 chars from the inserted text when the limit is not reached. That would mean some text could be lost if the cursor was in the middle (current behavior).


WDYT?

mlumeau:
> I wouldn't complicate things too much. Here's my suggestion:
> 
> * Disallow adding extra characters when the alt text reaches the limit regardless of the cursor position.
> * Keep the first 125 chars from the inserted text when the limit is not reached. That would mean some text could be lost if the cursor was in the middle (current behavior).
> 
> WDYT?

Sounds good to me! I doubt we'll get many complaints about this obscure edge case anyway.

AdamGrzybkowski:
@mlumeau  @hamorillo Done.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextPage.kt
```diff
@@ -188,7 +198,26 @@ internal fun AltTextPage(
                         textStyle = LocalTextStyle.current.copy(color = MaterialTheme.colorScheme.onSurface),
                         modifier = Modifier
                             .fillMaxWidth()
-                            .padding(start = 16.dp, top = 8.dp, bottom = 8.dp),
+                            .fillMaxHeight(),
```
hamorillo: ⛏ I think we can replace it with `fillMaxSize()`

mlumeau (approved):
👍 LGTM. I noticed a tiny weird thing, totally not blocking: the caret is still moving when trying to insert text at limit.

hamorillo (approved):
I was writing a similar comment to Maxime's about the cursor moving without inserting any text. 🙂

AdamGrzybkowski:
>I noticed a tiny weird thing, totally not blocking: the caret is still moving when trying to insert text at limit.

Yeah, I noticed that. The thing is we don't control the cursor position and to do that we have to keep that in the ViewModel. I did a quick try and it's a pain, to be honest. I will merge this for now and we can improve this later if we feel like it's a big issue.

PR #531, opened 2025-01-10
QuickEditor: Load Avatar object from `AvatarsRepository` in `AltTextViewModel`

### LINKED ISSUE #523: Quick Editor: Alt Text - Refactor AltText page to reduce the number of requiered params
AltTextPage should work just with:
- email: String,
- avatarId: String,

With the `avatarId`, we can load the rest of the required data from AvatarStorage.

hamorillo:
Closes #523 

### Description

With the recent implementation of the `AvatarStorage` we are ready to remove the extra params we needed to open the AltTextPage (url and alt_text). 

We can load the avatar data from the repository just by using the `avatarId` so we can simplify the navigation as we only need the email and the avatarId to open the AltTextPage.

### Testing Steps

- Smoke test the AltText feature (everything should work as before)

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextAction.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextAction.kt
@@ -4,4 +4,6 @@ internal sealed class AltTextAction {
     data object AltTextUpdated : AltTextAction()
 
     data object AltTextUpdateFailed : AltTextAction()
+
+    data object AvatarCantBeLoaded : AltTextAction()
 }
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextViewModel.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextViewModel.kt
@@ -37,9 +37,7 @@ internal class AltTextViewModel(
 
     private fun getAvatarData() {
         viewModelScope.launch {
-            val avatar = avatarRepository.getAvatars(
-                Email(email),
-            ).replayCache.firstOrNull()?.firstOrNull { it.imageId == avatarId }
+            val avatar = avatarRepository.getAvatar(Email(email), avatarId)
             if (avatar != null) {
                 originalAltText = avatar.altText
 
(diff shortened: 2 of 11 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextAction.kt
```diff
@@ -4,4 +4,6 @@ internal sealed class AltTextAction {
     data object AltTextUpdated : AltTextAction()
 
     data object AltTextUpdateFailed : AltTextAction()
+
+    data object AvatarCantBeLoaded : AltTextAction()
```
hamorillo, line 8: This case shouldn't occur in the QE, but just in case and as a way to handle it.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextViewModel.kt
```diff
@@ -19,18 +19,42 @@ import kotlinx.coroutines.launch
...
+
+    init {
+        getAvatarData()
+    }
+
+    private fun getAvatarData() {
+        viewModelScope.launch {
+            val avatar = avatarRepository.getAvatars(
+                Email(email),
+            ).replayCache.firstOrNull()?.firstOrNull { it.imageId == avatarId }
```
AdamGrzybkowski: 💡 We could have that as a function to get Avatar based on the id in the repository.

gravatar-quickeditor/src/test/java/com/gravatar/quickeditor/ui/alttext/AltTextViewModelTest.kt
```diff
@@ -84,7 +98,7 @@ class AltTextViewModelTest {
             viewModel.onEvent(AltTextEvent.AvatarAltTextSaveTapped)
 
             viewModel.uiState.test {
-                skipItems(2)
+                skipItems(3) // Skipping the following values: initial, init block and alt text change event
```
AdamGrzybkowski: I suggested a different approach to this [here](https://github.com/Automattic/Gravatar-SDK-Android/pull/529#discussion_r1910171557) - I'm not a fan of those `skipItems` because tests may break when some logic changes that are not related to this particular test.

hamorillo: I don't have a strong opinion here. I see benefits in both approaches, but I've refactored the code. 🙂

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/alttext/AltTextViewModel.kt
hamorillo: Great idea! Done.

AdamGrzybkowski (approved):
:shipit:

PR #532, opened 2025-01-10
Change screenshot threshold to 0.005

AdamGrzybkowski:
### Description

This [PR](https://github.com/Automattic/Gravatar-SDK-Android/pull/528) should fail but didn't so I found out the the 0,01 threshold was to big. 

It's a bit weird because the change from that PR seems to be quite significant, but 0.005 made it fail so...

### Testing Steps

The only way is to apply this change to the mentioned PR and verify the screenshot tests.


mlumeau (approved):
Tested this on the mentioned PR and it fixed the screenshots tests 👍

PR #534, opened 2025-01-13
Fix the QE recomposition when selecting an avatar

AdamGrzybkowski:
### Description

There's a bug that's causing the QE to reload on Avatar change. The problem is this [line](https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt#L188). When an avatar is selected the `onAvatarSelected` lambda is invoked and this triggered the recomposition caused by the changed `startDestination`. 

Changing the start destination was convenient but not necessary, so I reworked the code to not require it. 

### Testing Steps

1. Launch the QE
2. Select an avatar
3. Confirm the QE did not recompose
4.  Test for regression in the back button navigation - Go through the OAuth flow, Alt text page, etc.

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
@@ -54,9 +54,9 @@ internal fun GravatarQuickEditorPage(
         composable(route = QuickEditorPage.SPLASH.name) {
             SplashPage(email = gravatarQuickEditorParams.email) { isAuthorized ->
                 if (isAuthorized) {
-                    navController.navigateAndClean(QuickEditorPage.EDITOR.name)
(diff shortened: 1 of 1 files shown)


hamorillo (commented):
I've tested the PR, it works as expected.

> When an avatar is selected the onAvatarSelected lambda is invoked and this triggered the recomposition caused by the changed startDestination.

Can you explain this a little bit more, please. I don't fully understand why that was happening.

AdamGrzybkowski:
>Can you explain this a little bit more, please. I don't fully understand why that was happening.

That's a good question. I only have a guess but no concrete answer. The `startDestination` is part of the `Graph` that is a param in a `NavHost` Composable. The `onAvatarSelected` is a lambda passed from the very top that goes through all the composables and must be invoking the recomposition due to a modified `Graph` state on its way up.

hamorillo (approved):
🚀

PR #536, opened 2025-01-15
Make sure the save button can always be reached

### LINKED ISSUE #533: Alt Text: Save button can be moved outside of the visible area
https://github.com/Automattic/Gravatar-SDK-Android/pull/529#pullrequestreview-2542491079

AdamGrzybkowski:
Closes #533

### Description

The Save button could be pushed away from the screen. It was not possible to reach it. There were two reasons for that:
1. The Input field could grow its height without any limits
2. The whole page was not scrollable, making it impossible to scroll to it

I applied changes to limit the input field height (5 lines) and make the whole page scrollable. 

Here's the video with both font and display scaled up a lot.

https://github.com/user-attachments/assets/25f48ea4-e245-4172-8232-5e47beae2c15

### Testing Steps

1. Open the Alt text page
2. Type enter a few times to increase the number of lines
3. Confirm you can always reach the Save button
4. Feel free to play with the font&display scaling


hamorillo (approved):
Nice improvement! 🚀

PR #537, opened 2025-01-15
Quick Editor: AltText - Cancel button in TopBar

### LINKED ISSUE #496: Quick Editor: Alt Text - Cancel button in the header of the bottom sheet

hamorillo:
Closes #496 

### Description
As shown in the following image, the AltText section should use a different label for the top bar left's button:

image

The TopBar was shared between all possible sections inside `GravatarQuickEditorBottomSheet,` making the customization of the callback / UI in a concrete section complex. It was convenient before the AltText section, but it is not anymore. 

The proposed solution extracts the TopBar section from the shared BottomSheet and lets each section add its own TopBar. With this approach, we can have a customizable composable.

### Testing Steps

- In the QE, open the AltText for an avatar
- Verify the TopBar contains a "Cancel" button instead of the generic "Done" button.
- Click the "Cancel" button
- Verify the AltText page is closed and you are redirected to the AvatarPicker
--- 
- Smoke test the whole QE, paying special attention to the TopBar (OAuth, Splash, Avatar Picker...)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QETopBar.kt
```diff
@@ -16,16 +17,45 @@ import androidx.compose.material3.TextButton
...
 import com.gravatar.ui.GravatarTheme
 
 @Composable
-internal fun QETopBar(onDoneClick: () -> Unit, modifier: Modifier = Modifier, onGravatarIconClick: () -> Unit = {}) {
+internal fun QETopBarWithContent(
+    onLeftButtonClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    leftButtonLabel: String = stringResource(R.string.gravatar_qe_bottom_sheet_done),
+    content: @Composable () -> Unit = {},
+) {
```
AdamGrzybkowski: ❓ I'm not sure the slot API is a good choice here. This serves well when you assign certain spaces for item in the TopBar, but the content here is not the scope of the TopBar at all. Any reasons you went with this? 

I imagine that it's likely to avoid doing this
```Column { TopBar ...```

but we could achieve this by having a `QEPage` component that would make this, rather than making the TopBar aware of the content at all.

File: gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QETopBar.kt [hunk 2]
```diff
@@ -16,16 +17,45 @@ import androidx.compose.material3.TextButton
...
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
+import com.gravatar.GravatarConstants
 import com.gravatar.quickeditor.R
 import com.gravatar.ui.GravatarTheme
 
 @Composable
-internal fun QETopBar(onDoneClick: () -> Unit, modifier: Modifier = Modifier, onGravatarIconClick: () -> Unit = {}) {
+internal fun QETopBarWithContent(
+    onLeftButtonClick: () -> Unit,
```
AdamGrzybkowski: 💡 I would consider replacing this with 

```suggestion
    leftButton: @Composable () -> Unit,
```
Alternative name: `upButton`.

This would be more flexible in the future if we wanted to use icons instead of text buttons.

hamorillo: I considered this but decided not to do it because it looked like an early improvement we currently don't need. 

Anyway, as you've mentioned, I'll work on the refactor. 🙂

AdamGrzybkowski: Sure, this one was just a suggestion since it's very easy to end up with a ton of params in Compose. So exposing a slot helps with that. No need to fix this one for this PR.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QETopBar.kt
hamorillo: I've just pushed some changes related to this. Let me know what you think.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QETopBar.kt
hamorillo: I've changed it. :)

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QEPage.kt
```diff
@@ -4,6 +4,21 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 
+@Composable
+internal fun QEPageDefault(onDoneClicked: () -> Unit, content: @Composable () -> Unit, modifier: Modifier = Modifier) {
```
AdamGrzybkowski, line 8: I wouldn't mind keeping the same name `QEPage` that just has different params. I think this is quite common in Compose.

AdamGrzybkowski (approved):
Works great, thanks for the changes!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/QEPage.kt
hamorillo: The problem is that the function signature is the same under the hood `() -> Unit, () -> Unit, Modifier`. That's why I need to use a different name.

PR #538, opened 2025-01-15
Document permissions used by the QuickEditor

### LINKED ISSUE #505: Document permissions used by the Quick Editor
Describe which permissions are used and why.

AdamGrzybkowski:
Closes #505

### Description

I believe it's good to be transparent about the permission we define in the QuickEditor's manifest file. This PR adds a section showing what permissions we use and why. 

Screenshot 2025-01-15 at 15 16 38

1. Is there anything else we should add? 
2. Do you think there's a better place for it?

### Testing Steps

Read the docs.

### DIFF
--- a/docs/get-started/get-started.md
+++ b/docs/get-started/get-started.md
@@ -500,3 +500,14 @@ public fun Avatar(
 ```
 
 By setting `forceRefresh` to true, you ensure that the avatar is always fetched with the latest changes.
+
+### Android Permission
+
+The Quick Editor module requires certain permissions to function correctly. Below is a table listing the permissions and the reasons why they are needed:
+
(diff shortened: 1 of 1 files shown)


hamorillo (approved):
Nice idea! 

> Is there anything else we should add?

Nothing that comes into my mind.

> Do you think there's a better place for it?

I think that's a good place, but I have to recognize that I find the information in the GetStarted file difficult to read. Maybe it's because we have a mix of the core library, UI, and QE. 

Maybe we can leave the GetStarted file for basic shared information and split each module's concrete details into a separate file. 

As that reflection is not related to this PR :shipit:, we can discuss this in a different place.

PR #540, opened 2025-01-17
Make working with cache busting easier

### LINKED ISSUE #514: QuickEditor: Avatar in the ProfileCard reload when coming back from a different destination
https://github.com/user-attachments/assets/f5c5f510-ce0c-44c5-88de-5c01fd1073b8

AdamGrzybkowski:
Closes #514

### Description

This PR fixed the problem with an extra Avatar image reload when coming back from the AltText page. It was caused by the `forceRefresh` param set to true on the Avatar composable. On each composition, it would reload the image by creating a new cache buster param. 

The fix was to extract the cache-busting param update logic to ViewModel and only change it when a new avatar should be loaded. 

The above required a new `Avatar` component that takes a raw URL (in the form of a String). An alternative would be to keep the current `Avatar` component that takes `Profile` model and add an extra param for the cache buster, but in general I think we should only ask for what's needed, because passing the full `Profile` model may cause extra recompositions of the `Avatar` component  when stuff unrelated to the Avatar changes. 

| Before | After |
|---|---|
| reload | no_reload |

### Testing Steps

1. Launch the QE
2. Make sure you have at least one avatar uploaded
3. Tap on the `...` avatar's button
4. Tap `Alt text`
5. Tap `Cancel` or use the system back button
6. Make sure the avatar in the profile card doesn't blink

### DIFF
--- a/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
+++ b/gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
@@ -94,6 +94,30 @@ public fun Avatar(
     }
 }
 
+/**
+ * [Avatar] is a composable that displays a user's avatar.
+ *
+ * @param state The state of the avatar, when loaded it should contain the Avatar URL
+ * @param size The size of the avatar
+ * @param modifier Composable modifier
+ */
+@Composable
+public fun Avatar(state: ComponentState<String>, size: Dp, modifier: Modifier = Modifier) {
+    when (state) {
+        is ComponentState.Loading -> SkeletonAvatar(size = size, modifier = modifier)
+
+        is ComponentState.Loaded -> {
+            Avatar(
+                model = state.loadedValue,
+                size = size,
+                modifier = modifier,
+            )
+        }
+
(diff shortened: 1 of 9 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
```diff
@@ -60,6 +60,26 @@ public fun Avatar(
...
+    size: Dp,
+    modifier: Modifier = Modifier,
+) {
+    Avatar(
+        model = avatarURL,
+        size = size,
+        modifier = modifier,
+    )
+}
+
```
hamorillo: 🤔 I think we don't need this one.

AdamGrzybkowski: yes!

mlumeau (approved):
LGTM. Works well, code changes are OK, I think it even makes the cache busting logic easier to understand.

PR #541, opened 2025-01-21
QuickEditor: Make sure the QE works well when the app get's killed during the OAuth flow

### LINKED ISSUE #525: Quick Editor: OAuth approval is required twice the first time an e-mail is used
https://github.com/user-attachments/assets/a13eaca1-cb3d-478f-b870-a9b8d2cc78b0

AdamGrzybkowski:
Closes #525

### Description

The current code has a problem with the OAuth flow when the app get's killed by the system (or when the "Don't keep activities" setting is ON). Here's how Hector [described it](https://github.com/Automattic/Gravatar-SDK-Android/issues/525#issuecomment-2595700254): 

> ...
>
>The culprit of the problem is that we add the [OnNewIntentListener in the OAuthPage](https://github.com/Automattic/Gravatar-SDK-Android/blob/ca5b79ac430c5e278962f1d88f966f3b090d0062/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthPage.kt#L81) so while that composable and the activity where it's contained is alive there is no problem.
>
>However, when the activity is destroyed after opening the CustomTab, it will be recreated after the OAuthFlow has finished. In the MainActivity example, when the OAuthFlow finishes, the steps are:
>
>MainActivity.onCreate()
MainActivity.onNewIntent()
The onCreate method recreates the activity as it was, keeping the QE with the OAuthScreen opened, but the onNewIntent is called before the composables are inflated and the listener in the OAuthPage composable registered, so we are missing it.
>
>We can, somehow, try to resolve this in an activity under our control (GravatarQuickEditorActivity), but I think it is more complicated in a third-party app as (MainActivity).

So far, the main QE's flow would make the third-party activity handle the deep linking for the OAuth flow, so there was very little we could do to fix this. Hector explored a suggestion that maybe we could achieve this by using Deep Linking via Jetpack Compose Navigation but our NavGraph lives only in the Bottom Sheet, not the activity so making it work seems equally tricky. 

This PR proposes a different approach that not only fixes this issue #525 but also **removes the requirement from third-party apps to use `singleTask` launch mode on the activity that launches the QuickEditor.**

The solution:

Instead of relying on third-party activity to handle the OAuth deep links, the QuickEditor opens a fully transparent activity `GravatarOAuthActivity` before launching the Custom Tab to start the OAuth. From now on this is the only Activity that requires the `launchMode=singleTask`. This activity also handles the `onNewIntent` and we no longer need to handle that within the Composable so the whole timing issue described by Hector is no longer a problem.

I had to set a local flag to make sure the `GravatarOAuthActivity` is finished when the Custom tab is dismissed (without it the QE would appear as focused but wouldn't actually work because it would be under the activity).

Additionally, I had to add `SavedStateHandle` to the `OAuthViewModel` so as not to send the `StartOauth` action in the init block. When the activity is recreated, so is the ViewModel, so we need to make sure that action hasn't been emitted before. 

Here's the result:

https://github.com/user-attachments/assets/9deb049f-c992-4c02-850d-dbbe54ea737b

Once this is merged we can deprecate the `GravatarQuickEditorActivity` as its only purpose was so that third parties don't have to set `singleTask` on their Activity. We will have to adjust the docs as well. 

### Testing Steps

1. Go to the Developer settings and enable the "Don't keep activities" - you can also play by killing the app via ADB
2. Fresh install the Demo app
3. Launch the QE
4. Dismitt the custom tab via `x` in the Toolbar
5. Confirm you see the QE with the intro page
6. Tap `Continue`
7. Tap `Deny`
8. Confirm the QE was closed and error toast was shown
9. Launch the QE
10. Tap `Approve`
11. Confirm you were logged in and you can see your Avatars
12. Disable the "Don't keep activities" setting
13. Repeat the above steps, and make sure all works as expected

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/GravatarOAuthActivity.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/GravatarOAuthActivity.kt
@@ -1,0 +1,173 @@
+package com.gravatar.quickeditor.ui.oauth
+
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import android.os.Bundle
+import androidx.activity.result.contract.ActivityResultContract
+import androidx.appcompat.app.AppCompatActivity
+import androidx.browser.customtabs.CustomTabsIntent
+import com.gravatar.quickeditor.ui.GravatarQuickEditorActivity
+import com.gravatar.types.Email
+import java.net.URLDecoder
+
+/**
+ * Activity to handle OAuth authentication with Gravatar.
+ *
+ * This activity launches a custom tab for the user to authenticate with Gravatar.
+ * Once the authentication is complete, it retrieves the access token from the redirect URI
+ * and returns it as a result.
+ *
+ * It's used internally, but it's exposed as a public as it needs to be declared in your AndroidManifest.xml file
+ * with the proper deep linking setup.
+ */
+public class GravatarOAuthActivity : AppCompatActivity() {
+    private var oAuthStarted = false
+    private lateinit var clientId: String
+    private lateinit var redirectUri: String
+    private lateinit var email: String
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        clientId = intent.getStringExtra(CLIENT_ID_KEY)!!
+        redirectUri = intent.getStringExtra(REDIRECT_URI_KEY)!!
+        email = intent.getStringExtra(EMAIL_KEY)!!
+
+        oAuthStarted = savedInstanceState?.getBoolean(OAUTH_STARTED_KEY) ?: false
+
+        addOnNewIntentListener { newIntent ->
+            val token = newIntent.data
+                ?.encodedFragment
+                ?.split("&")
+                ?.associate {
+                    val split = it.split("=")
+                    split.first() to split.last()
+                }
+                ?.get("access_token")
+                ?.let { URLDecoder.decode(it, "UTF-8") }
+
+            if (token != null) {
+                val resultIntent = Intent().apply {
+                    putExtra(ACTIVITY_RESULT, RESULT_TOKEN_RETRIEVED)
+                    putExtra(TOKEN_KEY, token)
+                }
+
+                setResult(RESULT_OK, resultIntent)
+            } else {
+                setResult(
+                    RESULT_OK,
+                    Intent().apply {
+                        putExtra(ACTIVITY_RESULT, RESULT_TOKEN_ERROR)
+                    },
+                )
+            }
+            finish()
+        }
+    }
+
+    override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+
+        outState.putBoolean(OAUTH_STARTED_KEY, oAuthStarted)
+    }
+
+    override fun onResume() {
+        super.onResume()
+
+        if (!oAuthStarted) {
+            launchCustomTab(
+                this,
+                clientId = clientId,
+                redirectUri = redirectUri,
+                email = Email(email),
+            )
+            oAuthStarted = true
+        } else {
+            setResult(
+                RESULT_OK,
+                Intent().apply {
+                    putExtra(ACTIVITY_RESULT, RESULT_CANCELED)
+                },
+            )
+            finish()
+        }
+    }
+
+    private fun launchCustomTab(context: Context, clientId: String, redirectUri: String, email: Email) {
+        val customTabIntent: CustomTabsIntent = CustomTabsIntent.Builder()
+            .build()
+        customTabIntent.launchUrl(
+            context,
+            Uri.parse(WordPressOauth.buildUrl(clientId, redirectUri, email)),
+        )
+    }
+
+    internal companion object {
+        private const val CLIENT_ID_KEY = "client_id"
+        private const val REDIRECT_URI_KEY = "redirect_uri"
+        private const val EMAIL_KEY = "email"
+        private const val OAUTH_STARTED_KEY = "oauth_started"
+        internal const val TOKEN_KEY = "auth_token"
+
+        const val ACTIVITY_RESULT: String = "oAuthActivityResult"
+        const val RESULT_CANCELED: Int = 1000
+        const val RESULT_TOKEN_RETRIEVED: Int = 1001
+        const val RESULT_TOKEN_ERROR: Int = 1002
+
+        internal fun createIntent(context: Context, oAuthParams: OAuthParams, email: String): Intent {
+            return Intent(context, GravatarOAuthActivity::class.java).apply {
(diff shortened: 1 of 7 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/GravatarOAuthActivity.kt
```diff
@@ -0,0 +1,173 @@
...
+public class GravatarOAuthActivity : AppCompatActivity() {
+    private var oAuthStarted = false
+    private lateinit var clientId: String
+    private lateinit var redirectUri: String
+    private lateinit var email: String
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        clientId = intent.getStringExtra(CLIENT_ID_KEY)!!
```
mlumeau: I was able to trigger a `NullPointerException` here by opening the customtab url in a browser tab, finishing the flow, and going back to that tab later on which tried to open the app again through the deep link.

Is there a way for us to handle these edge cases more gracefully?

AdamGrzybkowski: >I was able to trigger a NullPointerException here by opening the customtab url in a browser tab, finishing the flow, and going back to that tab later on which tried to open the app again through the deep link.

Hmm I'm not sure I got the steps correctly. Can you clarify what do you mean by `by opening the customtab url in a browser tab`? Or better, post a screen recording.

>Is there a way for us to handle these edge cases more gracefully?

We could close the Activity - it can only work if those params are set.

AdamGrzybkowski:
@mlumeau I pushed the commit that makes this change backward-compatible. You can test it by applying this patch 
[manifest.patch](https://github.com/user-attachments/files/18506378/manifest.patch) - this will make the third-party activity handle the `onNewIntent`. 

If you change nothing, you can verify this still works as expected and this warning log `Setup GravatarOAuthActivity in your AndroidManifest.xml to handle the OAuth flow.` is not logged.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/GravatarOAuthActivity.kt
mlumeau: Here's a screen recording, sorry for the French language, but I think it's explicit:

https://github.com/user-attachments/assets/2afbc37b-0a9a-46ca-964f-9f54f887227f

Here you can see at the end that I'm going back to Firefox, refresh the page (which happens automatically if I open Firefox from the launcher) and the demo app flashes and closes down abruptly, I have this stack trace in the log:

```
FATAL EXCEPTION: main (Ask Gemini)
Process: com.gravatar.demoapp, PID: 16745
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.gravatar.demoapp/com.gravatar.quickeditor.ui.oauth.GravatarOAuthActivity}: java.lang.NullPointerException
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:4129)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:4316)
	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:222)
	at android.app.servertransaction.TransactionExecutor.executeNonLifecycleItem(TransactionExecutor.java:133)
	at android.app.servertransaction.TransactionExecutor.executeTransactionItems(TransactionExecutor.java:103)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:80)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2719)
	at android.os.Handler.dispatchMessage(Handler.java:109)
	at android.os.Looper.loopOnce(Looper.java:232)
	at android.os.Looper.loop(Looper.java:317)
	at android.app.ActivityThread.main(ActivityThread.java:8787)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:591)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:871)
Caused by: java.lang.NullPointerException
	at com.gravatar.quickeditor.ui.oauth.GravatarOAuthActivity.onCreate(GravatarOAuthActivity.kt:33)
	at android.app.Activity.performCreate(Activity.java:9019)
	at android.app.Activity.performCreate(Activity.java:8997)
	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1528)
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:4111)
```

AdamGrzybkowski: Ohh ok, now it's clear. You go back to the wp.com OAuth site after successfully finishing the flow early. Yeah, I think what I mentioned before will work for this. This is a weird edge case, thanks for testing!

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthPage.kt
```diff
@@ -63,23 +63,23 @@ internal fun OAuthPage(
...
     }
 
+    // Kept for backwards-compatibility.
+    // This will be removed in the future and the GravatarOAuthActivity should be used instead.
     if (activity != null) {
         DisposableEffect(Unit) {
             val listener = Consumer<Intent> { newIntent ->
+                Log.w(
+                    "QuickEditor",
+                    "Setup GravatarOAuthActivity in your AndroidManifest.xml to handle the OAuth flow.",
```
mlumeau: It works, but I'd be a little more explicit:

```suggestion
                    "GRAVATAR SDK WARNING: You're using a deprecated version of the Gravatar QuickEditor OAuth flow. Set up GravatarOAuthActivity in your AndroidManifest.xml to handle the OAuth flow and to remove this warning message. ",
```

Just a suggestion, feel free to improve this message.

AdamGrzybkowski: Looks good to me! A few UPPER CASE characters as a warning will help! :D

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/GravatarOAuthActivity.kt
AdamGrzybkowski: I pushed the changes to fix this NPE.

AdamGrzybkowski:
@mlumeau I removed the enter transition to be on the safe side. I don't think it was visible because of the Custom tab but this should get rid of it regardless. Sorry for the last minute update.

mlumeau (approved):
LGTM! 🚀

PR #544, opened 2025-01-23
Deprecate GravatarQuickEditorActivity

### LINKED ISSUE #542: QuickEditor: Deprecate GravatarQuickEditorActivity
Its only purpose was not to forace third-party app to set one of their activities to `singleTask`. Now that we will use `GravatarOAuthActivity` for the OAuth flow, the `singleTask` need to be set only on this activity so the `GravatarQuickEditorActivity` serves no purpose.

AdamGrzybkowski:
Closes #542

### Description

Now that [this PR](https://github.com/Automattic/Gravatar-SDK-Android/pull/541) was merged we can deprecate the `GravatarQuickEditorActivity`. Its only purpose was not to force third-party apps to set one of their activities to singleTask. Now that we will use `GravatarOAuthActivity` for the OAuth flow, the singleTask needs to be set only on this activity so the `GravatarQuickEditorActivity` serves no purpose.

I will update docs in a separate PRs. 

### Testing Steps

None, really.


mlumeau (approved):
LGTM 😉

PR #557, opened 2025-01-24
Set the cache buster value in the initial state

### LINKED ISSUE #548: QuickEditor: Old image shown on initial QE launch
The avatar is shown in the profile card after closing/reopening the QE even though there is no selected avatar.

AdamGrzybkowski:
Closes #548 

### Description

The cache buster param is not set when initializing the `AvatarPickerViewModel` so we do load a cached image. This PR sets the value in the state initialization to fix that.

| Before | After |
|---|---|
| cached_avatar_before | avatar_cache_after |

### Testing Steps

1. Launch the QE
2. Select a different avatar
3. Reopen the QE
4. Confirm the Avatar in the Profile card is up to date

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/avatarpicker/AvatarPickerViewModel.kt
@@ -46,7 +46,13 @@ internal class AvatarPickerViewModel(
     private val clock: Clock,
 ) : ViewModel() {
     private val _uiState =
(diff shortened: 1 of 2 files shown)


mlumeau:
Wouldn't it be simpler to set the default value in the `AvatarPickerUiState` to `0` instead of `null`?

AdamGrzybkowski:
>Wouldn't it be simpler to set the default value in the AvatarPickerUiState to 0 instead of null?

0 is not a great cache buster, because it's not unique. If you try to do the testing steps quickly, you will see it doesn't fix the problem.

mlumeau (approved):
Couldn't reproduce the issue with these changes so I guess it fixed it 👍

PR #558, opened 2025-01-24
Show all avatars in the Profile card regardless of the rating

### LINKED ISSUE #549: QuickEditor:  The profile card doesn’t show the lower rating avatars.
The profile card doesn’t show the lower-rating avatars. It was fixed on iOS by adding the rating query param(as explicit) to the avatar fetch call.

AdamGrzybkowski:
Closes #549

### Description

The profile card in the QE wouldn't show avatars that are not G-rated. This fixes the problem by updating the request URL param.

| Before | After |
|---|---|
| avatar_rating_before | avatar_rating_after |

### Testing Steps

1. Launch the QE 
2. Set one Avatar's rating to X
3. Select that avatar
4. Confirm you can see it displayed in the Profile card

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/ProfileCard.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/ProfileCard.kt
@@ -20,6 +20,7 @@ import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.gravatar.AvatarQueryOptions
+import com.gravatar.ImageRating
 import com.gravatar.extensions.avatarUrl
 import com.gravatar.extensions.defaultProfile
 import com.gravatar.restapi.models.Profile
@@ -50,6 +51,7 @@ internal fun ProfileCard(
                             avatarUrl(
                                 AvatarQueryOptions {
                                     preferredSize = sizePx
+                                    rating = ImageRating.X
                                 },
                             ).url(avatarCacheBuster).toString()
                         },


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/ProfileCard.kt
```diff
@@ -50,6 +51,7 @@ internal fun ProfileCard(
                             avatarUrl(
                                 AvatarQueryOptions {
                                     preferredSize = sizePx
+                                    rating = ImageRating.X
```
AdamGrzybkowski, line 54: @pinarol Is that the rating you're using in the iOS SDK?

pinarol: Yep.

mlumeau: I suggest we rename `rating` to `maxRating` in `AvatarQueryOptions`. Currently, I think it's a bit confusing that devs may need to explicitly set the `rating` to `X` in a user's profile card, even though the avatar won't have this rating 99% of the time.

AdamGrzybkowski: I agree about it being confusing, but that would be a breaking change. We can't simply rename it. 

As far as I'm aware the naming matches the URL query params on purpose and [the docs](https://docs.gravatar.com/api/avatars/images/) mention how this works.

mlumeau (approved):
LGTM

PR #559, opened 2025-01-24
Location component empty state placeholder

### LINKED ISSUE #555: Update ProfileSummary empty text placeholder
Image

AdamGrzybkowski:
Closes #555

### Description

The placeholder copy in the `Location` component is confusing as it mentioned fields that are not displayed by this component. We're updating it to be just "Location".

### Testing Steps

1. Launch the QE with a random email that shouldn't have a profile
2. Confirm you see the new placeholder in the Profile card


gravatar-ui/src/main/res/values/strings.xml
```diff
@@ -5,4 +5,5 @@
     <string name="gravatar_ui_empty_state_display_name">Your Name</string>
     <string name="gravatar_ui_empty_state_user_info">Add your location, pronouns, etc</string>
```
mlumeau, line 6: Do we still need that string?

AdamGrzybkowski: Yes, it's used in the `UserInfo` component.

mlumeau: I see, but it's used for the same purpose, right? shouldn't we use the new string in that case too?

AdamGrzybkowski: I don't think so. It was a problem in the `Location` component because the location component doesn't show information like `pronouns`, it only shows the location. `UserInfo` does show those fields, including location.

mlumeau (approved):
LGTM!

PR #566, opened 2025-01-28
Implement extension function to remove orphan words

### LINKED ISSUE #565: Orphans in text fields
The default Text component doesn't handle orphan words. We should fix that.

AdamGrzybkowski:
Closes #565 

### Description

This PR adds a String extension that inserts a non-breaking space before the last word to make sure no orphan words are left in the last line. It's used in the `QESectionMessage`

### Testing Steps

1. Launch the QE
2. In the Avatar picker play with the font scaling and make sure the text `Choose or upload your favorite avatar images and connect them to your email address.` will never have orphans.


mlumeau:
I like this in principle, but I'm not sure we need a new composable for that. We could just keep the extension function from this PR and apply it to every `Text` where we need it? 
What if the Text composable evolves and adds new attributes in the future? We'll need to keep up then.

AdamGrzybkowski:
> What if the Text composable evolves and adds new attributes in the future? We'll need to keep up then.

That makes sense, I will remove it.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/StringExtensions.kt
```diff
@@ -0,0 +1,8 @@
+package com.gravatar.quickeditor.ui
+
+internal val String.removeOrphans: String
```
mlumeau, line 3: Can we add a unit test for this function?

AdamGrzybkowski: Sure!

mlumeau (approved):
LGTM!

PR #568, opened 2025-02-04
Fix: Avatar is not visible when the profile is disabled but the Avatar is set to public

### LINKED ISSUE #567: Make sure the Avatar is shown even if the profile is private

AdamGrzybkowski:
Closes #567

### Description

This PR adjusts the `ProfileCard` component in the QuickEditor to show the Avatar when it's set to public even if the profile is set to private. 

To accomplish this I had to create a new `Avatar` component that takes the `Email` and not the `ComponentState`. It now relies on the Coil callback to show either the Loading or Error state. Loading is for the Skeleton and the Error is for the empty avatar image. Here's how it looks now with a private profile: 

| Avatar private | Avatar public |
|---|---|
| Screenshot 2025-02-04 at 13 45 32 | Screenshot 2025-02-04 at 13 45 46 |

This won't be the default behavior of all the Profile cards from the `:ui` module though, due to the `ComponentState` - we can only get the email from the profile once it's loaded, and at that point, this is a bit late. Even if we wanted to fall back when a profile can't be loaded (ComponentState.Empty) we wouldn't have any profile information. So that would require some API updates. 

### Testing Steps

1. Go to your Gravatar profile website 
2. Open Account Setting and switch to privacy
3. Uncheck the `public profile` but keep the avatar public
4. Launch the Demo app
5. Tap Log out
6. Open the QE 
7. Confirm you can see the Avatar but not the profile
8. Uncheck the `public avatar` settings
9. Reopen the QE
10. Confirm you can't see the avatar


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/oauth/OAuthPage.kt
```diff
@@ -172,7 +172,9 @@ internal fun OauthPage(
                         uiState.profile?.let {
                             ProfileCard(
                                 profile = it,
+                                email = email,
                                 modifier = Modifier.padding(top = 16.dp),
+                                avatarCacheBuster = uiState.avatarCacheBuster,
```
AdamGrzybkowski, line 177: We probably don't have to use the cache buster on this page, but:
- it makes testing easier, and
- it makes sure the the QE always shows the latest and correct avatar

gravatar-ui/src/main/java/com/gravatar/ui/components/atomic/Avatar.kt
```diff
@@ -144,11 +151,70 @@ private fun EmptyAvatar(size: Dp, modifier: Modifier = Modifier) {
 private fun Avatar(model: Any?, size: Dp, modifier: Modifier) {
     AsyncImage(
         model = model,
-        contentDescription = "User profile image",
```
AdamGrzybkowski, line 147: This was hardcoded in the component, we could probably skip the waiting for the translation for this one, as it will delay the release again.

mlumeau (approved):
It works! Code changes LGTM too.

PR #577, opened 2025-02-13
Bump target&compile SDK to 35 in the demo app

AdamGrzybkowski:
### Description

This should help us catch issues like the one in the uCrop with edge-to-edge enforcement. 

### Testing Steps

Build the demo app and test it.

### DIFF
--- a/demo-app/src/main/AndroidManifest.xml
+++ b/demo-app/src/main/AndroidManifest.xml
@@ -78,12 +78,6 @@
             </intent-filter>
         </activity>
 
-
-        <!-- Lib activities -->
-        <activity
-            android:name="com.yalantis.ucrop.UCropActivity"
-            android:theme="@style/Theme.AppCompat" />
-
         <provider
             android:name=".DemoFileProvider"
             android:authorities="${applicationId}.fileprovider"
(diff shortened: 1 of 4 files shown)


demo-app/src/main/AndroidManifest.xml
```diff
@@ -78,12 +78,6 @@
             </intent-filter>
         </activity>
 
-
-        <!-- Lib activities -->
-        <activity
-            android:name="com.yalantis.ucrop.UCropActivity"
-            android:theme="@style/Theme.AppCompat" />
-
```
AdamGrzybkowski, line 86: This is not needed, so removing to clean things up.

demo-app/src/main/java/com/gravatar/demoapp/theme/Theme.kt
```diff
@@ -59,7 +58,6 @@ fun GravatarDemoAppTheme(
     if (!view.isInEditMode) {
         SideEffect {
             val window = (view.context as Activity).window
-            window.statusBarColor = colorScheme.primary.toArgb()
```
AdamGrzybkowski, line 62: Deprecated.

demo-app/src/main/java/com/gravatar/demoapp/ui/DemoGravatarApp.kt
```diff
@@ -151,7 +149,7 @@ private fun GravatarTabs(
         stringResource(R.string.tab_label_avatar_update),
     )
 
-    Column(modifier = Modifier.fillMaxSize()) {
+    Column(modifier = modifier.fillMaxSize()) {
```
AdamGrzybkowski, line 152: I noticed this didn't work well after switching to 35. We should always pass the modifier to the Top level composable, not its children.

PR #586, opened 2025-02-20
Add publishing.gradle.kts for Maven Central

AdamGrzybkowski:
Closes #

### DIFF
--- a/gradle/publication.gradle.kts
+++ b/gradle/publication.gradle.kts
@@ -1,0 +1,106 @@
+apply(plugin = "maven-publish")
+apply(plugin = "signing")
+
+extra["contralSonatypeBearer"] = System.getenv("CENTRAL_SONATYPE_BEARER")
+extra["signingKeyId"] = System.getenv("SIGNING_KEY_ID")
+extra["signingKey"] = System.getenv("SIGNING_KEY")
+extra["signingPassword"] = System.getenv("SIGNING_PASSWORD")
+
+afterEvaluate {
+    configure<PublishingExtension> {
+        repositories {
+            maven {
+                val contralSonatypeBearer = extra["contralSonatypeBearer"] as? String
+                if (contralSonatypeBearer == null || contralSonatypeBearer.isEmpty()) {
+                    logger.warn("CENTRAL_SONATYPE_BEARER is not set.")
+                }
+                name = "mavenCentral"
+                url = uri("https://central.sonatype.com/api/v1/publisher/deployments/download/")
+                credentials(HttpHeaderCredentials::class) {
+                    name = "Authorization"
+                    value = "Bearer $contralSonatypeBearer"
+                }
+                authentication {
+                    create("header", HttpHeaderAuthentication::class)
+                }
+            }
+//            maven {
+//                name = "s3"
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">This PR is tagged with <code>do not merge</code> label(s).</td>
    </tr>
  </tbody>
</table>

</p>

gradle/publication.gradle.kts
```diff
@@ -0,0 +1,89 @@
...
+        publications {
+            val moduleArtifactId: String? = extra["artifactId"] as? String
+            create<MavenPublication>("maven") {
+                from(components["release"])
+
+                groupId = "com.gravatar"
+                artifactId = moduleArtifactId
+                // version is set by `publish-to-s3` plugin
+            }
+            create<MavenPublication>("mavenCentral") {
```
AdamGrzybkowski: Hi @AliSoftware 

I'm working on the Maven Central config but I hoped that I could keep the publish-to-s3 so we can have `test` versions published there. So far this is not possible because the plugin tries to publish all the publications. Have you done something like this? Does it even make sense?

AliSoftware: Oof not really my area of expertise, tbh @ParaskP7 or @wzieba would be more likely to be helpful here 😅

wzieba: > because the plugin tries to publish all the publications

I understand you mean our plugin: [publish-to-s3](https://github.com/Automattic/publish-to-s3-gradle-plugin). In this case, I think we could update logic of our `publish-to-s3` or have the following logic in Gravatar config:
- if this is release (building on tag): add Maven Central publication only
- else: add S3 publication only

I think this is fine, especially that Maven Central has staging, so you can always test the SDK before releasing.

AdamGrzybkowski: >I understand you mean our plugin: [publish-to-s3](https://github.com/Automattic/publish-to-s3-gradle-plugin)

Yes, I found this [here](https://github.com/Automattic/publish-to-s3-gradle-plugin/blob/trunk/plugin/src/main/kotlin/com/automattic/android/publish/ProjectExtensions.kt#L42-L44).

>I think we could update logic of our publish-to-s3 or have the following logic in Gravatar config:

The problem I see is that when the `publish-to-s3` plugin is applied it will set `PublishToMavenRepository` to depend on `prepareToPublishToS3`. This seems to be causing trouble when running publish to maven commands directly. For example when I try to run: `gw publishToMavenLocal` I will get an error:
```
Execution failed for task ':gravatar:prepareToPublishToS3'.
> Either tagName or both branch name and sha1 command line arguments needs to be provided!
```

So I would have to disable that somehow on tagged builds?

wzieba: I haven't tested it, but would it be possible to apply prepare-to-s3 plugin on tagged builds only?

The check for tagged builds would have to be probably a part of initialization process, so very likely you'd have to put this in `settings.gradle.kts`, but maybe it's possible.

PR #590, opened 2025-02-26
Configure publishing to Maven Central

### LINKED ISSUE #583: Publish the SDK in Maven Central

AdamGrzybkowski:
Closes #583 

### Description

This PR migrates the publishing from the private s3 to Maven Central. 

Publishing to Maven Central via Publisher API requires uploading a [valid bundle](https://central.sonatype.org/publish/publish-portal-api/#uploading-a-deployment-bundle). This is not supported by default in the vanilla maven-publish plugin, so I used https://github.com/vanniktech/gradle-maven-publish-plugin as it's one of the alternatives [here](https://central.sonatype.org/publish/publish-portal-gradle/).

Initially, I was hoping to keep publishing to s3 (especially the non-tagged builds), but I found issues when using both plugins as they both configure the single variant for publishing creating a build error. The Vanniktech will add the option to publish SNAPSHOT versions to Maven Central in the next version, so I think this is a more straightforward solution rather than trying to combine the two plugins together. 

### Testing Steps

Green CI and verify there's a published version in Maven Central.


build.gradle.kts
```diff
@@ -46,9 +47,11 @@ apiValidation {
     ignoredProjects.addAll(listOf("demo-app", "uitestutils"))
 }
 
+val isTagBuild: Boolean = System.getenv("BUILDKITE_TAG")?.isNotEmpty() == true
+
 val sdkVersion = providers.exec {
     commandLine("git", "describe", "--tags")
-}.standardOutput.asText.get().trim()
+}.standardOutput.asText.get().trim() + if (isTagBuild) "" else "" // Add -SNAPSHOT suffix once 0.31.0 version is released
```
AdamGrzybkowski: As the comment says, publishing snapshots to Maven Central will be supported in 0.31.0 so this line will have to be modified once we can update. There's no rush at this point so we can wait for this.

gravatar-quickeditor/build.gradle.kts
```diff
@@ -132,16 +133,47 @@ dependencies {
...
         }
+
+        organization {
+            name.set("Gravatar.com")
+            url.set("https://www.gravatar.com/")
+        }
     }
+
+    signAllPublications()
 }
```
AdamGrzybkowski: I would like to extract this so that it can be reused across modules.

AdamGrzybkowski: This was done in a separate PR - https://github.com/Automattic/Gravatar-SDK-Android/pull/592

wzieba (commented):
Looks nice! Does the published artifact contain sources and documentation as well?

gravatar-quickeditor/build.gradle.kts
```diff
@@ -132,16 +133,47 @@ dependencies {
...
+        licenses {
+            license {
+                name.set("Mozilla Public License, Version 2.0")
+                url.set("http://www.mozilla.org/MPL/2.0/index.txt")
+            }
+        }
+
+        scm {
+            connection.set("scm:git:github.com:Automattic/Gravatar-SDK-Android.git")
+            developerConnection.set("scm:git:ssh://github.com:Automattic/Gravatar-SDK-Android.git")
```
wzieba: WDYT about not suggest using ssh - I think clone by https (especially of open source project) is rather default
```suggestion
            developerConnection.set("scm:git:https://github.com/Automattic/Gravatar-SDK-Android.git")
```

wzieba (approved):
Artifacts on Maven Central look good! One minor comment

AdamGrzybkowski:
0.31.0 is out. I was able to publish the SNAPSHOT (it's [here](https://central.sonatype.com/service/rest/repository/browse/maven-snapshots/com/gravatar/gravatar/)). I will merge this once the CI is 🟢

PR #592, opened 2025-02-27
Share Maven Publish logic between modules with Convention plugin

### LINKED ISSUE #591: Tooling: Introduce build-logic module with convention plugins
This will clean up the `build.gradle.kts` files and allow for easier logic sharing between the submodules.

AdamGrzybkowski:
Part of #591

### Description

This PR introduced `build-logic` with convention plugins to share build logic between the modules. In order to keep it small I only extracted the config for Maven Publishing.

### Testing Steps

Code review and CI should be enough


wzieba (approved):
Neat!

PR #593, opened 2025-02-27
Move more common logic from build.gradle.kts to convention plugins

### LINKED ISSUE #591: Tooling: Introduce build-logic module with convention plugins

AdamGrzybkowski:
Closes #591

### Description

This PR moves more logic from submodules' `build.gradle.kts`. I have introduced three convention plugins: `GravatarAndroidLibraryConventionPlugin`, `GravatarComposeConventionPlugin`, and `GravatarOpenApiGeneratorConventionPlugin`.

I've kept dependencies and some custom logic in each `build.gradle.kts`.

FYI: I haven't touched the `build.gradle.kts` of the DemoApp in this PR.

### Testing Steps

Code review, building the project should be enough. 

You can also run `./gradlew :gravatar:openApiGenerate` to make sure the generation the same way (no changes expected after running this command).

You can also check the `build/reports/compose` folder for the generated metric files.

### DIFF
--- a/build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
@@ -1,18 +1,22 @@
 import com.android.build.gradle.LibraryExtension
+import com.android.build.gradle.LibraryPlugin
 import com.gravatar.configureDetekt
 import com.gravatar.configureKotlinAndroid
+import io.gitlab.arturbosch.detekt.DetektPlugin
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.configure
+import org.jetbrains.kotlin.gradle.plugin.KotlinAndroidPluginWrapper
+import org.jlleitschuh.gradle.ktlint.KtlintPlugin
 
 class GravatarAndroidLibraryConventionPlugin : Plugin<Project> {
     override fun apply(target: Project) {
         with(target) {
-            apply(plugin = "com.android.library")
-            apply(plugin = "org.jetbrains.kotlin.android")
-            apply(plugin = "org.jlleitschuh.gradle.ktlint")
-            apply(plugin = "io.gitlab.arturbosch.detekt")
+            apply<LibraryPlugin>()
+            apply<KotlinAndroidPluginWrapper>()
+            apply<KtlintPlugin>()
+            apply<DetektPlugin>()
 
             extensions.configure<LibraryExtension> {
                 configureKotlinAndroid(this)
--- a/build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
@@ -1,0 +1,40 @@
+import com.android.build.gradle.LibraryExtension
+import com.gravatar.configureDetekt
+import com.gravatar.configureKotlinAndroid
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.apply
+import org.gradle.kotlin.dsl.configure
+
+class GravatarAndroidLibraryConventionPlugin : Plugin<Project> {
+    override fun apply(target: Project) {
+        with(target) {
+            apply(plugin = "com.android.library")
+            apply(plugin = "org.jetbrains.kotlin.android")
+            apply(plugin = "org.jlleitschuh.gradle.ktlint")
+            apply(plugin = "io.gitlab.arturbosch.detekt")
+
+            extensions.configure<LibraryExtension> {
+                configureKotlinAndroid(this)
+                defaultConfig.apply {
+                    targetSdk = 34
+                    consumerProguardFiles("consumer-rules.pro")
+                }
+                configureBuildTypes()
+            }
+            configureDetekt()
+        }
+    }
+
+    private fun LibraryExtension.configureBuildTypes() {
+        buildTypes {
+            getByName("release") {
+                isMinifyEnabled = false
+                proguardFiles(
+                    getDefaultProguardFile("proguard-android-optimize.txt"),
+                    "proguard-rules.pro",
+                )
+            }
+        }
+    }
+}
--- a/build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
@@ -10,6 +10,10 @@ import org.gradle.kotlin.dsl.configure
 import org.jetbrains.kotlin.gradle.plugin.KotlinAndroidPluginWrapper
 import org.jlleitschuh.gradle.ktlint.KtlintPlugin
 
+private const val TARGET_SDK = 34
+internal const val COMPILE_SDK = 34
+internal const val MIN_SDK = 21
+
 class GravatarAndroidLibraryConventionPlugin : Plugin<Project> {
     override fun apply(target: Project) {
         with(target) {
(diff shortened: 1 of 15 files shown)


build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    private fun LibraryExtension.configureBuildTypes() {
+        buildTypes {
+            getByName("release") {
+                isMinifyEnabled = false
+                proguardFiles(
+                    getDefaultProguardFile("proguard-android-optimize.txt"),
+                    "proguard-rules.pro",
+                )
```
wzieba, line 44: Are we sure that this SDK needs `proguard-rules.pro`?

build-logic/convention/src/main/kotlin/GravatarComposeConventionPlugin.kt
```diff
@@ -0,0 +1,61 @@
...
+    override fun apply(target: Project) {
+        with(target) {
+            apply(plugin = "io.github.takahirom.roborazzi")
+
+            extensions.configure<LibraryExtension> {
+                buildFeatures {
+                    compose = true
+                }
+                composeOptions {
+                    kotlinCompilerExtensionVersion = "1.5.15"
```
wzieba: Minor, but is it possible to extract this to the toml file?

build-logic/convention/src/main/kotlin/GravatarComposeConventionPlugin.kt
```diff
@@ -0,0 +1,61 @@
...
+                        freeCompilerArgs += listOf(
+                            "-P",
+                            "plugin:androidx.compose.compiler.plugins.kotlin:metricsDestination=" +
+                                    project.layout.buildDirectory.get().dir(composeReportsDir).asFile.absolutePath,
+                        )
+                        freeCompilerArgs += listOf(
+                            "-P",
+                            "plugin:androidx.compose.compiler.plugins.kotlin:reportsDestination=" +
+                                    project.layout.buildDirectory.get().dir(composeReportsDir).asFile.absolutePath,
+                        )
```
wzieba: Would such simplification work?

```kotlin
freeCompilerArgs += listOf(
    "-P", "plugin:androidx.compose.compiler.plugins.kotlin:stabilityConfigurationPath=${project.rootDir}/compose_compiler_config.conf",
    "-P", "plugin:androidx.compose.compiler.plugins.kotlin:metricsDestination=${project.layout.buildDirectory.get().dir(composeReportsDir).asFile.absolutePath}",
    "-P", "plugin:androidx.compose.compiler.plugins.kotlin:reportsDestination=${project.layout.buildDirectory.get().dir(composeReportsDir).asFile.absolutePath}"
)
```

wzieba: Or even

```kotlin
val prefix="plugin:androidx.compose.compiler.plugins.kotlin"

freeCompilerArgs += listOf(
    "-P", "$prefix:stabilityConfigurationPath=${project.rootDir}/compose_compiler_config.conf",
    "-P", "$prefix:metricsDestination=${project.layout.buildDirectory.get().dir(composeReportsDir).asFile.absolutePath}",
    "-P", "$prefix:reportsDestination=${project.layout.buildDirectory.get().dir(composeReportsDir).asFile.absolutePath}"
)
```

to highlight real options changes

build-logic/convention/src/main/kotlin/GravatarOpenApiGeneratorConventionPlugin.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.gradle.kotlin.dsl.apply
+import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.withType
+import org.openapitools.generator.gradle.plugin.extensions.OpenApiGeneratorGenerateExtension
+import org.openapitools.generator.gradle.plugin.tasks.GenerateTask
+
+class GravatarOpenApiGeneratorConventionPlugin : Plugin<Project> {
+    override fun apply(target: Project) {
+        with(target) {
+            apply(plugin = "org.openapi.generator")
```
wzieba: Minor, but is it possible to reference this plugin by type instead of name?

wzieba (commented):
Interesting, thanks for working on that! We've been discussing usage of convention plugins in other project, but haven't done much in this direction.

I left some questions and suggestions, wdyt about them?

build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
AdamGrzybkowski: At this point, the `proguard-rules.pro` are empty. We do have one rule in the consumer file for the SDK code. So... we could remove it, but does this setup negatively impact the project? If we need a rule, the setup is there, and very little is required to add it.

build-logic/convention/src/main/kotlin/GravatarComposeConventionPlugin.kt
AdamGrzybkowski: Looks good!

build-logic/convention/src/main/kotlin/GravatarOpenApiGeneratorConventionPlugin.kt
AdamGrzybkowski: Yes, TIL

build-logic/convention/src/main/kotlin/GravatarComposeConventionPlugin.kt
AdamGrzybkowski: It is. It doesn't work as nicely as in the `.gradle` files, but you can access it. I've pushed the changes.

build-logic/convention/src/main/kotlin/GravatarOpenApiGeneratorConventionPlugin.kt
wzieba: I see other place `GravatarAndroidLibraryConventionPlugin` that could have this suggestion applied as well

build-logic/convention/src/main/kotlin/GravatarComposeConventionPlugin.kt
wzieba: Looks good, thanks :)

build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
wzieba: > So... we could remove it, but does this setup negatively impact the project?

Let's leave this as it is - I only suggested this to reduce the configuration, but that's not a big deal 👍

build-logic/convention/src/main/kotlin/GravatarComposeConventionPlugin.kt
```diff
@@ -0,0 +1,60 @@
...
+                        // For Roborazzi
+                        isIncludeAndroidResources = true
+                        all {
+                            // -Pscreenshot to filter screenshot tests
+                            it.useJUnit {
+                                if (project.hasProperty("screenshot")) {
+                                    includeCategories("com.gravatar.uitestutils.ScreenshotTests")
+                                } else {
+                                    excludeCategories("com.gravatar.uitestutils.ScreenshotTests")
+                                }
```
wzieba, line 35: Out of curiosity and I haven't tested it but do we have to have both "include" and "exclude" mechanisms? Aren't things "excluded" by default?

```suggestion
                                if (project.hasProperty("screenshot")) {
                                    includeCategories("com.gravatar.uitestutils.ScreenshotTests")
                                }
```

wzieba (approved):
Thanks for applying improvements! 

As further work on convention plugins, I'd suggest applying a code formatter maybe for this module as well at some point.

build-logic/convention/src/main/kotlin/GravatarOpenApiGeneratorConventionPlugin.kt
AdamGrzybkowski: Right! All are applied by type now.

AdamGrzybkowski:
@wzieba 
>As further work on convention plugins, I'd suggest applying a code formatter maybe for this module as well at some point.

Good call! I have updated the CI pipeline as well.

PR #598, opened 2025-03-21
Use authenticated profile endpoint to fetch profile in the QE

### LINKED ISSUE #596: Use an authorized profile endpoint for the QE

AdamGrzybkowski:
Closes #596 

### Description

A new endpoint to get a profile that takes the auth token was added. This improves the flow when using the QE in the Jetpack app for a user who hasn't opened the Quick Editor before because it will create an identity for him. 

Instead of seeing an empty profile card on the initial load, the user will see his profile now.

### Testing Steps

Techincally, this can be tested with the demo app with the following steps:

1. Create a new account on WordPress.com and verify email
2. Obtaing the auth token 
3. Make sure the QE is not opened before that anywhere
4. Run the demo app
5. Enable the bearer token and paste it
Screenshot 2025-03-24 at 10 10 21

6. Tap on the "Update Avatar"
7. Conirm the profile card shows the profile (the username should be visible)

### DIFF
--- a/gravatar/src/main/java/com/gravatar/services/ProfileService.kt
+++ b/gravatar/src/main/java/com/gravatar/services/ProfileService.kt
@@ -178,4 +178,41 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
     ): GravatarResult<Boolean, ErrorType> = runCatchingRequest {
         checkAssociatedEmail(oauthToken, email)
     }
+
+    /**
+     * Retrieves the profile information for the authenticated user.
+     *
+     * @param oauthToken The OAuth token to use for authentication
+     * @return The profile information for the authenticated user
+     */
+    public suspend fun retrieveProfile(oauthToken: String): Profile = runThrowingExceptionRequest {
+        withContext(GravatarSdkDI.dispatcherIO) {
+            val service = GravatarSdkDI.getGravatarV3Service(okHttpClient, oauthToken)
+
+            val response = service.getProfile()
+            if (response.isSuccessful) {
+                response.body() ?: error("Response body is null")
+            } else {
+                // Log the response body for debugging purposes if the response is not successful
+                Logger.w(
+                    LOG_TAG,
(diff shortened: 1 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
```diff
@@ -178,4 +178,41 @@ public class ProfileService(private val okHttpClient: OkHttpClient? = null) {
...
         checkAssociatedEmail(oauthToken, email)
     }
+
+    /**
+     * Retrieves the profile information for the authenticated user.
+     *
+     * @param oauthToken The OAuth token to use for authentication
+     * @return The profile information for the authenticated user
+     */
+    public suspend fun retrieveProfile(oauthToken: String): Profile = runThrowingExceptionRequest {
```
AdamGrzybkowski: I'm not convinced about this name. We use `retrieve` in other places but the `retrieve(String)` is already used, so that won't work. I'm leaning towards sth like `retrieveWithToken` or `retrieveAuthenticated` to highlight the difference. Thoughts?

etoledom: I personally like to read `profile` in the name, something like `profileAuthenticated(withToken:)`, to know explicitly on name that this method will return a profile (and it's also easier to search all the get profile options via autocomplete).

For consistency, I think something like `retrieveAuthenticated(withToken:)` would be a good fit following the naming pattern already in use.

etoledom (approved):
Looks good!

gravatar/src/main/java/com/gravatar/services/ProfileService.kt
AdamGrzybkowski: Thanks for your input! I went with `retrieveAuthenticated`.

PR #602, opened 2025-04-07
Adam/remove retrofit

### LINKED ISSUE #600: Remove Retrofit from :gravatar module
The fewer dependencies, the better. We should be able to fairly simply replace Retrofit with Okhttp directly. 

This only covers the :gravatar module. :quick-editor should be handled separately.

AdamGrzybkowski:
Closes #600 

### Description

This PR removes the Retrofit dependency from the `:gravatar` module. In order to do that, I had to:

1. Update the OpenAPI generator config to use OkHttp instead of Retrofit.
2. Regenerate the code and apply some custom `.mustache` templates to best suit our codebase (this included some custom code for the network response as I didn't like the generated code that felt more like Java rather the Kotlin).
3. Adjust the public API of our SDK to internally use the newly generated code. 

This change should be invisible from the outside, no Retrofit classes were exposed (this made this change possible). 

I know the diff looks big, but:
1. A lot of new lines come from the `.mustache` templates that were mostly copied. 
2. Classes generated from step 2 added quite a bit as well. The whole `restapi` directory is generated.

### Testing Steps

It's best to use the QuickEditor and try to perform all the possible actions that require the network request. You can pay special attention to how errors are shown when loading the avatars (opening the QE) and uploading them. Those two have some extra error handling that either checks the response code or uses the message from the error body.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

etoledom (approved):
Tested:

- Token expired error
- OAuth login
- Avatars first load
- Avatar selection 
- Rating edition 
- Alt edition
- Avatar deletion 
- Avar upload
- Avatar upload retry
- Deleting avatar
- Deleting selected avatar
- Uploading avatar without previous selection

I couldn't test:
- Uploading an unsupported image

About the errors:

I've tested the no-connection error. All texts expect for the initial load had a generic error message

CleanShot 2025-04-07 at 18 11 21@2x

But the first load error had a more specific message:
CleanShot 2025-04-07 at 18 16 14@2x


If this is expected, or if it is the same than it was before this PR, then I'd say it's good to move forward ✅

AdamGrzybkowski:
Thanks for testing!

>I've tested the no-connection error. All texts expect for the initial load had a generic error message

Correct, only the initial avatar load and the avatar upload have more specific error messages.

PR #610, opened 2025-04-18
Properly init in-memory usage when restoring the state

### LINKED ISSUE #609: "Session expired" after re-opening the QuickEditor
**Describe the bug**
The Gravatar screen now says "Session expired" and tapping the Close button does nothing.

**To Reproduce**
Steps to reproduce the behavior in the Jetpack App:

1. Enable "Don't keep activities"
2. Tap to view my avatar
3. Tap "View Profile"
4. Return to the app
5. Tap to view my avatar again

Originally report [here](https://github.com/wordpress-mobile/WordPress-Android/pull/21808#issuecomment-2809238925)

AdamGrzybkowski:
Closes #609

### Description

When the "Don't keep activities" is ON, the Quick Editor state won't properly display the data inside. It's caused by a timing issue. The `QuickEditorContainer.getInstance().useInMemoryTokenStorage()` that is called in the `DisposableEffect` is invoked after the `NavController` starts the restored destination, so the ViewModel and its dependencies don't use the in-memory storage as they should. 

Moving the `DisposableEffect` to a Composable higher up in the hierarchy fixes this timing issue. 

| Before | After |
|---|---|
| state_before | state_after |

### Testing Steps

1. [Enable Developer Options](https://developer.android.com/studio/debug/dev-options) on your device
2. Launch the demo app
3. Move to the Avatar update tab
4. Paste in your WP token and select the checkbox next to it. You can also add it to `secrets.properties` file with name `demo-app.bearer.token`.
5. Tap the avatar to launch the QE
6. Tap on the `View Profile`
7. Go back with a system button or a gesture
8. Confirm the QE is restored and no errors are visible

### DIFF
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/AvatarUpdateTab.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/AvatarUpdateTab.kt
@@ -77,7 +77,7 @@ import kotlinx.coroutines.withContext
 fun AvatarUpdateTab(modifier: Modifier = Modifier) {
     var userEmail by remember { mutableStateOf(BuildConfig.DEMO_EMAIL) }
     var userToken by remember { mutableStateOf(BuildConfig.DEMO_BEARER_TOKEN) }
-    var useToken by remember { mutableStateOf(false) }
+    var useToken by rememberSaveable { mutableStateOf(false) }
     var tokenVisible by remember { mutableStateOf(false) }
     val context = LocalContext.current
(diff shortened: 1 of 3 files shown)


etoledom (approved):
I was able to reproduce the issue in trunk, and these changes fixes it 🎉 

I also saw another problem where the profile card wasn't showing when opening the QE with a given token, but this also seem to have been fixed with these changes.

This last issue I wasn't to reproduce it again in trunk, so I can't get a video now.

PR #612, opened 2025-04-23
Move the Upload Image button outside the border to match new designs

AdamGrzybkowski:
### Description

PART1/3

This PR moves the Upload Image button outside the border in the main picker page.

| Before | After |
|---|---|
| Screenshot 2025-04-23 at 13 58 13 | Screenshot 2025-04-23 at 13 56 43 |

### Testing Steps

There are screenshot tests, so you can check them but you can also verify in the demo app:

1. Launch the QE
2. Log in
3. Confirm the button is outside the border
4. Remove all photos (or log in with a different email)
5. Confirm the empty state looks correct as well (see screenshots)
6. Repeat with the `vertical` scrolling


etoledom (approved):
Looking good! 🎉 

One small unrelated detail I noticed while testing:


https://github.com/user-attachments/assets/bb096e6f-9afe-4ce1-b6b0-1601c2be7825

On the vertical layout, there doesn't seem to be an animation after deleting an avatar

AdamGrzybkowski:
>On the vertical layout, there doesn't seem to be an animation after deleting an avatar

You're correct! I will check it later separately. Thanks

PR #615, opened 2025-04-24
Make the top views of the QE be reused by the scopes

AdamGrzybkowski:
### Description

This PR adds `QuickEditor` composable component that will host the shared UI (top bar, email, and the profile card) and different editor scopes as `content` that will be added in the next PRs.

Some of the logic from the `AvatarPickerViewModel` was now moved to the new `QuickEditorViewModel` as the profile fetching will only happen there (plus some extra logic to force refresh the avatar).

Additionally, the alert banner that's shown the no avatar is selected was moved below the profile card, so that it stays in the avatar picker content space. Apart from this, there should be no UI change.

### Testing Steps

As mentioned above, the UI should be unchanged (except for the alert banner position) but there are a few things worth testing:

**Avatar updates in the profile card**
Make sure the avatar is properly updated after each of the below:
1. Selecting a different avatar
2. Deleting the selected avatar

**Refreshing after network failure**
1. Launch the demo app
2. Disable the internet connection
3. Open the QuickEditor (make sure you are logged in)
4. You should see an error with a button to try again
5. Enable the internet connection
6. Tap the button to try again
7. Confirm the Profile card and the rest was refreshed and rendered


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/QuickEditor.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.gravatar.quickeditor.ui.components.EmailLabel
+import com.gravatar.quickeditor.ui.components.ProfileCard
+import com.gravatar.quickeditor.ui.components.QEPageDefault
+import com.gravatar.ui.GravatarTheme
+
+@Composable
+internal fun QuickEditor(
+    gravatarQuickEditorParams: GravatarQuickEditorParams,
+    handleExpiredSession: Boolean,
+    onAvatarSelected: () -> Unit,
```
etoledom, line 23: One thing we decided to do on iOS is to have a general `update()` callback (we call it `updateHandler()`, and it has a parameter with the information of which kind of update it is.

https://github.com/Automattic/Gravatar-SDK-iOS/blob/57d6110c3671f0da74480680e6fcb5012637be5e/Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift#L59

https://github.com/Automattic/Gravatar-SDK-iOS/blob/57d6110c3671f0da74480680e6fcb5012637be5e/Sources/GravatarUI/SwiftUI/ProfileEditor/QuickEditor.swift#L23-L35

Maybe is not needed yet but it will possible become handy when we start to implement the About Info editor.

etoledom (approved):
Looks good 👍 
All manual testing steps passed 🟢 

One comment on code, though not a blocker.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/QuickEditor.kt
AdamGrzybkowski: Thanks! The API design is yet to be decided on, so I will definitely look at iOS.

PR #619, opened 2025-04-28
Add Profile card to the demo app, to show profile changes

AdamGrzybkowski:
### Description

A small UI update in the demo App to better show profile updates in the future once we have the "About" editor.

| Before | After |
|---|---|
| Screenshot 2025-04-28 at 12 20 35 | Screenshot 2025-04-28 at 12 19 50 |

### Testing Steps

1. Open the Demo app
2. Go to Avatar Update tab
3. Type your email
4. Confirm the profile card is updated
5. Launch the QE
6. Update the avatar
7. Confirm the profile card is updated

### DIFF
--- a/gravatar/src/main/java/com/gravatar/restapi/apis/AvatarsApi.kt
+++ b/gravatar/src/main/java/com/gravatar/restapi/apis/AvatarsApi.kt
@@ -28,9 +28,7 @@ internal class AvatarsApi(
 ) : ApiClient(basePath, client) {
     internal companion object {
         @JvmStatic
-        val defaultBasePath: String by lazy {
-            System.getProperties().getProperty(ApiClient.BASE_URL_KEY, "https://api.gravatar.com/v3")
-        }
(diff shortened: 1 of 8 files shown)


gravatar/src/main/java/com/gravatar/restapi/apis/AvatarsApi.kt
```diff
@@ -28,9 +28,7 @@ internal class AvatarsApi(
 ) : ApiClient(basePath, client) {
     internal companion object {
         @JvmStatic
-        val defaultBasePath: String by lazy {
-            System.getProperties().getProperty(ApiClient.BASE_URL_KEY, "https://api.gravatar.com/v3")
```
AdamGrzybkowski, line 32: As mentioned in the git commit message. This was breaking the Compose preview feature and because we are not using it I simply removed the `System.getProperties` from the `.mustache`.

etoledom (approved):
Looking great 🎉 

Avatar updates from the QE and card updates with different emails are working as expected.

PR #606, opened 2025-04-09
Clean up the README.md file to focus on 3rd party devs

AdamGrzybkowski:
Closes #571

### Description

This PR cleans up the main README.md file to focus on providing the most important information for 3rd party app developers. Most of the file was about technical information that's more valuable for us (maintaners) and that shouldn't be the case. All these technical details were moved to a separate .md file. 

Additionally, I've added some extra sections (Features/Installation/Documentation) to align our docs with the iOS repo.

@wzieba Since you were the one to point out those things, would you mind reviewing the PR? 🙏 

### Testing Steps

Read the docs.

PR #319, opened 2024-09-17
QuickEditor: New OpenAPI spec

AdamGrzybkowski:
Closes #290

### Description

This PR:
1. Updates the OpenAPI spec
2. Regenerates the code
3. Updates the QE code to use new/updated endpoints. This includes:
- To fetch avatars we no longer need the `/identities` endpoint. Simplified to one only.
- When uploading the image, we do get the `Avatar` object in the response, meaning we can skip the extra request to refresh the avatar list. 

### Testing Steps

Smoke test the QE, especially avatar upload.

PR #620, opened 2025-04-29
AboutEditor page within the QE with input fields

AdamGrzybkowski:
### Description

This PR adds the new `AboutEditor` component that displays all the about fields: Display name, About me, Location, Pronouns, Pronouncations, Job title, Company.

`AboutEditorViewModel` fetches the profile and loads it, sets the values for those fields, and allows each field to be modified. However, there is no option to save the changes yet. 

**Note: The initial height, animation of the bottom sheet are not part of this PR. It will be handled separately. For now, the bottom sheet takes as much space as it needs, and because we show all the fields, it will take full screen height.**

I also skipped showing an error here after failing to fetch the profile, as the PR grew quite big already. 

Screenshot 2025-04-29 at 13 33 11

### Testing Steps

1. Launch the demo app
2. Select the `ABOUT` scope
3. Tap `Open QE`
4. Confirm you can see all the fields and that the UI matches Figma
5. Try to type in some of the fields
6. Confirm the `About me` is the only one take can expand vertically up to 4 lines.
7. To better test the scrolling you can modify the font or display scaling. In the emulator you can do it easily with this menu:

Screenshot 2025-04-29 at 13 34 45


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/QuickEditorScope.kt
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * The scope of the Quick Editor.
+ *
+ * WARNING: Not a final API. This API will change in the future.
+ */
+@Experimental
+public enum class QuickEditorScope {
+    AVATAR,
+    ABOUT,
+}
```
etoledom, line 14: Should we sync these names across platforms?

on iOS we have:
- `avatarPicker`
- `aboutEditor`

I don't have personal preferences though

AdamGrzybkowski: Yes! This is a temporary solution to be able to launch it from the demo app. That's why I marked it as experimental.

I will tackle the public API separately.

etoledom (approved):
Looking good!
Behaves as described

~Just one question in code review, but not blocker 👍~ Already solved

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/QuickEditorScope.kt
etoledom: Cool 👍

Milanhe92:
#329 #604 #606 #319 #605

PR #625, opened 2025-05-06
Alert dialog when closing without saving changes

AdamGrzybkowski:
### Description

This PR adds an alert dialog that is shown when the user tries to close the QuickEditor with unsaved changes in the About editor. 

The copy is not final yet, but because we aren't merging to `trunk`, the strings won't be uploaded to GlotPress.

https://github.com/user-attachments/assets/a0e09af3-df0d-410f-954e-6a0906e1c861

### Testing Steps

1. Launch the QE with About scope
2. Tap Done
3. Confirm the QE was closed
4. Open the QE again
5. Change some fields 
6. Tap Done
7. Confirm you see the alert dialog
8. Tap Exit
9. Confirm the QE was closed
10. Open the QE again
5. Change some fields 
6. Tap Done
7. Confirm you see the alert dialog
8. Tap keep editing 
9. Save changes
10. Tap Done
11. Confirm the QE was closed

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/QuickEditor.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/QuickEditor.kt
@@ -11,6 +11,9 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.viewmodel.compose.viewModel
 import com.gravatar.quickeditor.ui.abouteditor.AboutEditor
+import com.gravatar.quickeditor.ui.abouteditor.AboutEditorEvent
+import com.gravatar.quickeditor.ui.abouteditor.AboutEditorViewModel
+import com.gravatar.quickeditor.ui.abouteditor.AboutEditorViewModelFactory
 import com.gravatar.quickeditor.ui.avatarpicker.AvatarPicker
 import com.gravatar.quickeditor.ui.components.EmailLabel
 import com.gravatar.quickeditor.ui.components.ProfileCard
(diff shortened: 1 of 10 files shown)


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/QuickEditor.kt
```diff
@@ -30,10 +33,19 @@ internal fun QuickEditor(
     ),
 ) {
     val uiState by viewModel.uiState.collectAsState()
+    val aboutEditorViewModel: AboutEditorViewModel = viewModel(
+        factory = AboutEditorViewModelFactory(gravatarQuickEditorParams),
+    )
```
AdamGrzybkowski, line 38: I had to move up the creation of the `AboutEditorViewModel` to invoke it on the `done` button tap. The button is in a parent component, so we can't control it from the About Editor.

etoledom (commented):
Testing steps behave as expected 🎉 

I did notice that swiping down the modal will dismiss it loosing the changes. This might be the preferred way of closing the QE for some users.

Is there a way to prevent swiping down to close while there are unsaved changes (and maybe showing the alert right away?) Or prevent it completely on the About Editor?

https://github.com/user-attachments/assets/8a82f6de-c758-4301-84c5-dd444948267d

AdamGrzybkowski:
>I did notice that swiping down the modal will dismiss it loosing the changes. This might be the preferred way of closing the QE for some users.

>Is there a way to prevent swiping down to close while there are unsaved changes (and maybe showing the alert right away?) Or prevent it completely on the About Editor?


Thanks for catching this! So far I think this would be quite hard to achieve. From what I know, we can detect the target detent, but once we know that, it's a bit tricky to pause it and show the alert dialog. 

I'm not sure I like the idea of disabling the swipe-to-dismiss behavior - I would rather have it aligned with the avatar picker. 


What is the behavior on iOS?

etoledom (approved):
We've decided internally to take more time to decide about what to do with the other closing actions. It might require a different feature, so we move forward with this one 👍

PR #626, opened 2025-05-06
Use rememberSaveable for userEmail to preserve the value during config changes

AdamGrzybkowski:
### Description

As reported: 

>In short, I'm logged in with an email, then I try to log in with a different email, but when I finish the OAuth process, the app goes back to the previous email, and shows that previous account instead of the new one.

This happens with the `Don't keep activities` setting ON.

The fix is to use `rememberSaveable` instead of `remember`. The former preserves the value between config changes.

### Testing Steps

1. Enable "Don't keep activities"
2. Launch the QE and login with account nr 1
3. Close the QE
4. Type in email for account nr 2
5. Open the QE
6. Make sure to use correct email in the oauth screen
7. Finish oauth
8. Confirm you are logged in as account nr 2


etoledom (approved):
I can no longer reproduce the issue previously mentioned 🎉 
Thanks for the fix! 🙏

PR #628, opened 2025-05-09
Public API with new Quick Editor scopes

AdamGrzybkowski:
### Description

This PR introduced the new public API that will allow for Quick Editor's scope configuration. 

Here are our two rules that we want to follow:

1. The scopes are stated explicitly rather than as a combination of modes.
2. Adding a new scope in the future should NOT be a breaking change.

The above means we can't use Enums or Sealed classes for `QuickEditorScope` nor for the `QuickEditorPage`.
We must ensure that whenever we add a new scope in the future, it won't be a breaking change. 

**Scope**

Following the iOS implementation, we are introducing `QuickEditorScopeOption` class that will hold the config for each scope. The config can be configured via publicly available configurator classes (eg. `AvatarPickerConfiguration`) and internally kept as a sealed class for better type safety and ease of work. 

`QuickEditorScopeOption` creation will be limited to three static functions (with the constructor being private). Each function will take a corresponding configurator class. You can see an example of that in the `AvatarUpdateTab`.

```kotlin
QuickEditorScopeOption.avatarPicker(
    config = AvatarPickerConfiguration(
       contentLayout = AvatarPickerContentLayout.Horizontal,
    ),
)
```

This solution makes it easy to extend in the future, limits how it can be created for third parties, making it easy to follow with just the IDE suggestions. It's also compatible with our requirements, and adding new scopes won't break the compilation in the future. 

**Enum replacement**

For the direct backwards compatible enum replacement ,I've decided to use Kotlin `value classes`. 

```kotlin
public value class QuickEditorPage private constructor(
    public val value: String,
) {
    public companion object {
        public val AvatarPicker: QuickEditorPage = QuickEditorPage("avatar_picker")
        public val AboutEditor: QuickEditorPage = QuickEditorPage("about_editor")
    }
}
```

As you can see, I set the constructor to `private` so the only option to create `QuickEditorPage` is via its companion object helper properties. This way, only those specific types can be created. We could even extend this functionality and add some helper functions to list all available types, etc.

The above was used for `QuickEditorPage`, and `AboutInputField`.

**Misc**

The `avatarPickerContentLayout` now belongs to the scope config, but because we can't introduce breaking changes, we could only deprecate it in `GravatarQuickEditorParams`. We will keep it there and only use it if the scope config matches the default value. 

### Testing Steps

You can test the Quick Editor using the DemoApp. It is now using the `QuickEditorScope` and `QuickEditorPage`. The only thing that is not yet implemented is the `AboutInputField` - you can't configure it in the demo app yet. I will add it separately. 

The important part of this PR is to review the API rather than QE functionality.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

etoledom:
This looks very similar to what we have on iOS, but maybe a bit in reverse.

So, from what I understand:

```kotlin
GravatarQuickEditorBottomSheet(
                gravatarQuickEditorParams = GravatarQuickEditorParams {
                    email = Email(userEmail)
                    uiMode = pickerUiMode
                    scopeConfig = ScopeConfig { // --> The configuration is initiated with a scope.
                        scope = editorScope // --> Scope goes here
                        initialPage = editorInitialPage // --> An initial page if needed for mixed scopes
                        avatarPickerContentLayout = pickerContentLayout // -> Layout for the avatar picker scope
                    }
                },
               // ( ... )
            )
```

What we have in iOS is that the scope holds a configuration, and every scope can only be configured with what that particular scope needs:

```swift
public struct QuickEditorScopeOption {
    public static func avatarPicker(
        _ config: AvatarPickerConfiguration = .horizontalInstrinsicHeight
    ) -> Self 

    public static func aboutEditor(
        _ config: AboutEditorConfiguration = .init()
    ) -> Self 

    public static func avatarPickerAndAboutInfoEditor(
        _ avatarPickerAndAboutEditorConfig: AvatarPickerAndAboutEditorConfiguration = .init()
    ) -> Self 
}
```

So the usage can look like this:

```swift
// Selecting a scope
let scope = QuickEditorScopeOption.aboutEditor(.init(fields: .all))

// Or directly on the Quick Editor view:
view
    .gravatarQuickEditorSheet(
        isPresented: $isPresentingPicker,
        email: email,
        authToken: token,
 // ---> Here the scope is set. And the scope option will require it's own configuration with what it needs.
        scopeOption: .aboutEditor(AboutEditorConfiguration(fields: .all)),
        customImageEditor: customImageEditor(),
        updateHandler: { updateType in
           // ( ... )
        },
        onDismiss: {
            // ( ... )
        }
    )
```

We also avoid using enums publicly to avoid future breaking changes, and we use the same strategy for this than on this PR.

I do like the concept of a `Page` (which we don't have yet). I think we can use it on iOS also. Internally I named the presenting context as a `Mode`, but I do like it more as a presenting `Page`. And we still need to work on the public API to choose this initial page. (cc @pinarol )

In short, both public APIs look similar. I don't think they have to be the same. I shared iOS in case it gives you a new idea and you think it could be an improvement, but I'm not requesting changes 👍 

Let me know what you think! But I'm happy to 🚢  as it is!

AdamGrzybkowski:
Thanks @etoledom, I like what you did on iOS. I will adjust the code and update the description.

AdamGrzybkowski:
@etoledom This is now ready for another review. I've adjusted the code to match iOS (including some names).

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/QuickEditorPage.kt
```diff
@@ -0,0 +1,31 @@
...
+import kotlinx.parcelize.Parcelize
+
+/**
+ * The page of the Quick Editor. The page corresponds to the part of your Gravatar profile that will be shown.
+ *
+ * @property value The value of the page.
+ */
+@Parcelize
+@JvmInline
+public value class QuickEditorPage private constructor(
```
AdamGrzybkowski: Should we keep it as a general page that could be extended by different ones in the future, or should we keep it tightly coupled with `AvatarPickerAndAboutEditorConfiguration` for better type safety?

AdamGrzybkowski: I think that not being able to pass the wrong page to the `AvatarPickerAndAboutEditorConfiguration` would be great.

etoledom: > I think that not being able to pass the wrong page to the AvatarPickerAndAboutEditorConfiguration would be great.

I agree with this.
The whole idea is that the config is tight to the possibilities of its scope.

Since we don't know what we will be developing in the SDK in the future, maybe it's better to do this specific, to avoid future deprecations. So: 

> or should we keep it tightly coupled with AvatarPickerAndAboutEditorConfiguration for better type safety?

I'd say probably this is a good option.

What do you think?

etoledom (approved):
Looking great!

Smoke tests works as expected 👍 

Just one small nit on the demo app side:

When the scope is set to About, the layout option has no effect, right?



On iOS we try to hide the options that are not set on the current selected scope. It will probably become more evident with the about fields selector option.

AdamGrzybkowski:
>When the scope is set to About, the layout option has no effect, right?

Correct - I haven't updated this part. I will change it.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/QuickEditorPage.kt
AdamGrzybkowski: Yes! I've added `Page` to the `AvatarPickerAndAboutEditorConfiguration` to keep it close. The idea is that each config would have its own `Page` if needed. Let me know what you think.

I've kept the `QuickEditorPage` as an internal type.

AdamGrzybkowski:
>On iOS we try to hide the options that are not set on the current selected scope. It will probably become more evident with the about fields selector option.

@etoledom Only config that is valid for a given scope should now be visible.

etoledom (approved):
Nice! Looks pretty good to me 🎉 

Let's follow this same setup for the Pages on iOS
And the demo app improvements, smooth 👍 👍

PR #629, opened 2025-05-12
Introduce UpdateHandler with QuickEditorUpdateType to handle different payloads in updates

AdamGrzybkowski:
### Description

Second PR with new public API. This one introduced an `UpdateHandler` - a callback that will replace the `onAvatarSelected` callback. 

```kotlin
typealias UpdateHandler = (QuickEditorUpdateType) -> Unit
```

`QuickEditorUpdateType` is an interface, and each new result should extend from it. I've used an interface to follow the same requirement of not breaking the compilation when the new scope is added in the future. 

### Testing Steps

1. Launch the demo app
2. Open the AvatarPicker
3. Update the avatar 
4. Confirm the avatar was updated in the demo app
5. Open AboutEditor
6. Update the display name
7. Confirm the profile card data was updated in the demo app

### DIFF
--- a/demo-app/src/main/java/com/gravatar/demoapp/ui/AvatarUpdateTab.kt
+++ b/demo-app/src/main/java/com/gravatar/demoapp/ui/AvatarUpdateTab.kt
@@ -26,8 +26,10 @@ import androidx.compose.material3.Text
 import androidx.compose.material3.TextField
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.neverEqualPolicy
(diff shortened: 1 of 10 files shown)


etoledom (commented):
Looks good! 
Manual tests worked as expected.

One question:

There's the idea to give back an instance of Profile on the about update callback.
Would that be possible using the interface strategy?

I'd guess the concrete type can hold any information we want. Just to be sure that if we decide to do this, it won't require big changes 👍

AdamGrzybkowski:
>There's the idea to give back an instance of Profile on the about update callback.
Would that be possible using the interface strategy?

The Profile is already there in the `AboutEditorResult`. Or are you asking about something different?

etoledom:
> 
> The Profile is already there in the `AboutEditorResult`. Or are you asking about something different?

```kotlin
public class AboutEditorResult(
    public val profile: Profile,
) 
```

Right! My bad 🙏 

All good to 🚢

etoledom (approved):

PR #630, opened 2025-05-15
Show Unsaved Changes alert dialog on modal dismissal actions (back button, tap outside, drag down).

AdamGrzybkowski:
### Description

This PR extends the number of triggers for the Unsaved Changes alert dialog. It will now be shown not only on `Done` button tap but also on any actions that trigger Modal dismissal. 

In order to do that, I had to update the Compose Unstyled library to have access to the `confirmDetentChange()` callback that was added and does exactly what we need.

Additionally, I had to:
- Remove the custom click handled on the `Scrim`
- Remove the custom `BackHandler`
Those two are now handled by the Modal sheet directly and hence are now routed via `confirmDetentChange` callback that we use. 

On top of that I had to change a bit where we create the `ModalBottomSheetState` as we need it in a certain place for the `confirmDetentChange` callback. 

Now to the actual implementation part: 

I ended up on a Plan B solution - that is to pass down a flag when the user tried to dismiss that dialog. When such a flag is received the `QuickEditorViewModel` will check the page and delegate the unsaved changes check.

Next to the flag we also need a new callback `onDismissIgnored()` to make sure we clear that flag when the user decide to "Keep editing" so that it will be possible to send it again later.

https://github.com/user-attachments/assets/cf998a01-fd0d-4071-bd02-d920c914fd5b

 

### Testing Steps

1. Launch the QE with the About scope or (AvatarAndAbout)
2. Try to close the QE using all the mentioned actions
3. Confirm it was closed fine
4. Open QE again
5. Update the profile (without saving)
6. Try to close the QE using all the mentioned actions
7. Tap `Keep editting`
8. Confirm you're back in QE, and the alert reappears every time
9. Try to dismiss it again
10. Tap `Exit`
11. Confirm the QE was closed
12. Run some basic tests for the Avatar scope as well to make sure it closes as before


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/AboutEditorViewModel.kt
```diff
@@ -40,23 +40,34 @@ internal class AboutEditorViewModel(
             is AboutEditorEvent.OnAboutFieldUpdated -> updateAboutField(aboutEditorEvent.aboutField)
             AboutEditorEvent.OnSaveClicked -> saveProfile()
             AboutEditorEvent.OnDoneClicked -> checkForUnsavedChanges()
-            AboutEditorEvent.OnDiscardDialogDismissed,
-            AboutEditorEvent.OnDiscardConfirmed,
-            -> dismissDiscardChangesDialog()
+            AboutEditorEvent.OnUnsavedChangesDialogDismissed -> dismissUnsavedChangesDialog()
+            AboutEditorEvent.OnUnsavedChangesDialogDiscarded -> discardUnsavedChangesDialog()
```
etoledom: Nit: `Dismiss` and `Discard` could be confusing to determine the action of each 🤔 

Both `DialogDismissed` - `DialogDiscarded`,
and `dismissUnsavedChanges` - `discardUnsavedChanges` pairs could mean the same to me.

I'd guess that `Dismiss` will only dismiss the alert and `Discard` will discard changes, as I know what's the idea of the alert and the feature.

On iOS we have `keep editing` and `discard changes` (though the internal implementaion is quite different).

But maybe `OnUnsavedChangesDialogDismissed` - `OnUnsavedChangesDialogChangesDiscarded`
and `dismissDialogUnsavedChanges` - `discardUnsavedChangesDialog`

As the dismiss is about the alert and the discard is about the changes 🤔

etoledom (commented):
It's working great overall!

- Showing the alert on About Editor from all the dismissal actions possible is working smoothly.
- Avatar Picker dismissal continues to work as expected.

🍅 When testing on the Avatar&About scope, I was able to dismiss the Quick Editor with unsaved changes, when dismissing from the Avatar Editor page:


CleanShot 2025-05-16 at 09 56 50

We work this out on iOS by being the QuickEditor (parent) who asks the models if there is any unsaved changes, and its also the one who presents the alert:

CleanShot 2025-05-16 at 10 00 51

AdamGrzybkowski:
>🍅 When testing on the Avatar&About scope, I was able to dismiss the Quick Editor with unsaved changes, when dismissing from the Avatar Editor page:


That's true. It was like that with the `Done` button as well so I kept the behavior for now. 

>We work this out on iOS by being the QuickEditor (parent) who asks the models if there is any unsaved changes, and its also the one who presents the alert:

I'm worried that showing the same alert while in the Avatar Picker could be confusing. It feels like it needs a better copy to say where the unsaved changes are with a quick action to navigate there.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/AboutEditorViewModel.kt
AdamGrzybkowski: Would you say that `OnUnsavedChangesExitClick` and `OnUnsavedChangesKeepEdittingClick` would be less confusing?

etoledom: > Would you say that OnUnsavedChangesExitClick and OnUnsavedChangesKeepEdittingClick would be less confusing?

That's perfectly clear! 🎉

etoledom:
> I'm worried that showing the same alert while in the Avatar Picker could be confusing. It feels like it needs a better copy to say where the unsaved changes are with a quick action to navigate there.

This is true, we can have a wider conversation about this.

I'm still hoping to have some visual cues for unsaved changes somewhere. But we agreed that this is outside of scope for now 🤔

AdamGrzybkowski:
>This is true, we can have a wider conversation about this.

I would keep the PR as is and extend the alert to the AvatarPicker page separately to align with iOS.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/AboutEditorViewModel.kt
AdamGrzybkowski: Updated!

etoledom (approved):
> I would keep the PR as is and extend the alert to the AvatarPicker page separately to align with iOS.

I agree! Let's do that.

Thank you for the update on the names 🙏

PR #634, opened 2025-05-19
About Editor - error handling

AdamGrzybkowski:
### Description

In this PR I'm adding a `CtaSection` component when an error is present in the state after load profile failure. 

Commint https://github.com/Automattic/Gravatar-SDK-Android/commit/b52179ffca458dcd0a818a7bb889dac127507cdc - Updates the `CtaSection` component a bit to make the internal border optionable due to a different view structure in the AboutEditor. 
Commit https://github.com/Automattic/Gravatar-SDK-Android/commit/3928bfafc3dd20fa0addbad35beb1fd555543ede - Adds the component to the AboutEditor and sets the error in the AboutEditorViewModel's state.
Commit https://github.com/Automattic/Gravatar-SDK-Android/commit/06d1d2d559ac955ea14ea53a1cce0937a3453dc0 - Adds the `onRefresh` callback so we can update the shared Profile card when the user triggers the refresh action via CTA button. The same pattern was used with the AvatarPicker.

| Unknown Error | No internet | Invalid token (only when token is provided) | onRefresh in action |
|---|---|---|---|
| Screenshot 2025-05-19 at 11 16 22 | Screenshot 2025-05-19 at 11 17 19 | Screenshot 2025-05-19 at 12 03 45 | onrefresh |

### Testing Steps

No internet error is the easiest - you can simply disable the internet and launch the QE.
Unknown requires other errors that don't have special handling, 500 from the server will do the job - you likely need to mock the response via the tool you use.
Invalid token - you can just type in a random string and select the token check box in the demo app.
OnRefresh callback - Do the No Internet test - Enable the internet - Tap Try again. Confirm the Profile card was shown and loaded.


etoledom:
I'm not sure if this was documented:

We do a special error handling on iOS when there's a token expired error while saving changes to the profile (About or Avatar)

CleanShot 2025-05-20 at 09 11 14


This was implemented on iOS since the Avatar Picker. 

It gives the user a recovery option within the editor.
It's easy to test using https://wordpress.com/me/security/connected-applications

etoledom (approved):
Looking good!
Tested all the scenarios I could imagine.

I noticed one difference between dismissing with the Done button and the Close button (from the error view):

With the close button, there is no dismiss animation:

CleanShot 2025-05-20 at 13 50 18

If we can do something about this, it would be great to do so I think.
Otherwise let's 🚢

PR #635, opened 2025-05-20
QuickEditor height adjustments for the new Scopes

AdamGrzybkowski:
### Description

This PR adjusts various aspect of the QE height in the new scopes:
- https://github.com/Automattic/Gravatar-SDK-Android/commit/e4c227cdb26cf76f885509cffdd50c9a6b374fc2 Limits the total height to make sure QE won't be drawn behind the status bar.
- https://github.com/Automattic/Gravatar-SDK-Android/commit/ddd5f6b9d7f7d612503896420f9190ca230627ef Add the `animateContentHeight` modifier to animated About editor height changes (this applies to any content change but in general when switching from loading to loaded or error state).
- https://github.com/Automattic/Gravatar-SDK-Android/commit/dacb12789584e8f3598306803f78ce97eb96bdd4 Updated the possible Detens and the initial detent based on scope (and other factors) for the QE.

| Max height | Content animation | Initial peek detent for About editor |
|---|---|---|
| Screenshot 2025-05-20 at 12 12 34 | animate_height | Screenshot 2025-05-20 at 12 17 00 |

Note: I've noticed a weird behavior in the AvatarAndAbout scope. With scrolling set to horizontal and the initial page to Avatar picker, the sheet reverts to Peek detent instead of FullyExpanded. I have a feeling this is rather related to the AvatarPicker UI hierarchy so I would like to work on it separately (GRA-103).

### Testing Steps

1. Max height is easy - just launch the About editor with all the fields and expand it fully. Same for the avatar picker in the vertical scrolling mode (with enough images).
2. The content animation is a bit tricky now that we have the peek detent as initial, but you can play around with it to see.
3. Sheet detents - Launch the About editor and confirm the sheet is not fully expanded. Rotate the phone to landscape and repeat - this QE should be fully extended as it doesn't make sense to use peek detent when we have very little screen height to work with.

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/AboutEditor.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/AboutEditor.kt
@@ -1,5 +1,6 @@
 package com.gravatar.quickeditor.ui.abouteditor
 
+import androidx.compose.animation.animateContentSize
 import androidx.compose.foundation.border
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
@@ -105,7 +106,10 @@ internal fun AboutEditor(
     }
 
     Surface {
-        Box(modifier = Modifier.wrapContentSize()) {
+        Box(modifier = Modifier
+            .wrapContentSize()
+            .animateContentSize()
+        ) {
             AboutEditor(
                 uiState = uiState,
                 onEvent = { event ->
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/AboutEditor.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/AboutEditor.kt
@@ -106,9 +106,10 @@ internal fun AboutEditor(
     }
 
     Surface {
-        Box(modifier = Modifier
-            .wrapContentSize()
-            .animateContentSize()
+        Box(
+            modifier = Modifier
+                .wrapContentSize()
+                .animateContentSize(),
         ) {
             AboutEditor(
                 uiState = uiState,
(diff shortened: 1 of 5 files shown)


etoledom:
I see one interesting big difference between platforms:

On iOS we have the Save button pinned at the bottom for any size of the sheet (either medium, expanded, or intrinsic content height)

I see on Android that, instead, the Save button will move out of the screen when the sheet is small:

CleanShot 2025-05-21 at 09 22 47

etoledom:
> Note: I've noticed a weird behavior in the AvatarAndAbout scope.

Is this what you mean?

I see that when opening the Avatar Picker on the Avatar & About scope, with horizontal layout, the initial height of the sheet is smaller than the content and some part of the bottom will get hidden under the screen.

It works good on Avatar Picker scope.

CleanShot 2025-05-21 at 09 26 39

etoledom:
For this reason also (Sticky save button), we decided to remove the save button from landscape mode with keyboard

https://github.com/Automattic/Gravatar-SDK-iOS/pull/756#issuecomment-2895381922

pinarol:
I remember like we wanted to always display the Save button. (The landscape + keyboard open mode is an exceptional case we discovered later though). But the idea to me was that the input fields inside the borders would be scrollable not the button. We can get some early design feedback once the feature is in a testable/mature state I think.

etoledom:
> We can get some early design feedback once the feature is in a testable/mature state I think.

@pinarol - Do you mean to keep it this way on Android for now?

AdamGrzybkowski:
>I see on Android that, instead, the Save button will move out of the screen when the sheet is small:


The button is sticky. The difference likely comes from a different platform. Having the sheet in the half-expanded state (peak detent) means it has its full height, and some parts are simply outside the screen area. Once the sheet is fully expanded (think more of sliding into the screen, rather than expanding in height), the button will be pinned at the bottom (when showing the keyboard, scaling up the UI, etc.).

Screenshot 2025-05-26 at 09 12 47

AdamGrzybkowski:
>Is this what you mean?

>I see that when opening the Avatar Picker on the Avatar & About scope, with horizontal layout, the initial height of the sheet >is smaller than the content and some part of the bottom will get hidden under the screen.

>It works good on Avatar Picker scope.

Yes, it's exactly this. I have this logged as GRA-103.

etoledom:
> The button is sticky. The difference likely comes from a different platform. Having the sheet in the half-expanded state (peak detent) means it has its full height, and some parts are simply outside the screen area.

This is probably another good reason to re-think the way we save changes.
Having to expand always to be able to save sounds to me like extra friction.

But probably not worth blocking this PR. As @pinarol mention we can get feedback about this during testing 👍

etoledom (approved):
Looking great 🎉

PR #636, opened 2025-05-20
Handle back press in the QE manually

AdamGrzybkowski:
### Description

This PR fixes a bug that must have been introduced with the `confirmDetentChange` callback. I've reverted the behavior and I'm again manually handling the Back press in the QE rather than using the Sheet's behavior. 

| Before | After |
|---|---|
| back_press_bug | back_press_fixe |

### Testing Steps

1. Open the QE with Avatar Picker
2. Go to Alt page
3. Focus the input field to show the keyboard
4. Use the system back button to hide the keyboard
5. Use the system back button to close the AltPage
6. Confirm the QE was not dismissed
7. Ideally, run some regression tests to make sure the `Unsaved changes` dialog behavior was not affected and that the system back button works on the "Into login" screen as well (before the OAuth).

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
@@ -7,6 +7,7 @@ import androidx.compose.animation.expandVertically
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.shrinkVertically
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.navigation.NavGraphBuilder
 import androidx.navigation.NavHostController
 import androidx.navigation.NavType
@@ -47,6 +48,12 @@ internal fun GravatarQuickEditorPage(
     onDismiss: (dismissReason: GravatarQuickEditorDismissReason) -> Unit = {},
 ) {
     val navController = rememberNavController()
+
+    LaunchedEffect(confirmDismissal) {
+        if (confirmDismissal && navController.currentDestination?.route != EditorNavDestinations.QUICK_EDITOR.name) {
+            onDoneClicked()
+        }
+    }
 
     NavHost(
         navController,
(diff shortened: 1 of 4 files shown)


etoledom (commented):
Test cases described working as expected 👍 
Including the unsaved changes alert.

I did find one interesting case:
If I try to dismiss dragging down the Alt text sheet, then I press the back button, the whole Quick Editor will dismiss:

CleanShot 2025-05-21 at 07 49 50

AdamGrzybkowski:
Amazing @etoledom - it's a separate bug that was also introduced with the `confirmDetent` callback. I will bring back to old behavior, meaning it will be possible to dismiss the sheet on all other "pages": OAuth, Splash and AltText.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/GravatarQuickEditorPage.kt
```diff
@@ -48,6 +49,12 @@ internal fun GravatarQuickEditorPage(
 ) {
     val navController = rememberNavController()
 
+    LaunchedEffect(confirmDismissal) {
+        if (confirmDismissal && navController.currentDestination?.route != EditorNavDestinations.QUICK_EDITOR.name) {
+            onDoneClicked()
+        }
+    }
```
AdamGrzybkowski, line 56: Now that we have to allow for sheet dismissal, each destination in this file needs to handle this separately. 
To keep it simple, I kept the logic here, and only the `EditorNavDestinations.QUICK_EDITOR` will have to handle it individually, since all the other pages confirm the dismissal.

etoledom:
So now the whole Quick Editor is dismissing after drag down from Alt Text editor.
If this was the intention then it's working 👍 

System back button and Cancel button will dismiss the Alt Text editor and go back to Avatar Picker
Drag down and tap outside of the sheet will dismiss the whole Quick Editor from the Alt Text Editor.

CleanShot 2025-05-22 at 08 29 31

Worth to mention that it was for me unexpected 🤔  
Probably because on iOS the behavior is different:

CleanShot 2025-05-22 at 08 32 46

Should we align behaviors here? Is this full sheet dismissal common on Android? (when there is more than one level of sheets on top)

cc @pinarol to help decide

AdamGrzybkowski:
>System back button and Cancel button will dismiss the Alt Text editor and go back to Avatar Picker
Drag down and tap outside of the sheet will dismiss the whole Quick Editor from the Alt Text Editor.

I have aligned the behavior we had before introducing that bug. AltPage is not a separate Sheet, so I don't think tapping outside (or dragging down) should go back to the picker.

etoledom:
> AltPage is not a separate Sheet, so I don't think tapping outside (or dragging down) should go back to the picker.

Right, it's presented as a navigation stack element. Yeah probably I expected that from experience on iOS, but if it's not a new modal on top, I'd agree it's expected for it to behave this way 👍

etoledom (approved):
Looking great! 🎉 
Let's continue moving forward 🚢

PR #637, opened 2025-05-26
First&Last name input fields in About editor

AdamGrzybkowski:
### Description

This PR adds two extra fields: First Name and Last Name. Since they differ in behavior from other personal fields, they will live under `Extras` section with a note.

Screenshot 2025-05-26 at 11 12 52

@etoledom I'm not sure about iOS but the `Extras` section always has the label and description visible even if no other sections are present. This differs from the personal and professional sections, which only have the section label visible if other sections are visible. 

### Testing Steps

1. Launch the Demo 
2. Select About scope and pick desired input fields
3. Verify the First&Last Name fields are visible and have the correct value
4. Update those fields and save the changes
5. Confirm the values were updated
6. Additionally, test the section title logic as mentioned above.


gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/components/AboutSection.kt
```diff
@@ -23,7 +23,7 @@ internal fun AboutSection(
 ) {
     Column(
         modifier = modifier,
-        verticalArrangement = Arrangement.spacedBy(32.dp),
+        verticalArrangement = Arrangement.spacedBy(24.dp),
```
AdamGrzybkowski, line 26: 32.dp is as in the design but it feels too much. It seems like iOS has less empty space between sections as well, but correct me if I'm wrong.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/components/AboutSection.kt
```diff
@@ -81,6 +93,20 @@ internal val AboutEditorField.descriptionRes: Int?
...
+    EXTRA,
+}
+
+private val AboutInputField.section: Section
+    get() = when {
+        isPersonal -> Section.PERSONAL
+        isProfessional -> Section.PROFESSIONAL
+        isExtra -> Section.EXTRA
+        else -> Section.PERSONAL
+    }
```
AdamGrzybkowski, line 108: I've added this so it's easier to work with sections.

etoledom:
> @etoledom I'm not sure about iOS but the Extras section always has the label and description visible even if no other sections are present. This differs from the personal and professional sections, which only have the section label visible if other sections are visible.

On iOS we remove the group title but keep the notice:



It does make sense to me if we choose to keep the group title also in this case 🤔

AdamGrzybkowski:
>On iOS we remove the group title but keep the notice:

@andrewdmontgomery @pinarol Any thoughts on this one?

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/components/AboutSection.kt
etoledom: Yes, we have 16pts of space, plus 8pts of padding on the field.



I think it looks good on both 👍

etoledom (approved):
Working great!
Every scenario works as described.

I'd say we can merge and then we modify the platform according to the decision on the `Extra` title :shipit:

pinarol:
Other titles might make sense on their own, but “Extras” doesn’t really. It makes you wonder, extras to what?

PR #639, opened 2025-05-26
Only enable the Save button when unsaved changes are present

AdamGrzybkowski:
### Description

The `Save` button is enabled even when no changes to the profile when made. This PR fixes that and the button will only be enabled when there's something to save.

Screenshot 2025-05-26 at 15 24 28

### Testing Steps

1. Launch the About Editor
2. Confirm the button is disabled
3. Make some changes
4. Confirm the button is enabled
5. Revert the changes - the button should get disabled again
6. Make some changes again
7. Save
8. Confirm the button is disabled
9. Feel free to test error scenarios when saving the profile as well


etoledom (approved):
Works as described 🎉

PR #641, opened 2025-05-27
Move animateContentSize modifier to a different component

AdamGrzybkowski:
### Description

There was a weird behavior when hiding the keyboard in the About editor. The whole sheet would look like it's going to close for a second. 

I'm not exactly sure why, but I narrowed it down to the `animateContentSize` modifier, and moving it to a component that enables scrolling in the About editor fixes the issue. 

| Before | After |
|---|---|
| back_bug | back_fixed |

### Testing Steps

1. Launch the About editor
2. Fully expand it
3. Tap on the input field to show the keyboard
4. Use back button or gesture to close the keyboard / Saving changes also works
5. Confirm there's no weird bounce in the sheet


etoledom (approved):
Works as described 🎉 
No more glitches when hiding the keyboard on a full height sheet

PR #642, opened 2025-05-27
Fix the About Editor in landscape mode

AdamGrzybkowski:
### Description

This PR adds extra handling to the About Editor UI whenever it's displayed in the Compact window. I'm not checking for orientation but rather the window size class - this ensures that when the space is available in landscape (eg, on tablets) we will show the whole UI.

There are several steps here (feel free to review by commit):
1. Email and the profile card are now hidden in Compact mode
2. The `save` button is part of the scrollable area in Compact mode
3. Custom IME padding modifier is applied to limit the extra padding added by the default one

Additionally, I have adjusted the AltTextPage with the same modifier.

I wanted to avoid listening to soft keybord visibility changes for several reasons:
1. The `WindowInsests.isImeVisible` is experimental, this means the API could be removed and this could cause issues for third-party apps.
2. Listening to that is not reliable for several reasons: it can behave differently on different system versions, and it depends on the host Activity config that we don't control.

Below you can see the portrait mode on the phone (this is EXPANDED window size class), landscape on the phone (COMPACT), and landscape on the Foldable device (MEDIUM). I know the foldable looks like it would also benefit from the same treatment, but this could target smaller phones in portrait mode, so I would avoid that. 

Although not perfect, this is at least usable now.

| Portait | Landscape | Foldable |
|---|---|---|
| Screenshot 2025-05-27 at 15 04 27 | Screenshot 2025-05-27 at 15 05 05 | Screenshot 2025-05-27 at 15 05 28 |

Extra, the Alt Page:
Screenshot 2025-05-27 at 15 15 16

### Testing Steps

1. Create a new resizeable emulator
3. Launch the demo app
4. Open the About Editor
5. Try different devices phone/foldable/table in both portrait and landscape


etoledom (approved):
It's working great as described 🎉 

I do have a couple of observations compared with iOS:

> Email and the profile card are now hidden in Compact mode

This is perfect. On iOS we also check for compact height instead of orientation and it works great 👍 

We do check for keyboard visibility (indirectly via text field focused state).

This give us the oportunity to only remove the card when a field is focused (showing the keyboard. The drawback is that this also happens on physical keyboard, which is not great.

```swift        
if !(vertcalSizeClass == .compact && model.isKeyboardPresented) { 
    // ... 
}
```

CleanShot 2025-05-27 at 16 45 04

I understand the technical limitations for this on Android, and I'm just wondering if we leave the diferences or we adapt iOS to behave as Android does 🤔 

Another difference derived from this is the Avatar Picker. Since we check for the keyboard, the avatar picker (without editable textfields) won't hide the profile card:

CleanShot 2025-05-27 at 16 37 47@2x

I do like the extra space on Android, but it seems to not be so needed on iOS side as the UI itself uses less vertical space 🤔 
This is also a question to either keep the difference as it is or to modify iOS to behave as Android.

PR #644, opened 2025-05-28
Upadte the default height for the loading section

AdamGrzybkowski:
### Description

With the AvatarAndAbout scope selected and horizontal scrolling, the sheet defaults to Peek detent instead of FullyExpanded. 

Tbh, I don't know the exact reason, but a quick fix here is to increase the height of the loading component. This should do for now, and I will try to investigate later.

| Before | After |
|---|---|
| peek_before | no_peek_after |

### Testing Steps

1. Select AvatarAndAbout
2. Open the QE
3. Confirm it's fully expanded

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/AboutEditor.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/abouteditor/AboutEditor.kt
@@ -42,6 +42,7 @@ import com.gravatar.quickeditor.ui.avatarpicker.titleRes
 import com.gravatar.quickeditor.ui.components.CtaSection
 import com.gravatar.quickeditor.ui.components.QEButton
 import com.gravatar.quickeditor.ui.editor.AboutInputField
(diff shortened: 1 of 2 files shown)


etoledom (commented):
Working as described on testing steps.

I noticed that this issue reapears when moving from About page to Avatar Picker page, when About page is medium height:

CleanShot 2025-05-28 at 13 28 39

Is this known?

AdamGrzybkowski:
>Is this known?

This is different - you have moved the sheet to the Peek detent (in About editor) and then switched to the Picker. The sheet tries to keep the same detent when changing between pages, so it stayed in the Peek one.

etoledom (approved):
> This is different - you have moved the sheet to the Peek detent (in About editor) and then switched to the Picker. The sheet tries to keep the same detent when changing between pages, so it stayed in the Peek one.

I see, so there are two detents close together on the Avatar Picker 🤔 

Since they are different things, let's move forward with this one 👍

AdamGrzybkowski:
>I see, so there are two detents close together on the Avatar Picker 🤔

Yes, ideally we would have different detents per page, but those have to be configured before launching the sheet and are not dynamic.

PR #645, opened 2025-05-28
Expand the sheet fully on focus

AdamGrzybkowski:
### Description

There's an issue when you tap on an input field and the keyboard shows up. In the Peek detent, the sheet won't be moved up above the keyboard, so the UI is essentially covered. In order to fix this, the sheet will now fully expand when it has the focus. 

| Before | After |
|---|---|
| keyboard_bug | keyboard_fix |

### Testing Steps

1. Open the About editor
2. Without expanding, tap on any of the visible input fields
3. Confirm the keyboard shows up and the sheet is expanded


etoledom (commented):
The animation alongside the keyboard works great on the About Editor 🎉 

I did notice something funky happening on the Alt text editor screen:
When expanded with the keyboard present, there's a small but constant vibration of the sheet (like scrolling up and down very rapidly and short distance). This is not visible on the gif though.

After trying different things expanding and scrolling it, this vibration became slower and longer. This can be seen at the end of the gif.

CleanShot 2025-05-28 at 16 18 14

etoledom (approved):
> I did notice something funky happening on the Alt text editor screen:

I saw this happening on a different branch also, so it's not related to the changes on this branch.

PR #646, opened 2025-05-28
Move sheet to hidden detent internally when dismissing

AdamGrzybkowski:
### Description

Normally, the sheet animates when being closed. There are a few scenarios related to the authorization where we wouldn't do that, but the sheet would be simply hidden by the consuming app (demo app in our case) on the `dismiss` callback. 

This is now fixed, and we are animating the sheet to a hidden state internally when the `onDismiss` is invoked.

| Before | After |
|---|---|
| close_before | close_after |

### Testing Steps

1. Open the Editor with the provided random auth token
2. Tap Close
3. Confirm the sliding animation is present


etoledom (approved):
Working great 🎉

PR #647, opened 2025-05-29
Consume navigation bar insets before applying imePadding

AdamGrzybkowski:
### Description

@etoledom [reported](https://github.com/Automattic/Gravatar-SDK-Android/pull/645#pullrequestreview-2875323596) weird behavior in the `AltTextPage` with the keyboard shown. I narrowed it down to the `positionAwareImePadding()` custom modifier. This forced me to look for alternatives, and I actually understood why we had the excessive padding in the first place. It turns out the insets are the sum of all, meaning if we set `statusBarPadding` and `imePadding`, both will be applied. The thing is, the keyboard handles the status bar padding itself so when it's shown we want to only have the `imePadding` applied. In order to fix that, we have to consume the padding set in the parent (the status bar padding) to make sure only the padding set in the children component is used. 

There should be no UI changes, except for the UI glitch that should now be gone. 

### Testing Steps
@etoledom I believe you now best how to test it!


etoledom (approved):
This PR completely solves the mentioned issue 🎉 

Thank you for the fix 🙏

PR #648, opened 2025-05-30
Rename leftover scopConfig with scopeOption

AdamGrzybkowski:
### Description

I missed some names when changing the public API structure and left `scopeConfig` instead of `scopeOption`. This PR aligns all the names. 

### Testing Steps

Green CI should be good.


etoledom (approved):
Looks good!

PR #652, opened 2025-06-03
Update the AboutMe input field's height

AdamGrzybkowski:
### Description

The AboutMe input field height should be fixed at 4 lines. This aligns it with iOS and the designs.

Screenshot 2025-06-03 at 15 19 05

### Testing Steps

1. Open the About editor
2. Confirm the AboutMe inptu field has 4 lines visible even if empty


etoledom (approved):
Looking good! 🎉

PR #654, opened 2025-06-04
Dismiss action should trigger unsaved changes dialog regardless of the page

AdamGrzybkowski:
### Description

Following the iOS implementation, the alert dialog shown when there are unsaved changes will also be triggered on the Avatar Picker page (in the combined scope).

In order to do that, I had to move the alert dialog component up in the hierarchy to a shared place (QuickEditor.kt). This actually simplified the code a bit as there is less communication between it and the `AboutEditor`.

### Testing Steps
1. Generally, test the confirmation logic to make sure it behaves as before - that means testing back button, drag down, tap outside the sheet, and the Done button with and without unsaved changes in the About Editor. 
2. Open the QE in the combined mode
3. Edit one of the input fields in the about section
4. Switch to the avatar picker
5. Try to close the QE
6. Confirm that you are presented with the alert dialog


etoledom (approved):
Everything working as expected 🎉 

The unsaved changes alert works as before from the About editor from any dismissal method.
It now also shows on the Avatar Picker page when there are unsaved changes on the Avatar & About scope.

PR #655, opened 2025-06-04
Adjust the top bar design

AdamGrzybkowski:
### Description

1. The drag handle is now removed
2. The font size of the button in the top bar is increased

| Before | After |
|---|---|
| old_toolbar | new_toolbar |


etoledom (approved):
Looks good to me.

Here's a side by side comparison to iOS:

CleanShot 2025-06-04 at 16 38 33@2x

PR #657, opened 2025-06-06
Bump Kotlin to 2.0.21

AdamGrzybkowski:
### Description

Kotlin bump to 2.0.21, along with an update to the Compose compiler setup, as it changed after Kotlin 2.0. 
`kotlinOptions {` is also deprecated now, so I replaced it with `kotlin { compilerOptions`.

It also looks like the implementation for Parcelize plugin was changed and it now generates final functions - this can be seen in the updated `.api` files. 

### Testing Steps

1. Build the project and perform a smoke test on the demo app.
2. Review


AdamGrzybkowski:
Interestingly, I can no longer reproduce this issue https://github.com/Automattic/Gravatar-SDK-Android/issues/183.

wzieba (approved):

PR #659, opened 2025-06-09
Change tonal elevation to deafult (0.dp) of the bottom sheet

AdamGrzybkowski:
### Description

As suggested by Shaun, I'm removing the tonal elevation that adds a blueish tint to the background. 

| Before | After |
|---|---|
| Screenshot_20250609_103333 | Screenshot_20250609_103246 |
| Screenshot_20250609_103327 | Screenshot_20250609_103258 |

### Testing Steps

Launch the QE in both color themes and confirm the background color.

### DIFF
--- a/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
+++ b/gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/editor/bottomsheet/GravatarQuickEditorBottomSheet.kt
@@ -317,7 +317,6 @@ private fun GravatarModalBottomSheet(
                                         hasFocus = state.hasFocus
                                     }
                                     .fillMaxWidth(),
-                                tonalElevation = 1.dp,
                             ) {
                                 Column(
                                     horizontalAlignment = Alignment.CenterHorizontally,


etoledom (approved):
Looking good!

Is it normal that there are no snapshot updates?

AdamGrzybkowski:
>Is it normal that there are no snapshot updates?

Hmm I'm not sure I understand. What snapshot updates?

hamorillo:
I guess @etoledom refers to not having any screenshot tests modified, which is also strange to me. (But it's been a while since the last time I worked on the project, so I may be missing something).

AdamGrzybkowski:
Oh right! We don't have a screenshot test for the sheet component `GravatarQuickEditorBottomSheet` as a whole - only for the Composables that are inside it. So essentially everything that is displayed in the sheet container, but not the container itself.

AdamGrzybkowski:
Ok I got it - the Dialog component that is used under the hood would throw an exception in the screenshot tests. I see that this might be fixed in the newer versions of the plugin, but simply updating it breaks all the screenshot tests 🙃

etoledom:
> I guess @etoledom refers to not having any screenshot tests modified

This! By "snapshots" I meant "screenshots" 😅 
Thank you 🙏

hamorillo:
We can create a new ticket to work on it when possible. 🙂

AdamGrzybkowski:
@hamorillo Done GRA-161

PR #660, opened 2025-06-11
Remove Peek detent from the QuickEditor bottom sheet

AdamGrzybkowski:
### Description

The Peek detent doesn't have much sense in the About Editor, and to stay consistent with behavior in the combined scope and the Avatar only, we are removing the Peek detent completely. This might come back later as a configurable option. 

### Testing Steps
1. Launch QE in different modes and confirm that it always goes to FullyExpanded detent.

cc @pinarol


hamorillo (approved):
🚀

PR #661, opened 2025-06-11
Set navigation bar padding on the content, not the sheet container

AdamGrzybkowski:
### Description

I've noticed that the background at the bottom of the sheet has a different color. This was introduced with the removal of the `tonalElevation` in one of the recent PRs.

To fix this, I moved the `statusBar` padding to the content so that the sheet container has the same background all the way down. 

| Before | After |
|---|---|
| Screenshot_20250611_091729 | Screenshot_20250611_091312 |

### Testing Steps

1. Launch the QE.
2. Confirm the sheet container has the same background from top to bottom.


hamorillo (approved):
Nice catch! 🚀

PR #663, opened 2025-06-12
Update the code in the usage-quickeditor.md

AdamGrzybkowski:
### Description

The doc shows a function that doesn't exist. 

### Testing Steps

Verify that the docs match the real code.


hamorillo (approved):
Nice catch! 🚀

PR #664, opened 2025-06-13
QE Top section updates

AdamGrzybkowski:
### Description

This PR makes some changes to the QE top bar and the profile card:
 1. The icon in the `View button` is changed to "external link".
 2. The email above the profile card is removed from the QE if the auth token is provided from a 3rd-party app.
 3. The font scaling inside the profile card is limited to 1.5f.

https://github.com/user-attachments/assets/04a5372d-e634-4814-954b-5883a1cd9cb1

### Testing Steps

1. Launch the QE with and without the token - confirm the email visibility.
2. Test the new icon in RTL - it should be mirrored.
3. Test the font scaling and confirm it stops at a certain threshold.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/ProfileCard.kt
```diff
@@ -57,74 +59,87 @@ internal fun ProfileCard(
...
-                            .padding(top = 5.dp),
-                    )
+                            },
+                            viewProfile = { state ->
+                                if (state !is ComponentState.Empty) {
+                                    ViewProfileButton(
+                                        state = state,
+                                        modifier = Modifier.height(32.dp),
+                                    ) {
+                                        Icon(
```
hamorillo, line 120: ❓ I wonder if (probably not now) we should replace the default icon in the UI library with this new icon.

hamorillo (approved):
🚀 Thanks for the changes! I left a minor question to discuss.


✅ Launch the QE with and without the token - confirm the email visibility.
✅ Test the new icon in RTL - it should be mirrored.
✅ Test the font scaling and confirm it stops at a certain threshold.

gravatar-quickeditor/src/main/java/com/gravatar/quickeditor/ui/components/ProfileCard.kt
AdamGrzybkowski: I believe iOS doesn't use the public component, so the new icon was changed only in the QE. That's why I didn't modify the public component as well. The icon can be easily modified (as we did here) so I don't think it's a big issue.

hamorillo: Yeah, it makes total sense for this PR. I was thinking aloud. I agree that it's not a big issue.

PR #667, opened 2025-06-27
Update OpenApi definition and regenerate new classes

hamorillo:
### Description

We've added some new fields in the Profile model, so we need to update the OpenApi definition and regenerate the models.

### Testing Steps

- CI and code review.


gravatar/openapi/api-gravatar.json
```diff
@@ -23,6 +23,14 @@
...
       "description": "Operations about user avatars"
+    },
+    {
+      "name": "qr-code",
+      "description": "Operations about QR codes"
+    },
+    {
+      "name": "experimental",
+      "description": "Experimental operations that might be subject to change. Use with caution."
     }
```
AdamGrzybkowski: I think we don't have any of those in the SDK, right?

hamorillo: 🤦 I removed other methods we currently don't support, but forgot to remove that one. Updated!

Thanks!

AdamGrzybkowski (approved):
LGTM 👍

PR #669, opened 2025-07-22
[Tooling] Update Ruby dependencies

iangmaia:
I've noticed the `dependabot` PR https://github.com/Automattic/Gravatar-SDK-Android/pull/668 was changing `nokogiri`'s and the `Gemfile.lock` `PLATFORMS`, so I took the chance to update `.bundle/config` to force Ruby platform instead and ran `bundle update`.


AliSoftware (approved):
Nice catch!

PR #675, opened 2026-02-13
Pin faraday >= 1.10.5 for security fix

mokagio:
## Summary

- Pin faraday to >= 1.10.5 to address [security vulnerability](https://github.com/lostisland/faraday/pull/1665)
- Faraday 1.10.5 backports the fix from 2.x to the 1.x line

## Test plan

- [ ] CI passes with updated dependency

🤖 Generated with [Claude Code](https://claude.com/claude-code)

*Posted by Claude (Opus 4.6) on behalf of @mokagio with approval.*

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,3 +6,7 @@ gem 'danger-dangermattic', '~> 1.1'
 gem 'fastlane', '~> 2.222'
 gem 'fastlane-plugin-wpmreleasetoolkit', '~> 13.0'
 gem 'rubocop', '~> 1.65'
+
+# Security: https://github.com/lostisland/faraday/pull/1665
+# Faraday 2.0 is not compatible with Fastlane
+gem 'faraday', '~> 1.10', '>= 1.10.5'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -95,7 +95,7 @@ GEM
     drb (2.2.3)
     emoji_regex (3.2.3)
     excon (0.112.0)
-    faraday (1.10.4)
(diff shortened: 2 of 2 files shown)


AliSoftware (approved):
💡  Would be nice to have the agent loop automatically take care of the follow-up needed to  address the Danger CI failure itself instead of us doing it manually

mokagio:
> 💡 Would be nice to have the agent loop automatically take care of the follow-up needed to address the Danger CI failure itself instead of us doing it manually

I've done it for other PRs in the past days, sounds like a great use case for a skill.

Some repos should already have labels and milestones instructions in their `AGENTS.md`, but I'm of two minds about this. It's super convenient to have the agent set label(s) and milestone. It's also something that requires a few API calls for it to discover which value to use. On the labels side, this could be address—at least for folks in Apps Infra—by instructing personal agents to use "tooling" (perhaps with a map of repo-label naming) but the milestone requires a lookup because it changes over time. This feels like a waste of tokens if we assume that one or more humans will eventually look at the PR to review and merge and it would be trivial, albeit toiling, for them to set the correct values. But maybe I'm overthinking this...

iangmaia:
> one or more humans will eventually look at the PR to review and merge and it would be trivial, albeit toiling, for them to set the correct values.

IMHO it should be on the PR author side (agent or human) to set these labels 🤔 handing the PR to a reviewer with failing checks should be avoided as much as possible.
Though sometimes milestones can relate to other project constraints, release schedules, etc, so it might be beyond what an agent will gather now...

